Issue #21, opened 2012-06-12
Fail compile if repo contains vendor/bundle


wuputah:
Thoughts?

Pros: Committing this directory is harmful, breaks native gems, etc.

Cons: People should be able to vendor everything if they really want to. (But there's a "right way" to do that, and this isn't it... (use `bundle package`))

PR #23, opened 2012-06-12
Remove `vendor/bundle` if it already exists

hone:
Code ideas for #21


schneems:
As a side note it looks like bundler changed the directory to `vendor/cache`. If you're finding this for historical reasons realize that at the time checking in gems after running bundle pack was at one time considered best practice. We don't do anything to stop people from checking in `vendor/cache`. If you're using no gems with native extensions technically you could deploy.

Issue #72, opened 2013-03-11
Detect manifest-(fingerprint).json instead of manifest.yml for rails4


brettgoulder:
manifest.yml is not created in rails4 anymore, it's manifest-(fingerprint).json

https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/rails4.rb#L40

PR #74, opened 2013-03-12
Detect rails 4 manifest file

catsby:
This patch implements #72 and checks for the manifest file generated by Rails 4 to detected if assets are already compile.


Zahweh:
+1

schneems:
why the plus one? We've done this for over 3 years now

PR #85, opened 2013-04-14
Use SSH key from env if provided.

ejholmes:
I think this is a better way to handle private GitHub gem dependencies.

If a user adds an SSH_KEY env var (and assuming they're using [user-env-compile](https://devcenter.heroku.com/articles/labs-user-env-compile)), this allows private gems to be cloned from GitHub, without going through hoops like [this](https://gist.github.com/masonforest/4048732).

Thoughts?

**Example**

```bash
heroku labs:enable user-env-compile -a foo-bar-1234
heroku config:set SSH_KEY="$(cat ~/.ssh/id_rsa)" -a foo-bar-1234
```

```ruby
source 'https://rubygems.org'

gem 'rake'
gem 'private-gem', git: 'git@github.com:ejholmes/private-gem.git'
```

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -68,6 +68,7 @@ class LanguagePack::Ruby < LanguagePack::Base
 
   def compile
     Dir.chdir(build_path)
+    write_ssh_key
     remove_vendor_bundle
     install_ruby
     install_jvm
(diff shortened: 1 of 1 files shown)


iainbeeston:
@schneems Is there an official way to do this? (Just wondering, since this PR was closed)

schneems:
The officially supported way to access private gems would be to use a private gem server

```
source "https://#{ENV["PRIVATE_GEMSERVER_CREDENTIALS}@example.com"
```

Alternatively you can use username password, or token when accessing your git dependencies from github. Unfortunately this exposes the URL in the `Gemfile.lock`. I think the right approach would be to add  a feature to bundler to allow for scrubbing out sensitive information

```
# something like this maybe?
gem 'private-gem', git: "https://#{ENV['USER_PASS']}@github.com:ejholmes/private-gem.git", as: "https://github.com:ejholmes/private-gem.git"
```

Or something similar. (maybe not storing basic auth at all, providing a secure key that gets stripped out, etc.)

In general magic ENV vars are difficult to work with and don't expose easily. It would be better to solve this problem from the community perspective and we can help provide support for the best community practices.

PR #227, opened 2014-02-18
Rails 4.1.0 Support

hone:
This adds supports for Rails 4.1.0 RC that is coming out today. The new changes here are the following:
- don't overwrite `config/database.yml` anymore, since @josevalim and @schneems worked on getting the env var`DATABASE_URL` to be merged into the yml file. See [this issue](https://github.com/rails/rails/pull/13463#issuecomment-31480799).
- support `config/secrets.yml`. We now setup the env var `SECRET_KEY_BASE`, which by default will be used in production. You can roll this credential by generating a new one and setting it as the config var:

```
$ heroku config:add SECRET_KEY_BASE=`ruby -rsecurerandom -e "puts SecureRandom.hex(64)"`
```

This also sets config vars again during `bin/release` on first app push. I want to have config vars that we are setting that the user can/should change be set here. Stuff like `SECRET_KEY_BASE`, `RAILS_ENV`, etc.

### DIFF
--- a/hatchet.json
+++ b/hatchet.json
@@ -47,5 +47,8 @@
     "sharpstone/rails3-to-4-no-bin",
     "sharpstone/rails4_windows_mri193",
     "sharpstone/rails4-fail-assets-compile"
+  ],
+  "rails41": [
+    "sharpstone/rails41_scaffold"
   ]
 }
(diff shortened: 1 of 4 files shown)


schneems:
To note, we didn't use config vars because you can only set those on first push.

PR #243, opened 2014-04-01
ruby_version is never falsy

elia:
I was wondering if there was some opportunity for a speedup by caching ruby installation, perusing the source code I found [these lines](https://github.com/heroku/heroku-buildpack-ruby/blob/cb9b6c85bc482fac5ff23f81a57cf80dc257e0fb/lib/language_pack/ruby.rb#L254):

```ruby
  def install_ruby
    instrument 'ruby.install_ruby' do
      return false unless ruby_version
```

and apparently there's no way ruby_version can be falsy ([is always an instance of `RubyVersion`](https://github.com/heroku/heroku-buildpack-ruby/blob/cb9b6c85bc482fac5ff23f81a57cf80dc257e0fb/lib/language_pack/ruby.rb#L177)), maybe it's a legacy of the extraction of `RubyVersion` (15df63760f18fd1197de06f75923dd295d2d3d85).

happy to be proved wrong :)

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -255,7 +255,7 @@ private
 
       FileUtils.mkdir_p(slug_vendor_ruby)
 
-      if ruby_version.changed?
+      if ruby_version.changed? or not(cache.exist? slug_vendor_ruby)
         puts "Preparing binaries ..."
         fetch_ruby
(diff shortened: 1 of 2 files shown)


schneems:
We track the benchmark time of ruby installs and they're all really trivial, fractions of seconds since we're running on AWS and pulling from S3.

PR #290, opened 2014-07-24
Remove sqlite3 from database_url generation.

allomov:
I'm not sure that this lines are needed since `sqlite3` officially is not supported. 

By the way I was able to deploy rails application that uses `sqlite3` by defining `DATABASE_URL=sqlite3:///some-strange-database-name.db`. Yes, it's odd solution, but possibly might be helpful to someone.


schneems:
I don't disagree, however i don't think keeping it in there is hurting anything.

PR #305, opened 2014-10-02
Jdk8

jkutner:


hone:
Passes for me locally. Fails on downloading node from npmjs.org. Working on a PR to use s3pository instead.

PR #306, opened 2014-10-08
user can configure # of bundler jobs

jjb:
As a workaround for the bundler bug: https://github.com/bundler/bundler/issues/3174

Also described in heroku support ticket 129852


schneems:
Bug was fixed when we upgraded to 1.7.12.

jjb:
image

PR #316, opened 2014-11-10
Add more generic application setup rake task

billywatson:
I've added an `application:setup` rake task and moved the Rails-specific `assets:precompile` into the Rails language pack. This will allow all ruby applications to do some final application setup tasks on deploy and not force them to use the Rails-specific name of `assets:precompile` to get it done. I've also updated the docs surrounding this change and added a basic test.

### DIFF
--- a/.ruby-gemset
+++ b/.ruby-gemset
@@ -1,0 +1,1 @@
+heroku-buildpack-ruby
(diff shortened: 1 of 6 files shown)


billywatson:
So did this not get merged down? Why?

schneems:
This PR would break existing ruby apps that rely on the functionality you removed. I'm not opposed to adding in another entry point like https://github.com/heroku/heroku-buildpack-ruby/pull/330 however it's a huge bike shed. 

For now if you really need something to hook into i would enhance `assets:precompile` which is a supported interface.


Thanks for the PR :heart: unfortunately it's really hard to take new features into the buildpack.

billywatson:
Np, just looking for more info. I'll look at #330

billywatson:
Also, couldn't I also just not remove the assets:precompile from where I removed it and just add the other task?

billywatson:
This doesn't have to be all or nothing, I'm happy to modify the pull request

schneems:
I understand, I know this is frustrating. It's not going to be merged in.



—

billywatson:
Well, not super frustrated or anything, was first trying to get an understanding as to why it wouldn't be merged down. I'll comment on the other thread

PR #323, opened 2014-12-10
Auto-scaling memory for JRuby

jkutner:


hone:
@jkutner I know we talked about this before, by why are we removing `default_java_tool_options` for config vars?

jkutner:
Because:
+  It's redundant (we're already setting it in profile.d)
+  It's redundant (we already load JAVA_OPTS). It probably just makes people puzzled as to whether they should put an arg in JAVA_OPTS, JRUBY_OPTS, or JAVA_TOOL_OPTIONS. Lots of overlap.
+  These defaults are permanent. Once they are set on an app, the buildpack can't change them -- only the user can. So that makes it impossible for the buildpack to adjust things when it upgrades the system (for example, let's say a Bundler upgrade suddenly requires some weird Java arg for JRuby). I'm dealing with a lot of this in the Java buildpacks because these defaults were set prior to profile.d existing

That said, it might be good to do something like (because some folks may have already customized this):

```ruby
def default_java_tool_options
  "-Xmx${JVM_MAX_HEAP:-\"384\"}m -Djava.rmi.server.useCodebaseOnly=true $JAVA_TOOL_OPTIONS"
end
```
But I think JAVA_OPTS is preferred because it will override JAVA_TOOL_OPTIONS.

hone:
@jkutner should we not be using java_tool_options at all then and simplify? The Java team back in the day told me to set this I believe.

jkutner:
In the JVM buildpacks (java, scala, clojure), I have the buildpack set JAVA_TOOL_OPTIONS and the user set JAVA_OPTS. This works nicely because the JAVA_TOOL_OPTIONS is intended to be set when "the command line to start the application is not readily accessible"[[1](http://www.oracle.com/technetwork/java/javase/envvars-138887.html#gbmsy)]. 

JAVA_OPTS is just a convention that's used by lots of frameworks such as Tomcat. There isn't anything in the JVM that actively picks that up. So it makes a good optional addition to the start command.

All that said. I don't see anywhere that you are using JAVA_OPTS. I think we can probably do away with it (unless JRuby picks it up, which it may do but I think the same stuff can be set with JRUBY_OPTS), and move those args to JRUBY_OPTS with the `-J` prefix.

hone:
@jkutner this looks right from what we talked about today?

jkutner:
Yes, this looks correct. I confirmed it as well by deploying with this branch, and I now I see this during compile (which indicates the script is correct because it sets Xmx to 768m):

```
remote: Picked up JAVA_TOOL_OPTIONS: -Xmx768m -Djava.rmi.server.useCodebaseOnly=true
```

PR #324, opened 2014-12-10
Add JVM install dir to the compile time PATH

jkutner:
This fixes a bug where gems like `scrypt` cannot be installed on cedar-14. On cedar classic they installed fine because the used the `java` command on the stack image.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -205,7 +205,6 @@ private
   # sets up the environment variables for the build process
   def setup_language_pack_environment
     instrument 'ruby.setup_language_pack_environment' do
-      ENV["PATH"] += ":bin" if ruby_version.jruby?
       setup_ruby_install_env
       ENV["PATH"] += ":#{node_bp_bin_path}" if node_js_installed?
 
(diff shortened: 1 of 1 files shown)


hone:
@jkutner I feel like this should work without this patch since the JVM is symlinked to "bin" and that is being added to the path both in `default_path` and above. Am I missing something? I think the bin logic should be simplified.

jkutner:
It does not work. You can put `scrypt` in any JRuby `Gemfile` to reproduce.  

I'm pretty sure the problem is that the PATH contains the path to the vendor/jvm/bin in the *slug dir* (i.e. `/app`), but at compile time, the vendor/jvm is still in a tmp dir (as part of Codon). @hunterloftis saw this with the node buildpack, which is why he has two separate paths for `profile.d` and and `export`:
https://github.com/heroku/heroku-buildpack-nodejs/blob/yoga/bin/compile#L182-L183

This will work once Java is on the stack image again. But I think most users will expect the compile phase to use the same JVM as runtime.

hunterloftis:
Yep, ran into that with multi buildpack apps and node. In codon, the node / npm / etc bins need to be executed in the build directory, something like `/foo/bar123445635435345` (not `/app...`)

hunterloftis:
Now I write the profile here: https://github.com/heroku/heroku-buildpack-nodejs/blob/master/lib/build.sh#L138-L143

And write the exports for multi bp here: https://github.com/heroku/heroku-buildpack-nodejs/blob/master/lib/build.sh#L145-L149

schneems:
What's the status here? There's some merge conflicts

jkutner:
It appears @hone has account for this in his refactoring of my other JRuby PR. I'm closing this one.

PR #327, opened 2014-12-22
Set RAILS_SERVE_STATIC_FILES by default

schneems:
Goes hand in hand with https://github.com/rails/rails/pull/18100


hone:
I was just thinknig it might be nice to segment the various rails changes but mostly b/c of the confusing bit when they check `heroku config`.

schneems:
seems good, i'm trying to get someone to bump my API limit, so the tester can run again :cake:

Issue #335, opened 2015-01-09
Node Download Times out causing App deploy failure


schneems:
Since moving from S3 to the new node endpoint the timeout isn't long enough sometimes. We're seeing ~100 deploys fail a day due to this. Ticket 208962

```
 !     Command: 'set -o pipefail; curl --fail --retry 3 --retry-delay 1 --connect-timeout 3 --max-time 30 http://nodejs.org/dist/v0.10.30/node-v0.10.30-linux-x64.tar.gz -s -o - | tar zxf - node-v0.10.30-linux-x64/bin/node' failed unexpectedly:
 !     
 !     gzip: stdin: invalid compressed data--format violated
 !     tar: Unexpected EOF in archive
 !     tar: Unexpected EOF in archive
 !     tar: Error is not recoverable: exiting now
```

We can bump the timeout or cache the node binary. The second is much harder as we need to also code invalidation logic, but it will make deploys faster.

josegonzalez:
Was there a reason the deploys moved from s3 to the node endpoint? I assume cost, correct?

schneems:
S3pository is on S3. This issue is closed by #336

hone:
@josegonzalez we actually moved from the official node binary hosts to our own hosted/cached versions of it b/c the nodejs.org infrastructure seems to be unreliable more often than we'd like.

PR #336, opened 2015-01-12
use s3pository over nodejs.org

hone:
nodejs.org is not reliable


hone:
@schneems thoughts?

schneems:
What is s3pository? Is it a thing we run? Would it fix #335 ?

hone:
@schneems yes, it's a thing we run. It's downloads cached on s3 run by us. the node buildpack [uses it](https://github.com/heroku/heroku-buildpack-nodejs/blob/master/lib/build.sh#L79).

schneems:
:shipit:

PR #338, opened 2015-01-20
Scale back JRuby heap size on PX dynos 

jkutner:
Scale back JRuby heap size on PX dynos to give room for native memory allocation


hone:
Thanks @jkutner

PR #341, opened 2015-01-21
Fail fast if ENV["STACK"] is missing

cyx:
Had some minor issues with it while trying to make a Dockerfile work, but TL;DR STACK wasn't set and I was getting a weird error which didn't seem intuitive.

Let me know if this general direction works, we can also maybe pass DEFAULT_LEGACY_STACK as second param to fetch if that's the preferred behavior as well.


schneems:
I'm fine with this, @hone thoughts?

MatthewMueller:
Running into this error now, what is `STACK` supposed to be?

cyx:
cedar-14 / cedar

PR #346, opened 2015-02-21
control all of our test repos

hone:
This test broke when @jkutner updated the kissaten/jruby-minimal repo. I just forked and rolled back so our test continue to work.


jkutner:
Sorry bro :(

hone:
@jkutner all good! we probably should have forked it in the first place, so you're not held to the ruby buildpack. :)

PR #347, opened 2015-02-21
Make `bundle clean` have the same context as `bundle install`

hone:
If a user has a crazy Gemfile where it executes code based on the environment, `bundle clean` might fail while `bundle install` works because the context is different. Let's pass the user env to `bundle clean` as well.


schneems:
:+1:

PR #349, opened 2015-02-27
Enable Rails 5 to work on Heroku

schneems:
- Add rails42.rb
- Set RAILS_SERVE_STATIC_FILES by default on Rails 4.2+ (rails/rails#18100)
- Add rails5.rb


hone:
Looks good to me.

PR #351, opened 2015-03-04
Allow user to use a gemfile in a different location

schneems:


hone:
The issue with this is we need to do this as well for `bin/detect`, but it doesn't have user env vars available to it. :( We'll need to discuss if we change that. Closing for now.

Issue #76, opened 2013-03-21
Remove .git directories from "gems" installed from git sources


danp:
ref #16 

If I have:

``` ruby
gem "rails", :git => "git://github.com/rails/rails.git", :ref => "3-0-stable"
```

In my Gemfile I end up with a directory like:

```
vendor/bundle/ruby/1.9.1/bundler/gems/rails-77403a9b0407
```

in my slug. This directory contains a `.git` directory. In a slug created from an app with the above line in its Gemfile that `.git` directory is 64M. I believe this directory can be removed from the slug and things will still work.

schneems:
We should probably do this

dpiddy:
:+1: I still think this should be safe.

rodrigoddalmeida:
Pinging @schneems. Any news on this?

jmccartie:
:+1:

cyx:
:+1:  Willing to take a stab on this if we're all good with the idea.

PR #352, opened 2015-03-10
Deleting .git directories

rodrigoddalmeida:
I was having some problems with the slug size limit and figured this should help it. (I don't know if this is the right place to do it, but seems to be working nicely)

In my case, the deploy time increased by 5±1 seconds, but decreased slug size by ~45%.

Is there anything I should add or do?

In time: ref to #76

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -601,6 +601,9 @@ ERROR
   end
 
   def post_bundler
+    # Deletes .git directories across gems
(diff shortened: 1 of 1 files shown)


Tonkpils:
@rodrigoddalmeida this can be closed since it was fixed by https://github.com/heroku/heroku-buildpack-ruby/commit/cd8894dd7b68a259344815c5fb17dc3906bfb898

rodrigoddalmeida:
@Tonkpils Thanks for the heads up!

PR #354, opened 2015-03-18
Add support for changing number of bundle jobs

titanous:
The `BUNDLE_JOBS` environment variable can be used to change the number of jobs that will be used to bundle (defaults to 4).

This can be set to 1 to work around bundler/bundler#3063 or a higher number to speed up deployments in some situations.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -509,10 +509,11 @@ WARNING
   def build_bundler
     instrument 'ruby.build_bundler' do
       log("bundle") do
+        bundle_jobs    = env("BUNDLE_JOBS") || "4"
         bundle_without = env("BUNDLE_WITHOUT") || "development:test"
         bundle_bin     = "bundle"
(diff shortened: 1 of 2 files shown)


titanous:
@schneems Is this something you're interested in merging? If not, I can close. Thanks!

schneems:
I don't think we'll merge in. Sorry we had a conversation about it internally and forgot to update the posting. This has come up a few times in a few different ways. 

Supporting this makes managing it difficult for the majority of customers. You can make your bundle installs be really slow by setting to an arbitrarily high number, or maybe you set it to 1 to work around the previous concurrency bug and never updated after it was fixed. I would encourage you to use your fork until we can bump up to a fixed version, but right now we won't be adding in the ability to specify job number. Adding this point in also takes away our ability to tune reliably in the future.

Thanks so much for the patch :heart: sorry you didn't hear anything for awhile. At least you can set `BUILDPACK_URL` to your branch until we get this sorted out. If your not already, I would recommend subscribing to the Heroku changelog https://devcenter.heroku.com/changelog or to notifications on this repo (we do almost all works via PRs and treat master as extremely stable).

titanous:
No worries, thanks!

PR #355, opened 2015-03-25
set RACK_ENV=deployment by default

wuputah:


leshik:
@wuputah @schneems could you please tell why this was closed without merging? I thought RACK_ENV has no valid options besides `none`, `deployment` and `development`. Thus `production` is the equivalent to `none`.

schneems:
We've been doing this for so long people depend on it.

PR #357, opened 2015-04-02
Support Ruby/Rails apps whose root folder is not the repo folder

shageman:
This is achieved through detecting a `.ruby_app_root` file in the root of the repo, which contains a single line with the relative path to the subdirectory containing the app root.

See https://github.com/shageman/sportsball/tree/web_container_dir for a deployable example (use `web_container_dir` branch).

_I am not looking for this PR to be accepted_, but rather I am looking for input on how to solve this problem with fewer hacks. 

Some of the problems with this code:
* compile: use of global variable `$folder` to not change arity of `detect` on language packs
* default config: `-e $RAILS_ENV` stopped working, so `-e production` is hardcoded for now
* semi-hardcoded paths in env variables
* running rake, e.g., `heroku run rake db:migrate`, does not yet work

Specific questions:
* Is the `.ruby_app_root` detection acceptable for the default ruby buildpack?
* Is the basic approach viable if I work out the kinks and hacks?
* What tests should be added for acceptance to be considered? One version of Rails or all supported language packs?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ vendor/bundle/*
 .ruby-version
 buildpacks/*
 .anvil/
+
+.idea
--- a/bin/compile
+++ b/bin/compile
@@ -9,7 +9,14 @@ require "language_pack/shell_helpers"
 
 begin
   LanguagePack::Instrument.trace 'compile', 'app.compile' do
-    if pack = LanguagePack.detect(ARGV[0], ARGV[1])
+    
+    def app_root
+      path_to_root = Pathname.new(ARGV[0] || "").join(".ruby_app_root")
+      $folder = File.exist?(path_to_root) ? File.read(path_to_root).delete!("\n") + "/": ""
+      File.join(ARGV[0], $folder)
+    end
+
+    if pack = LanguagePack.detect(app_root, ARGV[1])
       LanguagePack::ShellHelpers.initialize_env(ARGV[2])
       pack.topic("Compiling #{pack.name}")
       pack.log("compile") do
--- a/bin/detect
+++ b/bin/detect
@@ -5,6 +5,9 @@ require 'pathname'
 if Pathname.new(ARGV.first).join("Gemfile").exist?
   puts "Ruby"
   exit 0
+elsif Pathname.new(ARGV.first).join(".ruby_app_root").exist?
+  puts "Ruby (non-standard root folder)"
+  exit 0
 else
   puts "no"
   exit 1
--- a/bin/release
+++ b/bin/release
@@ -2,5 +2,11 @@
 
 require 'pathname'
 
-release = Pathname.new(ARGV.first).join("tmp/heroku-buildpack-release-step.yml")
+def app_root
+  path_to_root = Pathname.new(ARGV[0] || "").join(".ruby_app_root")
+  $folder = File.exist?(path_to_root) ? File.read(path_to_root).delete!("\n") + "/": ""
+  File.join(ARGV[0], $folder)
+end
+
+release = Pathname.new(ARGV.first).join("#{app_root}/tmp/heroku-buildpack-release-step.yml")
 puts release.read if release.exist?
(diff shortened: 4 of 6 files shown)


schneems:
We've talked about this in the past. Previously we decided that we're an opinionated platform and one of our opinions was that your app should be self contained http://12factor.net/codebase implying that if you're deploying a Rails app that it's the top level app. 

Even if that wasn't the case, this will add n times complexity when debugging and creating applications not to mention maintaining the buildpack. My gut says that this extra complexity would harm more than it would help. New developers pulling their hair out when they forget to add a secret file and the build mysteriously fails.

I appreciate the spike, i'm actually surprised it didn't take more code, though I suspect there's some hidden corners that aren't covered (there always are with the buildpack).

The bigger question to can or should we support this, is why do people want this? Is there some really really good use case that simply won't work without your Rails app being in a different folder? You're not the first person to ask for this, but you would be pretty close to the second in the last few years.

shageman:
Thanks for the reply, @schneems! Let me answer the various points you raise

#### 12 factor apps can always be in root
I understand that heroku wants to support 12 factor apps. And while “all apps in root” works, there are ways to violate http://12factor.net/codebase with and without that rule. https://github.com/shageman/sportsball/tree/web_container_dir does not violate it. It is one app, it just is not in the root of the repo.

#### Who would want this?
I want this, because I want people to develop better architected applications. Following https://www.youtube.com/watch?v=WpkDN78P884 - what is the most important part about a Rails app? It is actually not the fact that is it Rails, but rather the domain that is represented in the app (I can however see, how to heroku or any PaaS, the language and framework are much more important then the domain). I am seeking the flexibility to represent that in the codebase.

In the long run, I would also like to have multiple apps in one codebase. Truly shared code would only exist through dependency management (albeit local, using `path:` in the Gemfile). I don’t think this causes any (more) problems with 12 factor. However, the devs ability to maintain larger apps and quickly refactor would be greatly improved.

#### Buildpack complexity
The defaults should obviously work as is. I feel like the ruby buildpack is already unnecessarily complex by clobbering build packs for Ruby and various Rails versions. In that sense, the buildpack violates 12 factor as it runs multiple apps (i.e., supports different app types), but doesn’t specify its dependencies through a dependency manager (The deployment of a Rails 5 app depends on the deployment of a Rails 2 app and Ruby apps).

I realize that I am currently one of few that are asking, but I am trying to change that :)

schneems:
I'll watch the video, though I would recommend against running multiple apps from the same codebase. That factor is 1 codebase to 1 app. Sharing code may seem like a good idea, but in practice i've never seen it done well. Maybe the best example would be travis, they have shared components and they use existing infrastructure (i.e. rubygems and github) to accomplish the code sharing. Sharing logic between apps is best done through well defined interfaces, things like sharing code or database connections creates a huge surface area.

I'm not making an argument that we shouldn't do this due to 12factor, I'm making the argument that I don't want to support it, i think it's a generally bad idea personally, and I also happen to think 12factor supports my case, at least a little bit.

I've not watched the video, so maybe that will help give me some more context. 

> the buildpack violates 12 factor as it runs multiple apps

The buildpack isn't a web app, it's a script that gets executed as a small part in an extremely complicated build process. If you're arguing the buildpack is too complex, I think you're right but for different reasons. It's more like a library or code module, think of it as a build service script. It takes in a well defined input https://devcenter.heroku.com/articles/buildpack-api and outputs a slug ready to be run by a different component.

> The deployment of a Rails 5 app depends on the deployment of a Rails 2 app and Ruby apps

Not sure what you're talking about here? Class inheritance was chosen as an architectural decision for how the buildpack is written. Right decision? I would argue not, but it works and there's too little benefit to re-architecting at this point.

>  it runs multiple apps (i.e., supports different app types)

This is because of legacy. Aspen supported crazy things, bamboo supported different crazy things, cedar (this buildpack) supports most of the crazy things from bamboo, but attempts to do them in a saner way. If we add support for building and running rails apps in the non-root folder it's support we will have to continue for years and years to come.

I'm not saying **no** but i'm saying probably not.

PR #360, opened 2015-04-07
Remove ctrl+C message

cyx:
We're proposing to remove supporting the behavior moving
forward.


schneems:
Heroku won't be removing CTRL+C to cancel builds.

schneems:
To add some context here. The CTRL+C option isn't fully supported across HTTP git so if you're using that, the message is incorrect as CTRL+C won't stop your deploy. This is bad since we're explicitly saying that it will. The Build and Packaging team plans on removing the ability to CTRL+C from SSH git builds soon, removing this message would reflect current status of builds.

cyx:
^ bump :)

schneems:
We need to modify the rake task detection logic so this will show up less. That can happen sometime after railsconf

PR #366, opened 2015-04-24
First stab at cleaning out the .git folders

cyx:


cyx:
Possibly addresses #76 in full.

schneems:
This looks good, can you make this PR from inside the buildpack git repo instead of your own fork? 

```
$ git push git@github.com:heroku/heroku-buildpack-ruby.git clean-dot-git-in-vendor-base
```

If you do that then tests can run with the proper credentials in travis.

PR #372, opened 2015-04-30
Use shell in Detect for speed

### LINKED ISSUE #368: Re-write detect to remove ruby for speed
While it doesn't matter that much for the speed of an individual build, Ruby detect gets executed for EVERY build on the platform that doesn't specify an explicit buildpack. Looks like we could speed this up by using shell or bash:

https://github.com/heroku/heroku-buildpack-ruby/commit/dd75ed563577e305bead895d21c56e704f70d307#commitcomment-10919871

DyegoCosta:
Resolves #368

```shell
# ruby
$ ( for i in {1..1000}; do; bin/detect /ruby-app; done; )  29.98s user 6.32s system 97% cpu 37.266 total

# sh
$ ( for i in {1..1000}; do; bin/detect /ruby-app; done; )  1.32s user 1.43s system 98% cpu 2.789 total
```

### DIFF
--- a/bin/detect
+++ b/bin/detect
@@ -1,11 +1,9 @@
-#!/usr/bin/env ruby
+#!/bin/sh
 
-require 'pathname'
-
-if Pathname.new(ARGV.first).join("Gemfile").exist?
-  puts "Ruby"
+if [ -f $1/Gemfile ]; then
+  echo "Ruby"
   exit 0
 else
-  puts "no"
+  echo "no"
   exit 1
-end
+fi


schneems:
Thanks! I'm running tests on #374 so that travis can execute with env vars. When it passes, i'll merge

DyegoCosta:
@schneems Awesome
I see the build failed, is the error related to this patch? If there is something I can fix let me know

schneems:
It's a rate limiting issue with the test account, I need to fix it.



—

hone:
Thanks for the patch!

PR #375, opened 2015-05-01
First stab at cleaning out the .git folders

cyx:


schneems:
Thanks!

dpiddy:
:clap:

rodrigoddalmeida:
Sorry to be a bit pedant here, but wouldn't something like [this](https://github.com/rodrigoddalmeida/heroku-buildpack-ruby/blob/master/lib/language_pack/ruby.rb#L606) be clearer?

I don't really know how much it'll change in terms of performance.

PR #377, opened 2015-05-02
Allow specifying a bundler version via an ENV variable

segiddins:
Would allow users such as https://github.com/bundler/bundler/issues/3609 to use a newer bundler version when necessary.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -14,7 +14,7 @@ class LanguagePack::Ruby < LanguagePack::Base
   NAME                 = "ruby"
   LIBYAML_VERSION      = "0.1.6"
   LIBYAML_PATH         = "libyaml-#{LIBYAML_VERSION}"
-  BUNDLER_VERSION      = "1.7.12"
+  BUNDLER_VERSION      = env("BUNDLER_VERSION") || "1.7.12"
   BUNDLER_GEM_PATH     = "bundler-#{BUNDLER_VERSION}"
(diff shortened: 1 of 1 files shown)


schneems:
We won't be allowing the ability to specify bundler version.

segiddins:
@schneems can I ask why this is not an ability you want to grant to people? Given how far behind on bundler updates heroku are, I think its completely reasonable to allow users to opt in to using the new features in the latest bundler versions.

schneems:
> why this is not an ability you want to grant to people?

At the end of the day it comes down to minimizing surprise and risk.

We can't arbitrarily use any version of bundler, we have to explicitly build and make the versions we do use available on S3. So it's very likely that you could set to a version we don't support and your application crashes, you open a support ticket.

If a user is on an older version of bundler that has a bug and they don't know it, now we must have the technical debt of being aware of all common gotchas of all versions of bundler ever built and share them across every supporter and Ruby engineer. 

If you manually peg to a version, we can't upgrade you. I.e. when there are security releases or when we want to roll up to optimizing the buildpack to use new features of bundler, we can't do that if you've locked to an older version.

I understand the want to specify a bundler version, but to date it remains an edge case that would most likely create more confusion and problems and debt across the platform than it would be worth. 

I'm happy to let you know why, also you should know we're pretty firm on this. It's not up for debate. If you really want this ability, you can always fork the ruby buildpack to run any custom code you want.

segiddins:
@schneems thanks for explaining :+1:

PR #378, opened 2015-05-04
Update Bundler version

indirect:
This updates the version of Bundler used to the current newest version, 1.9.6.

We've been continuing to backport bugfixs to the 1.7.x series just for Heroku, but unless Heroku joins Ruby Together I don't have enough time available to make sure that continues to happen. In addition, there are _many_ features that are simply unavailable to Heroku users who want or need to use them, including the ability to keep Gem server credentials out of source control.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -14,7 +14,7 @@ class LanguagePack::Ruby < LanguagePack::Base
   NAME                 = "ruby"
   LIBYAML_VERSION      = "0.1.6"
   LIBYAML_PATH         = "libyaml-#{LIBYAML_VERSION}"
-  BUNDLER_VERSION      = "1.7.12"
+  BUNDLER_VERSION      = "1.9.7"
   BUNDLER_GEM_PATH     = "bundler-#{BUNDLER_VERSION}"
   DEFAULT_RUBY_VERSION = "ruby-2.0.0"
   RBX_BASE_URL         = "http://binaries.rubini.us/heroku"


schneems:
I built bundler 1.9.6 and put it our our S3 bucket so you should now be able to deploy with your branch. I would like to get to a newer version, thanks for the ping.

Any idea on large issues or changes that might affect users even if they're so-far un-reproduceable edge cases?

indirect:
I've bumped my PR up to 1.9.7 because it contains the tmdir fix that @hone helped with.

@schneems #3534 is the only edge-case-y bug that I know about, but it was fixed by running `bundle clean`, and Heroku does that automatically.

geoffharcourt:
This would be awesome. 1.7.x doesn't allow the use of blocks in the Gemfile to mark the sources for gem groups if you have multiple gem sources, so you either have to specify a source individually for every gem, or you have to just trust that your gem sources won't hijack gems from other sources (you can order the sources, but that wouldn't work with more than two sources or if one source wasn't always the "winner").

indirect:
@geoffharcourt I don't think that's true. 1.7 is the release that introduced source selection, including with blocks.

geoffharcourt:
@indirect, just tested it and you're right. I tried to use blocks on my app a couple months ago and it didn't work, and my (very cursory) research led me to believe it was because Bundler was pre-1.8. Apologies for the mistake, and thanks for the info.

schneems:
Doesn't look like 1.9.7 is released yet.

```
remote: ERROR:  Could not find a valid gem 'bundler' (= 1.9.7) in any repository
remote: ERROR:  Possible alternatives: bundler
```

Can you ping me when that happens?

indirect:
@schneems ugh, I guess the release task failed and I didn't notice. It's out now.

schneems:
Running tests over at #385

schneems:
Closed with #385

schneems:
This is merged and deployed to production. Thanks! Hopefully we won't have to rollback.

PR #381, opened 2015-05-06
Added override config var to build-time jruby-opts

jkutner:


schneems:
`ENV` is going to be environment variables provided by codon `env` is going to be environment variables provided by the user. We don't source user's environment variables and only inject them on certain commands since otherwise it can mess with the buildpack. If you want this to be user settable use lowercase `env`

PR #384, opened 2015-05-07
Added a redeploy step to JRuby test to cover jruby_opts at build time

jkutner:
Added test for jruby_opts output during build

Added jruby_opts to blacklist of env vars

Only set jruby_opts for build if user sets jruby_build_opts

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -246,11 +246,7 @@ EOF
 #{set_jvm_max_heap}
 echo #{default_java_tool_options}
 SHELL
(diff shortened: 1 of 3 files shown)


hone:
These two tests are failing:

```
rspec ./spec/rubies_spec.rb:83 # Ruby Versions should deploy jdk 8 on cedar-14 by default
rspec ./spec/rubies_spec.rb:100 # Ruby Versions should deploy jruby 1.7.16.1 (jdk 7) properly on cedar-14 with sys props file
```

PR #385, opened 2015-05-11
Update Bundler version

schneems:
This updates the version of Bundler used to the current newest version, 1.9.7.

We've been continuing to backport bugfixs to the 1.7.x series just for Heroku, but unless Heroku joins Ruby Together I don't have enough time available to make sure that continues to happen. In addition, there are _many_ features that are simply unavailable to Heroku users who want or need to use them, including the ability to keep Gem server credentials out of checked in files.

### DIFF
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -92,8 +92,11 @@ class LanguagePack::Helpers::BundlerWrapper
             }
       command = "bundle platform --ruby"
 
+      # Silently check for ruby version
       output  = run_stdout(command, user_env: true, env: env)
-      raise GemfileParseError.new(run(command, user_env: true, env: env)) unless $?.success?
+
+      # If there's a gem in the Gemfile (i.e. syntax error) emit error
+      raise GemfileParseError.new(run("bundle check", user_env: true, env: env)) unless $?.success?
       if output.match(/No ruby version specified/)
         ""
       else
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -14,7 +14,7 @@ class LanguagePack::Ruby < LanguagePack::Base
   NAME                 = "ruby"
   LIBYAML_VERSION      = "0.1.6"
   LIBYAML_PATH         = "libyaml-#{LIBYAML_VERSION}"
-  BUNDLER_VERSION      = "1.7.12"
+  BUNDLER_VERSION      = "1.9.7"
   BUNDLER_GEM_PATH     = "bundler-#{BUNDLER_VERSION}"
   DEFAULT_RUBY_VERSION = "ruby-2.0.0"
   RBX_BASE_URL         = "http://binaries.rubini.us/heroku"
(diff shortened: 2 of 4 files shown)


indirect:
[This failure](https://travis-ci.org/heroku/heroku-buildpack-ruby/builds/62151170#L590) is because it's expecting the text "Your Bundle is complete!", but getting the text "Bundle complete!".

[This entire build](https://travis-ci.org/heroku/heroku-buildpack-ruby/builds/62151192) failed because of the Heroku API returning errors.

AFAICT, that means the build is functionally passing?

Not relevant for Heroku, but I also just released 1.9.8 with a fix for sudo mode, which broke in 1.9.7.

schneems:
Thanks, I updated that test. There's also another failure

```
     Failure/Error: expect {LanguagePack::RubyVersion.new(@bundler.ruby_version)}.to raise_error(error_klass, /#{Regexp.escape(bundle_error_msg)}/)
       expected LanguagePack::Helpers::BundlerWrapper::GemfileParseError with message matching /Zomg\ der\ was\ a\ problem\ in\ da\ gemfile/, got #<LanguagePack::Helpers::BundlerWrapper::GemfileParseError: There was an error parsing your Gemfile, we cannot continue
       There was an error in your Gemfile, and Bundler cannot continue.
       > with backtrace:
         # /home/travis/build/heroku/heroku-buildpack-ruby/lib/language_pack/helpers/bundler_wrapper.rb:96:in `block in ruby_version'
```

We are essentially testing that if bundler throws an error that we want to make sure that error text is available in the message we give to the user. 

When I run `bundle install ` manually, I get the right thing

```
/Users/richardschneeman/Documents/projects/heroku-buildpack-ruby/repos/bundler/bad_gemfile_on_platform/Gemfile:3:in `eval_gemfile': Zomg der was a problem in da gemfile (RuntimeError)
       /Users/richardschneeman/.gem/ruby/2.2.2/gems/bundler-1.9.8/lib/bundler/dsl.rb:32:in `instance_eval'
       /Users/richardschneeman/.gem/ruby/2.2.2/gems/bundler-1.9.8/lib/bundler/dsl.rb:32:in `eval_gemfile'
       /Users/richardschneeman/.gem/ruby/2.2.2/gems/bundler-1.9.8/lib/bundler/dsl.rb:10:in `evaluate'
```

Not sure why the test is failing there

schneems:
Figured it out, this error is on the `bundle check` command. If there's an error in the Gemfile Bundler no longer emits the error message :(

This is a regression.

I can work around it in the short time by changing the command that emits error information to `bundle check`.

indirect:
@segiddins how do you feel about at least guaranteeing that the exception message gets printed as part of the new error formatter?

segiddins:
@schneems I'll look into this regression when I get home from work, and I'll make sure to add a test in bundler.

segiddins:
Fwiw, this issue doesn't exist on bundler master.

segiddins:
So, this regression was caused by https://github.com/bundler/bundler/commit/2066de222384ad74ad96f6b863510541ac66cc51 -- the error message is now `warn`ed, which is why it's not showing up on `bundle platform`. Again, this has been 'fixed' on Bundler master because I've completely re-worked exception reporting since the release of 1.9

indirect:
Ah, sorry for pointing at you! Thank for investigating. @schneems, how badly do you need runtime error messages from Gemfiles to be printed? The next minor release (already available as a pre) will have much better Gemfile error explanations.

schneems:
I added an in place fix. It's not a blocker. We can merge tomorrow. See how it goes on master for a few days. If no major issues probably deploy early next week.

schneems:
Merged, thanks for your help! :heart:

PR #386, opened 2015-05-15
New Ruby step: create config/*.yml files from config/*.yml.sample

tinogomes:


tinogomes:
@schneems  just closed, without justify? :(

schneems:
You opened a pull request with no comment or justification.

tinogomes:
I thought the title was self explaining, but ok...

description inserted!

schneems:
A pull request with only a title is like closing an issue with the comment "i am closing this issue" it explains the "what" but not the "why". I have no idea what you're thinking, why we would want this, what your motivation or background is for submitting the PR. You essentially dumped a bunch of code on us, asked me to maintain it, and then left without a word. 

PRs with no comments are a pet-peeve of mine, and i've not had a terribly good morning, sorry for being short.

## Why we don't want this

At Heroku, we don't like people putting secrets in source control, we have an entire factor dedicated to it http://12factor.net/config.

We also don't like config yml files. We want people to use environment variables. Rails out of the box supports DATABASE_URL for configuration (so you don't need a database.yml). It also defaults production to `SECRET_KEY_BASE` which is read by default. Using YML files encourages people to do things like create different environments which encourages really bad practices read https://devcenter.heroku.com/articles/deploying-to-a-custom-rails-environment for more information. These create more problems than they solve. It's really frustrating to lose half a day of work because you didn't notice that your app was running in the wrong environment.

I see the problem you're trying to solve, however I don't think this is the place to solve it. Instead we should be pushing out to the community to eliminate the need for yml based config, or at bare minimum enabling these yml files to read in env vars via ERB so that no one has to hard code in credentials into git. Once your credentials are in source control, you should already consider them compromised.

Essentially yml files for config are an anti-pattern and we don't want to encourage their use.

Thank you for adding in context to your code. I hope that this context of why I closed the PR makes more sense now.

## PR

I appreciate the PR, thank you for taking the time to fork and look through our code. I'm not accepting this , but it's not a total loss. You got some good experience working with the buildpack, so if you ever want to modify anything in the future you can do it even easier! Also since you can run a custom buildpack on Heroku, you can actually deploy your own projects using your own buildpack! Though if this is something you're planning on doing long term, I would recommend porting this code over to a micr-buildpack and using it with the multi-builpack to have it run before the ruby buildpack. Otherwise it's really hard to stay up-to-date and constantly be merging features back into your buildpack. If you're curious about this, I can give you some more info.

This PR is actually a great PR, it's got code, tests, and docs. I really do appreciate the submital, it's a good chance to share a bit of our philosophy, sorry for taking out my bad morning on you. Thanks for the PR :heart:

tinogomes:
I usually apply 12factor rules. My fault about to forget description on PR too. I made it on a over time work and I was sleepy.

In my case, sample config files was following 12factor config rules, but I believe I want to replace my development environment to use 12factor config rules too.

And about your bad morning, everyone have one. Don't worry, be happy! :heart: 

Best!

waiting-for-dev:
I landed in this pull request looking for a workaround to the problem stated in this issue in Sidekiq:

https://github.com/mperham/sidekiq/issues/503#issuecomment-258813417

In short: a lot of times it is needed to change the default ActiveRecord pool size (5). Currently, heroku recommended way is to push a `config/database.yml` file like the following:

```
production:
  pool: 25
```

From https://devcenter.heroku.com/articles/rails-database-connection-behavior#configuring-connections-in-rails-4-1:

> While the default connection information will be pulled from DATABASE_URL, any additional configuration options in your config/database.yml will be merged in. The idea is that DATABASE_URL only holds connection information, but not behavior information. You can still configure Active Record behavior such as pool setting [...]

This goes against the recommendations for tweve-factor applications. From https://12factor.net/config:

> An app’s config is everything that is likely to vary between deploys (staging, production, developer environments, etc). This includes:

> · Resource handles to the database, Memcached, and other backing services
> [...]

> Another approach to config is the use of config files which are not checked into revision control, such as config/database.yml in Rails. This is a huge improvement over using constants which are checked into the code repo, but still has weaknesses: it’s easy to mistakenly check in a config file to the repo; there is a tendency for config files to be scattered about in different places and different formats, making it hard to see and manage all the config in one place. Further, these formats tend to be language- or framework-specific.

> The twelve-factor app stores config in environment variables (often shortened to env vars or env). Env vars are easy to change between deploys without changing any code; unlike config files, there is little chance of them being checked into the code repo accidentally; and unlike custom config files, or other config mechanisms such as Java System Properties, they are a language- and OS-agnostic standard.

> [...]

These same reasons have been applied to close this pull request:

> We also don't like config yml files. We want people to use environment variables. Rails out of the box supports DATABASE_URL for configuration (so you don't need a database.yml).

> [...] we should be pushing out to the community to eliminate the need for yml based config [...]

> Essentially yml files for config are an anti-pattern and we don't want to encourage their use. [...]

So,  I think that in this case heroku is not following heroku twelve-factor recommendations. I completely agree with them, so I think a way to deal with connection pool size through environment variables should be provided. If not, then at least this pull request should be accepted (if you want, limiting its scope to the `config/database.yml` file).

Thank you!

* References https://github.com/heroku/heroku-buildpack-ruby/pull/17

schneems:
I'm not accepting this pull request

waiting-for-dev:
Hey @schneems !

But don't you think that current recommended way of changing ActiveRecord connection pool size breaks twelve factor rules? I agree that this pull request is not the best option. Instead, there should exist a way to change connection pool size through environment variables.

Thinking more about it, I think that this should be a native Rails feature. But, at least, I think heroku should not recommend pushing `config/database.yml` for the reasons you stated.

By the way, just in case anybody is interested, currently I'm changing the connection pool size in an initializer. It is far from ideal, because it adds some initialization overhead. But I prefer it over pushing `config/database.yml`.

Cheers

eugeneius:
> I think that this should be a native Rails feature

http://guides.rubyonrails.org/configuring.html#configuring-a-database:

> This same information can be stored in a URL and provided via an environment variable like this:
>
> ```ruby
> puts ENV['DATABASE_URL']
postgresql://localhost/blog_development?pool=5
> ```

It already is! 😄

waiting-for-dev:
It is true, I had forgotten :D 

The problem with heroku is that it recommends not modifying `DATABASE_URL` because it can change it at any time. So I guess heroku should build `DATABASE_URL` taking the `?pool=` part from another configuration variable.

schneems:
If you want a 12 factor active record connection get on Rails core and make it happen.

waiting-for-dev:
@schneems , as @eugeneius said, ruby on rails currently supports setting active record connection pool size in a [12 factor compliance manner](http://edgeguides.rubyonrails.org/configuring.html#configuring-a-database), but it can't be used in heroku because `DATABASE_URL` may change at any time.

waiting-for-dev:
Another option would be heroku respecting what it is after `?` when changing `DATABASE_URL`, because other things besides the pool can be configured there.

schneems:
Pretty sure it does get respected still it's better to be safe: open up a ticket help.heroku.com and search "DATABASE_URL". I'm not on that team and I don't know off hand.

waiting-for-dev:
Thanks, I'll check with them.

PR #390, opened 2015-07-02
Fix the link to Getting Started with Ruby [ci skip]

joneslee85:
I know, Ruby is not yet replaced by NodeJS yet.... :trollface:


hone:
Thanks!

PR #391, opened 2015-07-02
Allow path with spaces to be detected as Ruby

DyegoCosta:
Without the quotes it won't work for path with spaces

### DIFF
--- a/bin/detect
+++ b/bin/detect
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if [ -f $1/Gemfile ]; then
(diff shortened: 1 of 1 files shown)


hone:
I don't think we'll have path spaces, but good to have for people using it outside of heroku.

DyegoCosta:
@hone cool. I didn't realize that issue until now or it would be included in that other PR

hone:
@DyegoCosta no worries. I saw it, but didn't think it'd be a big deal. Thanks for the patch!

PR #397, opened 2015-07-24
Update readme with banner image. 

mattgraham:
Creating images for all the Heroku official buildpacks. Thoughts on this?


hone:
:+1:

mattgraham:
Updated, already...
ruby

mathias:
Any ideas why this failed? The build log is awfully noisy and I'm not familiar with this test suite.

mattgraham:
@hone ^^

schneems:
I think your branch doesn't live on heroku/heroku-buildpack-ruby so Travis won't run tests with the encrypted env vars as a security measure (otherwise you could submit a PR that is just `puts ENV` to grab all our secrets. This is a cosmetic change you can either [ci skip] it or merge in as is. Thanks for the work, looks great. :+1:

PR #401, opened 2015-07-28
adding hatchet setup information to readme

orlando:
# Description

This PR adds information on how to setup hatchet to the heroku-buildpack-ruby README.md file. This is required to run buildpack tests correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ it does not parallelize tests within a test file. To run the tests: clone the re
 $ bundle exec hatchet install
 ```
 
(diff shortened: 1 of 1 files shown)


schneems:
Thanks for the PR. It should still work without manually setting a token.

It's a good step though.

orlando:
@schneems Some tests where returning 422 responses. Once I configured hatchet, those turn green.

I still have 1 failing test, but I don't know what it causing it. If you can provide insight so I can fix it for both this PR and https://github.com/heroku/heroku-buildpack-ruby/pull/402, that would be great.

Thanks!

PR #402, opened 2015-07-29
[Rails 4] Don't overwrite public/assets files when loading cache

orlando:
# Background

I'm using a gulp based assets pipeline along with Sprockets. I'm only using Sprockets because I have gems that require it (rails-admin for example). `gulp-rev` generates a manifest file before running the ruby buildpack

The problem comes when we cache `public/assets`, and then load it, since this overwrites the manifest created by `gulp-rev` with an old version. Right now the only way to workaround this is using `heroku repo` plugin and running `heroku repo:purge_cache`

# Description

This PR changes how `public/assets` cache is loaded. Instead of overwriting all the files, we only copy the ones that are missing.

# Questions

I cannot think a way that this will cause problems with normal Rails apps, since `public/assets` should be only populated by `rails assets:precompile`, and if it's run on Heroku, there should not be any files before running this process, but of course, let me know if you see issues happening by merging this PR.

# Related Issues

* https://github.com/vigetlabs/gulp-rails-pipeline/issues/6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master
 
+* Cached asset file should never take precedent over existing file (#402)
+
 ## v138 (5/19/2015)
 
 * Bump bundler to 1.9.7 [Bundler changelog](https://github.com/bundler/bundler/blob/master/CHANGELOG.md#196-2015-05-02) (#378)
--- a/lib/language_pack/cache.rb
+++ b/lib/language_pack/cache.rb
@@ -42,13 +42,18 @@ class LanguagePack::Cache
     copy (@cache_base + path), dest
   end
 
+  def load_without_overwrite(path, dest=nil)
+    dest ||= path
+    copy (@cache_base + path), dest, '-a -n'
+  end
+
   # copy cache contents
   # @param [String] source directory
   # @param [String] destination directory
-  def copy(from, to)
+  def copy(from, to, options='-a')
     return false unless File.exist?(from)
     FileUtils.mkdir_p File.dirname(to)
-    system("cp -a #{from}/. #{to}")
+    system("cp #{options} #{from}/. #{to}")
   end
 
   # copy contents between to places in the cache
--- a/lib/language_pack/rails4.rb
+++ b/lib/language_pack/rails4.rb
@@ -79,7 +79,7 @@ WARNING
 
         topic("Preparing app for Rails asset pipeline")
 
-        @cache.load public_assets_folder
+        @cache.load_without_overwrite public_assets_folder
         @cache.load default_assets_cache
 
         precompile.invoke(env: rake_env)
--- a/spec/rails41_spec.rb
+++ b/spec/rails41_spec.rb
@@ -24,4 +24,27 @@ describe "Rails 4.1.x" do
       end
     end
   end
+
+  it "should not overwrite existing files with cached files" do
+    string = SecureRandom.hex(13)
+    new_string = SecureRandom.hex(13)
+
+    Hatchet::Runner.new("rails41_scaffold").deploy do |app, heroku|
+      # First Deploy
+      `mkdir public/assets`
+      `echo #{string} > public/assets/file.txt`
+      `git add -A; git commit -m 'adding file.txt'`
+      app.push!
+
+      # Second Deploy
+      `echo #{new_string} > public/assets/file.txt`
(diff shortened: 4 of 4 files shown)


rowanoulton:
+1 would like to see this merged, am seeing the same issues running a gulp-based asset pipeline on Heroku.

castrolem:
+1, this is an issue also for us

greypants:
This looks great! Thanks for taking the time to figure it out! Can you add some info to the README around this, so people implementing on their existing codebases are aware?

orlando:
@greypants the idea is to get it merged with upstream. If that doesn't happen, I'll change the name of the repo and maintain the fork.

schneems:
I'm thinking of ways we could test this, right now I think we could maybe deploying with an app that writes the current time to a public/assets/file.html when deployed. We can then curl the contents of the file, then re-deploy and make sure that the contents have changed.

@hone can you give this PR a review? I believe it's safe, I can't imagine a scenario where someone was relying on the old behavior to have the cache over-write something that was currently on disk, maybe you know of one.

@orlando thanks for the PR, I wasn't aware of this failure condition. Can you squash to one commit and remove the `.gitignore` entry?

orlando:
@schneems Sure I'll do that.

If you can point me on how to do that testing (deploy for example), I'm willing to help.

Thanks

schneems:
We have a repo full of example apps https://github.com/sharpstone you can fork one. From there add it to hatchet.json in the root, and you can write a test like this: https://github.com/heroku/heroku-buildpack-ruby/blob/a7a1942dea5407726417908d50352ab67d021649/spec/rails41_spec.rb#L13-L19

orlando:
@schneems Cool, I'll implement a test for it a update the PR, probably today EOD.

Also, this patch is only for Rails 4. I don't if you guys want to implement this behaviour for other versions of rails or other ruby frameworks.

schneems:
Rails 4 introduced asset caching, it's the only caching we support. All later versions of Rails in the buildpack inherit from Rails 4.

orlando:
@schneems cool that makes sense.

Thanks!

schneems:
I always forget, can you add a changelog entry? Follow convention, can put it in the "master" section, something like 

```
* Cached asset file should never take precedent over existing file (#402)
```

orlando:
PR updated with changelog entry

hone:
I can take a look sometime this week or next. I'm a bit off/on while in
Europe.

orlando:
test added https://github.com/heroku/heroku-buildpack-ruby/pull/402/files#diff-c4f2de8bcab48b98f90cc4168d70ee47R28

schneems:
Thanks, that was super quick. The tests won't run for branches that are against a different master since we need secure keys. I copied your changes over, and i'm running them here: https://github.com/heroku/heroku-buildpack-ruby/pull/406

orlando:
@schneems no problem, thanks for taking the time to review this PR

schneems:
Merged this into master. Quite a few people use master to deploy, so we treat it as a stable branch. We also like to keep changes there for at least a few days to see if anyone has a problem with a change before doing a proper deploy. @hone let me know if you think we need to revert when you get the chance. 

Thanks for the contribution :rocket: :heart:

orlando:
:100:

whit537:
> We also like to keep changes there for at least a few days to see if anyone has a problem with a change before doing a proper deploy.

@schneems Out of curiosity, what is "a proper deploy"? I read the [buildpacks doc](https://devcenter.heroku.com/articles/buildpacks) and searched the [build forum](https://discussion.heroku.com/category/build) and am not finding a solid answer. I also searched the [branches in this repo](https://github.com/heroku/heroku-buildpack-ruby/branches/all) (you've got a lot stale branches ;), but didn't find anything that looked like an official `deploy` branch.

schneems:
We deploy based off of master. We have an internal procedure and service for "deploying" a buildpack. Anything on master is assumed to be deployable. This change was deployed quite some time ago: https://github.com/heroku/heroku-buildpack-ruby/blob/master/CHANGELOG.md#v139-8312015 

When we deploy we also cut a release https://github.com/heroku/heroku-buildpack-ruby/releases

whit537:
> When we deploy we also cut a release https://github.com/heroku/heroku-buildpack-ruby/releases

Cool, thanks. :-)

Issue #364, opened 2015-04-23
Don't write database.yml for ActiveRecord 4.1+


damonmorgan:
Commit https://github.com/heroku/heroku-buildpack-ruby/commit/41dd02b147674e6bb793e9ddb3df47601eae0655 stopped writing `config/database.yml` in Rails 4.1 apps but it would actually be better if it stopped it writing for any apps using ActiveRecord 4.1+ (for example I have a Sinatra app using ActiveRecord 4.2 where I would like to be able to set additional options like pool size in my config/database.yml).

schneems:
Yep, I agree, it's on the radar but hasn't been a priority. Could you stub out a minimal patch for me with an example sinatra app?

afeld:
Closed by #404, yes?

shaqq:
Out of curiosity, why would we need to overwrite the `config/database.yml` for non-Rails apps? It seems like the `config/database.yml` is overwritten because of Railties, no? I can't see how a Sinatra app w/ ActiveRecord would be affected.

So there's this gem: https://github.com/janko-m/sinatra-activerecord/blob/master/lib/sinatra/activerecord.rb

But that doesn't affect us since we deploy [Napa](https://github.com/bellycard/napa) apps and have our own way of setting up ActiveRecord

schneems:
Turns out there were some big caveats. I added a note to the changelog entry https://devcenter.heroku.com/changelog-items/709

shaqq:
@schneems Could you point me to the relevant section in that changelog entry? I can't seem to find it...

PR #403, opened 2015-07-29
Don't write database.yml if using ActiveRecord >= 4.1

damonmorgan:
Patch for https://github.com/heroku/heroku-buildpack-ruby/issues/364

An example app for testing can be found at https://github.com/damonmorgan/activerecord41_scaffold

### DIFF
--- a/hatchet.json
+++ b/hatchet.json
@@ -29,7 +29,8 @@
     "sharpstone/ruby_193_jruby_17161_jdk8",
     "sharpstone/jruby-minimal",
     "sharpstone/empty-procfile",
-    "sharpstone/bad_ruby_version"
+    "sharpstone/bad_ruby_version",
+    "sharpstone/activerecord41_scaffold"
   ],
   "rack": [
     "sharpstone/default_ruby",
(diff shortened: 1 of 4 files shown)


hone:
We created an empty method in Rails41 class so it wouldn't. I believe your patch actually undoes that work since pre 4.1, we create a database.yml.

damonmorgan:
Yes, but that should be redundant code now as [this change](https://github.com/heroku/heroku-buildpack-ruby/pull/403/files#diff-3fc66e13e389ff4d15c7ca3ddb86464eR632) should mean that any application using ActiveRecord 4.1+ (including Rails 4.1+, but not limited to only Rails now) shouldn't have the database.yml file written now? I left the tests in for Rails 4.1 so they should still pass.

schneems:
Tests are now passing over at https://github.com/heroku/heroku-buildpack-ruby/pull/404

I modified it to preserve the accidental functionality we had that would write a config/database.yml before when you don't have activerecord. We can consider killing that later, but It is easier to be backwards compatible than to deal with broken apps in support tickets for now.

schneems:
merged in #404. It will live on master for awhile to see if there are any problems/regressions before it gets deployed. Thanks for the help :heart:

damonmorgan:
Thanks

PR #404, opened 2015-08-03
Running tests for #403

schneems:


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Master
 
 * Cached asset file should never take precedent over existing file (#402)
+* Do not write `database.yml` when using active record >= 4.1 (previously we only detected >= Rails 4.1) (#403)
```
afeld: I'm assuming these weren't meant to be the same versions?

schneems: Versions are correct. We switched from checking "rails" to checking "active record"

afeld: Heh, didn't see that difference. Thanks!

PR #408, opened 2015-08-17
Clarify top is showing version of bundler

zunda:
Sometimes users are confused with the version number 1.9.7 to be from something other than bundler so it may be better to change

```
remote: -----> Installing dependencies using 1.9.7
```

to

```
remote: -----> Installing dependencies using bundler 1.9.7
```


schneems:
Seems good, thanks!

zunda:
ty!

PR #412, opened 2015-08-27
Added a JRuby specific error message if version is not found

jkutner:
I tried spliting this up so that the common parts were only in code once, and not repeated like this... but honestly it just became a mess.  So i think it might be better just to duplicate the message like this.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -368,12 +368,21 @@ WARNING
 
     true
   rescue LanguagePack::Fetcher::FetchError => error
-    message = <<ERROR
+    if ruby_version.jruby?
+      message = <<ERROR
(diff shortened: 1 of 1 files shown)


schneems:
Tests

```
body: "No such stack.
```

ಠ_ಠ

I swear, I spend 2x the amount of time fixing random errors like this than fixing bugs on the buildpack

schneems:
I'm :+1:  on this. Any thoughts @hone ?

hone:
:+1: 

For that error message, is it now `cedar-10`?

hone:
Build now [passes](https://travis-ci.org/heroku/heroku-buildpack-ruby) and merged with 8150fb62ca6c24100fcccc54cc5240e09dd8a1c0

PR #413, opened 2015-08-29
[RFC] Allow projects in subdirectories

beedub:
We had the use-case of deploying applications that were in a subdirectory in the project. The changes below were required to get it working.

A couple of questions:

1. is this something the main heroku buildpack would want? if so, then how does this PR look?
1. why is caching `.bundle/config` is necessary?

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -9,8 +9,8 @@ require "language_pack/shell_helpers"
 
 begin
   LanguagePack::Instrument.trace 'compile', 'app.compile' do
+    LanguagePack::ShellHelpers.initialize_env(ARGV[2])
     if pack = LanguagePack.detect(ARGV[0], ARGV[1])
-      LanguagePack::ShellHelpers.initialize_env(ARGV[2])
       pack.topic("Compiling #{pack.name}")
       pack.log("compile") do
         pack.compile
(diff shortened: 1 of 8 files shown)


schneems:
This has been proposed a number of times. The concept has quite a few problems and very few benefits. It's not something we can safely support. If you want this maybe you can either make a micro buildpack that works with heroku-buildpack-multi to change to the directory you need, or you can maintain your own fork of the ruby buildpack (I recommend against this unless you're getting paid to do it, you would have to pull from master on a daily basis, handle merge conflicts, and there's no guarantee we won't break the functionality you added, also custom forks of buildpacks aren't officially supported by help.heroku.com).

> why is caching .bundle/config is necessary?

I don't think it is, but this was written before my time. There's quite a  large amount of cruft we leave in as changing it might break something someone was relying on. When you are deploying millions of Ruby apps a 1 in a thousand edgecase happens pretty frequently. We're generally pretty conservative with our changes.

schneems:
Here's a previous PR https://github.com/heroku/heroku-buildpack-ruby/pull/357#issuecomment-88946959

beedub:
Fair points.

> If you want this maybe you can either make a micro buildpack that works with heroku-buildpack-multi to change to the directory you need

This sounds pretty interesting. I'm going to research this presently, but if you wouldn't mind elaborating on this, then I would appreciate it. It's not immediately obvious how this would work based on the buildpack docs that I've seen

PR #422, opened 2015-09-29
Add support for custom JDK install

jkutner:
**Do not merge this until after Oct. 10, 2015**

This allows users to set custom JDK tarballs by setting a config var thusly:

```
$ heroku config:set JDK_URL_1_8="http://lang-jvm.s3.amazonaws.com/jdk/openjdk1.8.0_51-cedar14.tar.gz"
```

This is needed to revert to previous version of the JDK if there is problem with an update.

I know it seems hokey to require the `http://lang-jvm.s3.amazonaws.com/jdk` prefix and then remove it. But this way it ensures parity with the [jvm-common buildpack](https://github.com/heroku/heroku-buildpack-jvm-common) and uses the same `Fetcher` class.


jkutner:
Closing in favor of #423. But leaving this branch around until we migrate customers off of it.

PR #423, opened 2015-09-30
Added support for custom JDK versions

jkutner:
**Do not merge this until after Oct. 10, 2015**

This allows users to set custom JDK versions in the `system.properties` file or define custom tarball URLs by setting a config var.

The supported JDK versions can be set in the `system.properties` like so:

```
java.runtime.version=1.8.0_51
```

The custom URL can be set with an env var thusly:

```
$ heroku config:set JDK_URL_1_8="http://lang-jvm.s3.amazonaws.com/jdk/openjdk1.8.0_51-cedar14.tar.gz"
```
This is needed to revert to previous version of the JDK if there is problem with an update.

The env provides parity with the jvm-common buildpack. The custom version in `system.properties` will be a new feature in jvm-common

### DIFF
--- a/lib/language_pack/helpers/jvm_installer.rb
+++ b/lib/language_pack/helpers/jvm_installer.rb
@@ -4,11 +4,12 @@ class LanguagePack::JvmInstaller
   include LanguagePack::ShellHelpers
 
   SYS_PROPS_FILE  = "system.properties"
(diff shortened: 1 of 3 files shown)


jkutner:
Please merge this when ready. I'm back and able to support.

PR #425, opened 2015-10-08
Don't force the postgresql addons on Rails apps

zimbatm:
This will revert to the global ruby default where the addon is inserted only
if the app has a 'pg' gem in it's Gemfile.

A rails app which uses an external database like RDS will get it's
config/database.yml overwritten because of the PG addon and this will break
the app. Combined with the fact that the new pipeline feature and the app.json
format has no way of disabling the default addons, it means that new review
apps always get broken by default.

### DIFF
--- a/lib/language_pack/rails2.rb
+++ b/lib/language_pack/rails2.rb
@@ -70,12 +70,6 @@ private
     end
   end
 
-  # most rails apps need a database
-  # @return [Array] shared database addon
-  def add_dev_database_addon
-    ['heroku-postgresql:hobby-dev']
-  end
-
   # sets up the profile.d script for this buildpack
   def setup_profiled
     super


schneems:
Upgrade to rails 4.1 or higher and we don't over write the database.yml anymore that should solve your problem.

zimbatm:
Hmm [you're right](https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/ruby.rb#L651). I wonder why I have been seeing the issue. I'm currently using:
* Rails 4.2.3
* cedar-14
* ruby 2.2.2

zimbatm:
```ruby
require "bundler"
Bundler::LockfileParser.new(File.read("Gemfile.lock")).specs.each_with_object({}) {|spec, hash| hash[spec.name] = spec }["activerecord"].version >= Gem::Version.new('4.1.0.beta1')
# => true
```

schneems:
Can you open up a support ticket with your app name? Go to https://help.heroku.com and search for "review apps".

zimbatm:
Thanks, waiting on support.

I still think this PR is valid though. Is it worth me pursuing it or should I just close it ?

schneems:
We can't merge this, it would break too much stuff. Your use case is still valid and we should figure how to get it working. Postgres by default should be a convince to those who want it and silently there for those that don't. Let's figure out why your app is failing.

I'm flying out today, might be a bit before I can respond. Can you give me your support ticket number so I can follow up?

zimbatm:
Thanks a lot, here's the link to the issue: https://help.heroku.com/tickets/288549

I'll close that issue then.

PR #427, opened 2015-10-15
Use non default node version for tests

schneems:
Previously we were checking for a custom version that coincidentally matched the default version of node that the ruby buildpack was using. This was a mistake and not noticed when the default version was rev-d. The fixture has been updated to use 4.1.2.


hunterloftis:
:+1:

PR #428, opened 2015-10-28
Added automatic heap settings for IX dyno

jkutner:


lib/language_pack/ruby.rb
```diff
@@ -244,7 +247,7 @@ def default_jruby_opts
   # default JAVA_TOOL_OPTIONS
   # return [String] string of JAVA_TOOL_OPTIONS
   def default_java_tool_options
-    "-Xmx${JVM_MAX_HEAP:-\"384\"}m -Djava.rmi.server.useCodebaseOnly=true"
```
hone: Do we not need this any more? This change seems unrelated to the PR.

hone:
Couldn't repro this failure locally. Rerunning the build. Suspect that even though we hardcoded `engines.node` to 0.10.3 that it installed 4.1.2. O_o /cc @hunterloftis

hone:
Ugh, n/m https://github.com/sharpstone/node_multi/commit/b2e1971aa602f40ffa188c0725fe08e78042d1ac broke the test. I'll patch this.

PR #429, opened 2015-10-29
Fix node buildpack detection

schneems:


schneems:
Ping @hone what do you think of this implementation for fixing the node detect problem?

schneems:
I double checked with the https://github.com/heroku/ruby-getting-started app and it deploys fine without node set, adding the node buildpack and a package.json sets it to the correct node version app name is `peaceful-spire-6675`

hone:
lgtm, do you need to add some other tests around this?

PR #434, opened 2015-11-11
Added a check for preinstalled JDK

jkutner:
This will prevent two JDKs from being installed when the JRuby is used with the jvm-common buildpack. The jvm-common buildpack can be added to install tools.


schneems:
Can you add a test similar to https://github.com/heroku/heroku-buildpack-ruby/blob/8d7d58a77c8df1f027217093c49a54c07c10fe5f/spec/node_spec.rb#L3-L9 ?

jkutner:
Yea, I'll work on that. But I don't think we can really do a good test until Hatchet supports multiple buildpacks set on an app. The existing tests are using heroku-buildpack-multi, which isn't really a good representation.

jkutner:
Depends on this changed being merged and released:
https://github.com/heroku/heroku.rb/pull/94

jkutner:
heroku-api has been released. I updated this PR and it's ready to merge.

I got two test failures locally. But couldn't explain them. I don't think they are related.

schneems:
Thanks

PR #435, opened 2015-11-11
Add retry to `bundle install`

jingweno:
It’d be nice that `bundle install` can retry a bit instead of failing the deployment.

Related: https://github.com/heroku/api/issues/5129.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -542,7 +542,7 @@ WARNING
       log("bundle") do
         bundle_without = env("BUNDLE_WITHOUT") || "development:test"
         bundle_bin     = "bundle"
-        bundle_command = "#{bundle_bin} install --without #{bundle_without} --path vendor/bundle --binstubs #{bundler_binstubs_path}"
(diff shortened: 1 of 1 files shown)


schneems:
The default for the retry in bundler is already set to 3. I wrote this feature :wink: https://github.com/bundler/bundler/pull/2601

It looks like maybe that call 

```
Fetching gem metadata from https://rubygems.org/
```

Isn't wrapped in a retry block and needs to be added. I only added retry functionality to git fetching and `gem install` i've previously never seen anything fail on fetching metadata.

jingweno:
@schneems Cool, didn't know bundler is already retrying :). Hmm...sounds like we need to change bundler itself and that's a bit more involved. Would appreciate any help :wink:.

schneems:
I'm going to Rubyconf, then i'll be OOO for some babysitting duties until after thanksgiving. Happy to help in the mean time, or when I get back. Feel free to ping me on hipchat. We've already got a general purpose class for retrying things, the key would be to find where that output is coming from, and wrap the needed lines in the retry class. Ping me on hipchat if you've got specific questions.

PR #436, opened 2015-11-16
Mention that non-recommended usages are not recommended in warning output

jonahwh:
The best practice output does not mention that these practices are not...best. It only says that they are happening, which the user probably already knows.

This adds another note that these are not recommended practices.


schneems:
I'm not sure "This is best practice" sounds a bit strong. We don't recommend you do these things, I don't exactly know if I feel we can claim ownership of best practices. Also maybe some users legitimately might need those things as they are the best solution for their needs. Even if most of the time that's not the case we don't want to alienate them too much.

crazydog115:
Then instead of "best practice" how about "This is not recommended". It's just unclear why the message is displayed without visiting the provided link.

schneems:
I like that wording better, can you update?

schneems:
Not sure why I never merged. Guess I didn't realized you updated. Thanks!

Issue #437, opened 2015-11-16
JAVA_MEM env variable needs default value


aminjam:
We investigated that when pushing a jruby app there is no default `JVM_MAX_HEAP` value set when switch statement fails to find a value for `ulimit -u`. We have fixed the downstream by adding a default value of `384`. We wanted to open an issue and possibly update the upstream. The [Replace JAVA_TOOL_OPTIONS with JAVA_MEM to avoid JRuby overriding](https://github.com/heroku/heroku-buildpack-ruby/commit/694c3039e710778d6286989ba6a006fd582b88c1) commit seems to have caused the issue. We'd be happy to send in a PR.

The following commit fixes the issue:
https://github.com/cloudfoundry/ruby-buildpack/commit/51b7ed0b7825c7037c038bc9f71767eab3ffc026

```
-  export JAVA_MEM=${JAVA_MEM:--Xmx${JVM_MAX_HEAP}m}
+  export JAVA_MEM=${JAVA_MEM:--Xmx${JVM_MAX_HEAP:-384}m}
```

PR #438, opened 2015-11-16
Java Memory should use the default 384m

aminjam:
Java Memory should use the default 384m when setting the JAVA_MEM env variable. This is in regards to [issue #437](https://github.com/heroku/heroku-buildpack-ruby/issues/437)

Signed-off-by: @bdshroyer


jkutner:
LGTM

PR #439, opened 2015-11-19
Support local builds via BUNDLE_LOCAL env variable

wireframe:
Projects that use `bundle package` should not connect to rubygems for installations.  If BUNDLE_LOCAL flag is set, install gems without network calls.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -544,6 +544,7 @@ WARNING
         bundle_bin     = "bundle"
         bundle_command = "#{bundle_bin} install --without #{bundle_without} --path vendor/bundle --binstubs #{bundler_binstubs_path}"
         bundle_command << " -j4"
+        bundle_command << " --local" if env("BUNDLE_LOCAL")
 
         if bundler.windows_gemfile_lock?
           warn(<<-WARNING, inline: true)


schneems:
I don't think `bundle package` will always work if you're building locally and then deploying on Heroku, especially with c-extensions. You may have different libraries installed or in different locations, and using a different OS. For example a `bundle package` from windows definitely wouldn't be able to be pushed to Heroku. It does look like you can `--all-platforms` but i've honestly never tried it.

What is your main reason for wanting this? Are you looking for faster deploys? All bundle installs are cached, so you should only have to hit them on your first deploy.

wireframe:
My goal here is to support continuous deployment of my app without failures when rubygems (or other gem sources like rails-assets.org) are down.

Gems with C-extensions work fine and the *only* gem i've run into issues with is libv8 which packages separate gems for different architectures.  This is by packaging and checking in both versions locally before pushing to heroku).  With development docker images, it's trivial to run same architectures in development and production so it really hasn't been an issue.  

Here's a related bundler issue that would simplify my "workaround": https://github.com/bundler/bundler-features/issues/4

FWIW I've used this pattern for years w/ capistrano builds and i can't count the number of times this has been necessary.  It has become my default for both CI build servers and deployment production servers.

schneems:
If you have a `vendor/cache` checked into your app then makes your app use `--local` flag https://github.com/bundler/bundler/blob/master/CHANGELOG.md#100rc2-july-29-2010

This should work out of the box right now with bundler 1.11.2.

schneems:
Can you confirm this works?

schneems:
Confirmed that this works in https://github.com/heroku/heroku-buildpack-ruby/issues/481

PR #443, opened 2015-12-21
Instruct curl to follow redirects.

brixen:
The curl command is able to follow redirects but this option was not used previously. Since redirects are a fundamental aspect of how things work on the internet, it would be super helpful to use this.

Thanks! :heart:

### DIFF
--- a/lib/language_pack/fetcher.rb
+++ b/lib/language_pack/fetcher.rb
@@ -31,7 +31,7 @@ module LanguagePack
 
     private
     def curl_command(command)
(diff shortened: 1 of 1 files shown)


schneems:
Running tests on #444. Is Rubinius using redirects?

hone:
@schneems yeah, it recently switched over.

brixen:
@schneems yes, we are migrating to a new AWS account and from rubini.us to rubinius.com. In general, following redirects would be a good idea (eg the standard rubygems link to a .gem is/was a redirect).

brixen:
@schneems just  curious, but is #444 needed to run the specs with Heroku secure env values?

hone:
@brixen yeah basically.

PR #446, opened 2015-12-27
[WIP] Add Lotus language pack

joneslee85:
TODO:

* [ ] Replace assets compilation task with `lotus assets precompile` CLI command
* [ ] Add sample app and test

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Heroku Buildpack for Ruby
 ![ruby](https://cloud.githubusercontent.com/assets/51578/8882726/52caf102-3217-11e5-8a6c-2d47c57f0030.jpg)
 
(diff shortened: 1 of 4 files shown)


schneems:
We talked about this and think this would add complexity to the Ruby ecosystem. We have a standard hook for running asset compilation `rake assets:precompile` all Ruby apps will run that. We also expose DATABASE_URL but don't actually do anything special with it in Rails. All DATABASE_URL functionality is provided by Rails. If anything we are working to reduce logic and introduce less forking to the buildpack. If you want you could build a micro lotus buildpack that you can run with `heroku buildpack:add` after Ruby runs. So the ruby buildpack would install ruby, then the micro buildpack could do whatever custom stuff you want if you're not able to get support upstream into lotus.

joneslee85:
@schneems yes, I've talked with Luca about this already. It's been decided that Lotus would change default DB URI to `DATABASE_URL` as of https://github.com/lotus/lotus/pull/435

schneems:
Thats great news :rocket: :heart:

PR #447, opened 2015-12-27
Change parent class of Rails23 langpack to Rack langpack

joneslee85:
I think Rails 2.3.x should inherit from Rack base instead of Ruby base.

### DIFF
--- a/lib/language_pack/rails2.rb
+++ b/lib/language_pack/rails2.rb
@@ -3,7 +3,7 @@ require "language_pack"
 require "language_pack/rack"
 
 # Rails 2 Language Pack. This is for any Rails 2.x apps.
-class LanguagePack::Rails2 < LanguagePack::Ruby
+class LanguagePack::Rails2 < LanguagePack::Rack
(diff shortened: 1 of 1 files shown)


schneems:
We can't make arbitrary changes to the buildpack. Even the smallest most trivial changes have a very long tail of breaking changes our test coverage is far from comprehensive. Is there something this change buys us?

joneslee85:
Nothing really for the language pack pretty much duplicated whatever there in rack language pack. It's just weird that rails23 require rack pack and yet did not use it

joneslee85:
@schneems I don't see any errors in the TravisCI log, it seems to me the build just hang. This seems to be also the case with other PRs, is this a known issue?

schneems:
The env vars are secure, you can't trigger a passing build, you have to have commit access to the repo to run the tests with travis.

joneslee85:
I see. I guess nothing much I could do. Please advice me if you want me to close this PR or not

schneems:
Thanks for the idea and the PR

PR #448, opened 2015-12-27
Remove securerandom loading from Rails 5

joneslee85:
Because Rails41 class has already included it


schneems:
I'll have to do this on a branch for tests. Pretty sure you're right but people deploy production apps of of our master branch, have to be very careful.

PR #449, opened 2016-01-06
don't specify the heroku-postgresql plan, let addon pick

hone:


will:
_technically_ core picks, the addon has no say.

PR #450, opened 2016-01-12
Added pgconfig jar to JDK for JRuby JDBC

jkutner:
This change is necessary to support new dedicated single-tenant instances of postgres that require an SSL connection. By default the Postgres JDBC driver defaults to using an unencrypted connection. 

This change has already been rolled out for Java, Scala, Clojure and Groovy apps. For more information [read this  blog post](http://jkutner.github.io/2016/01/12/postgres-ssl.html).

/cc @neovintage

### DIFF
--- a/lib/language_pack/helpers/jvm_installer.rb
+++ b/lib/language_pack/helpers/jvm_installer.rb
@@ -11,6 +11,8 @@ class LanguagePack::JvmInstaller
   JVM_1_7_25_PATH = "openjdk1.7.0_25.tar.gz"
   JVM_1_6_PATH    = "openjdk1.6-latest.tar.gz"
 
+  PG_CONFIG_JAR   = "https://lang-jvm.s3.amazonaws.com/pgconfig.jar"
+
   def initialize(slug_vendor_jvm, stack)
     @vendor_dir = slug_vendor_jvm
     @stack = stack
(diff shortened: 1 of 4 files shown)


lib/language_pack/helpers/jvm_installer.rb
```diff
@@ -88,4 +92,8 @@ def fetch_env_untar(key)
     fetch_untar(jvm_version)
     true
   end
+
+  def install_pgconfig_jar(url)
+    run("curl --retry 3 -s -L #{url} -o #{@vendor_dir}/jre/lib/ext/pgconfig.jar")
```
schneems: Could we use a fetcher here instead? That way if we need to update our curl commands we can do it all in one space.

schneems:
I think this looks good overall. Would prefer to use a fetcher where noted. Is it possible to test an SSL connection is made and possible via something like a `heroku run` something command? While the tests do assert the file is there, it wouldn't fail if the permissions changed or wasn't usable some other way. On the flip side if it would make the suite brittle we can avoid it.

jkutner:
Added the fetcher and used a common base bucket URL for jdk and the pgconfig.jar.

@schneems integration testing the pg connection is unfortunately brittle. The only way is to create a database that requires SSL and ensure that it connects. anything else would just be testing that the property itself is set, which isn't much better than what we have now.

Currently, there are no heroku database plans that require SSL. In the future the standard-0/1 will, which is why this feature is required, but I'm not sure if we want to provision those in the tests.

The `pgconfig.jar` itself is tested outside of this buildpack, and I think that's the place to put the brittle tests. That's a future task for me.

hone:
That sounds reasonable. These changes look good to me.

PR #454, opened 2016-01-28
change default ruby to 2.2.4 from 2.0.0

hone:
Ruby 2.0.0 is being sunset on Feb. 24, 2016. This changes the default
for new apps that haven't set a ruby in their Gemfile. By doing so, we
won't be putting users on a Ruby being sunset soon.

In addition to Ruby being sunset, Rails 5 beta needs a newer Ruby as
well.

This replaces #441 with 2.2.4 instead of 2.2.3.


schneems:
:shipit:

PR #459, opened 2016-02-17
Warn on no Procfile use, don't mention Webrick

schneems:
While someone __may__ be using webrick they may also be puma or some other webserver with an integrated rack handler. Make the error more generic. The webpage was updated to say that a `Procfile` is best practice. It still talks about WEBrick but this is secondary.


schneems:
cc @hone

PR #462, opened 2016-02-29
[close #358] Fail on Rake detect for Rails Apps

### LINKED ISSUE #358: assets:precompile fails, app build continues
A gem was unable to be required during assets:precompile and caused a state where there were no assets available, but the app build continued: https://gist.github.com/calebthompson/7daa3551b44f8a2fe510

schneems:
Rails applications should be reasonably expected to run `rake` when a Rakefile is present.

Right now this is the behavior we support https://github.com/heroku/heroku-buildpack-ruby/issues/358#issuecomment-90250883. While it might be reasonable that a Ruby application might not expect us to run Rake, a Rails app with the asset pipeline, while they might not be using the asset pipeline, should be able to run the `rake` command when a Rakefile is present.

### DIFF
--- a/hatchet.json
+++ b/hatchet.json
@@ -49,7 +49,9 @@
     "sharpstone/rails3_12factor",
     "sharpstone/rails3_one_plugin",
     "sharpstone/rails3_runtime_assets",
-    "sharpstone/rails3-fail-assets-compile"
+    "sharpstone/rails3-fail-assets-compile",
+    "sharpstone/rails3-fail-rakefile"
+
   ],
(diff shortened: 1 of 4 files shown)


schneems:
@hone @dmathieu @calebthompson thoughts?

calebthompson:
Nice.

schneems:
Should probably add a test to ensure this actually fails on Rails 3+ apps. Also need to verify that this won't result in __too__ many errored builds.

- [ ] Add Test
- [ ] Verify minimal impact

PR #465, opened 2016-03-01
Alternative to #463 always warn on Bundler change

schneems:
Warn when upgrading bundler version. Instead of trying to detect errors or problems make the fact that we upgraded bundler much more noticeable. This may result in more false flag support ticketes but it also points people to how to test their app to ensure the problem was not in-fact caused by the bundler version change by setting the release version on their app.

This warning will only be present for one successful build.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -829,9 +829,23 @@ params = CGI.parse(uri.query || "")
       rubygems_version_cache  = "rubygems_version"
       stack_cache             = "stack"
 
+      old_bundler_version  = @metadata.read(buildpack_version_cache)
       old_rubygems_version = @metadata.read(ruby_version_cache).chomp if @metadata.exists?(ruby_version_cache)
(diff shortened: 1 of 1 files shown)


lib/language_pack/ruby.rb
```diff
@@ -829,10 +829,24 @@ def load_bundler_cache
       rubygems_version_cache  = "rubygems_version"
       stack_cache             = "stack"
 
+      old_bundler_version  = @metadata.read(buildpack_version_cache)
```
hone: you probably need to chomp here to get rid of the newline

schneems: I merged this maybe 2 seconds before you left that comment. I'll update with a chomp.

schneems: And...I used the wrong metadata directory, fixing.

schneems: https://github.com/heroku/heroku-buildpack-ruby/pull/468

schneems: Merged and fixed

PR #471, opened 2016-03-10
Warn on .bundle/config file presence

schneems:
Discussion at https://github.com/bundler/bundler/issues/4351#issuecomment-194490139

Also do not load the stored `.bundle` directory. We are passing in all the flags that we need to `bundle install`. If we also store a `.bundle/config` file between builds than if you accidentally include `.bundle/config` file in your repo with a `BUNDLE_WITHOUT` value ,it will get persisted to the `.bundle/config` file to the cache and deleting the file from your repo does not get rid of the problem. The config is un-needed, since we're using all flags all the time, let's not rely on the `.bundle/config` file.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -544,6 +544,16 @@ WARNING
         bundle_command = "#{bundle_bin} install --without #{bundle_without} --path vendor/bundle --binstubs #{bundler_binstubs_path}"
         bundle_command << " -j4"
 
+        if File.exist?("#{Dir.pwd}/.bundle/config")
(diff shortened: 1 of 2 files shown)


lib/language_pack/ruby.rb
```diff
@@ -558,7 +568,6 @@ def build_bundler
         else
           # using --deployment is preferred if we can
           bundle_command += " --deployment"
-          cache.load ".bundle"
```
schneems: We talked about this in hipchat. You can't detect if a user has created this file or if it was previously generated by Heroku. We've had this change merged into master for some time now and it hasn't caused any issues. Going to deploy.

PR #479, opened 2016-04-08
Allow custom asset prefix for rails 4 and up

dennisvdvliet:
# Heroku ruby buildpack and assets in custom path

Although most developera will let heroku handle compiling their assets there are some use cases when precompiling your assets locally make sense.

By default the heroku buildpack for ruby checks in the  **public/assets** directory for a manifest. If a manifest is found it is assumed that assets are already compiled and this step of the [compile step](https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/rails4.rb#L72) is skipped.

With the default rails settings this works out of the box. However if you use the *config.assets.prefix* setting as described in the [Rails documentation](http://guides.rubyonrails.org/asset_pipeline.html#precompiling-assets) this stops working.

The build pack only check in the **public/assets** directory so if the manifest is saved in an other directory it will not be detected and thus assets will always be compiled during deploy.

I created a little example app to showcase current and (in my eyes) desired behaviour.

## Current behaviour - default settings

```
git clone git@github.com:dennisvdvliet/heroku-buildpack-rails-demo.git
git checkout current-behaviour-default-settings
heroku create
git push heroku current-behaviour-default-settings:master
```

This will trigger a build on heroku and after installing your gems you will see this in the log.

```
remote:        Cleaning up the bundler cache.
remote:        Detected manifest file, assuming assets were compiled locally
```

## Current behaviour - custom prefix

Now we alter application.rb and add

```ruby
config.assets.prefix = '/custom/path'
```

Assets will now be precompiled in **public/custom/path**
Now push to heroku again.

```
git checkout current-behaviour-custom-prefix
git push heroku current-behaviour-custom-prefix:master --force
```

This will result in the following log message, that clearly tells us that the precompiled assets were ignored.

```
remote: -----> Preparing app for Rails asset pipeline
remote:        Running: rake assets:precompile
remote:        Asset precompilation completed (1.52s)
remote:        Cleaning assets
```

## Desired behaviour

For this I made a small change to the [buildpack](https://github.com/dennisvdvliet/heroku-buildpack-ruby/commit/87f3aa169818c88d6dd09a1663c1f1cc6a779a2e) to scan the **public** and subdirectories directory for the manifest.

```
heroku create --buildpack https://github.com/dennisvdvliet/heroku-buildpack-ruby.git#rails-asset-prefix rails-asset-prefix
heroku git:remote --app rails-asset-prefix --remote rails-asset-prefix
git push rails-asset-prefix current-behaviour-custom-prefix:master --force
```

Checking the output from the deploy log confirms that this does what is expected.

```
remote:        Bundle completed (26.14s)
remote:        Cleaning up the bundler cache.
remote:        Detected manifest file, assuming assets were compiled locally
```

I'm no buildpack expert so happy to hear what the rest of you think about this.

### DIFF
--- a/lib/language_pack/rails4.rb
+++ b/lib/language_pack/rails4.rb
@@ -69,7 +69,7 @@ WARNING
   def run_assets_precompile_rake_task
     instrument "rails4.run_assets_precompile_rake_task" do
       log("assets_precompile") do
-        if Dir.glob("public/assets/{.sprockets-manifest-*.json,manifest-*.json}", File::FNM_DOTMATCH).any?
+        if Dir.glob(["public/**/*/{.sprockets-manifest-*.json,manifest-*.json}"], File::FNM_DOTMATCH).any?
           puts "Detected manifest file, assuming assets were compiled locally"
           return true
         end


schneems:
I don't want to support compiling to a custom directory.

If you don't want us to compile your assets for you , add a manifest file to your public/assets directory (it can be blank). This would fail for someone who had a manifest file that isn't part of sprockets, and they did want you to run `assets:precompile`.

dennisvdvliet:
Good point, what about using an environment variable allowing users to override the search pattern used in the Dir.glob is that something you would consider?

schneems:
I don't think so, not for now. Maybe in the future. I'll have to think more about it.

elia:
Here's my usecase for this.

We use different asset prefixes for different deploy stages

- _development_ goes to `/assets_dev` and of course compiles through sprockets on the fly
- _production_ uses `/assets` and keeps the manifest in the git repo

I'd like to have _staging_ and ad-hoc apps for feature branches to use `/assets_staging` and have assets precompiled at buildtime. Now the buildpack just detects the production manifest and refuses to do it. This would be solved by either adding support for custom prefixes var or by having a way to force/prevent precompilation. Both solutions can be implemented using env vars.

e.g.

```rb
  def run_assets_precompile_rake_task
    instrument "rails4.run_assets_precompile_rake_task" do
      log("assets_precompile") do
        if Dir.glob("public/#{ENV['RAILS_ASSETS_PREFIX'] || 'assets'}/{.sprockets-manifest-*.json,manifest-*.json}", File::FNM_DOTMATCH).any?
          puts "Detected manifest file, assuming assets were compiled locally"
          return true
        end
```

Issue #452, opened 2016-01-24
Bundle support for version operators in the ruby version


blackjid:
The bundler team has merged a PR that let you use version operators for semver in the `ruby` method in the `Gemfile`. This is going to be released in the next version of bundler. https://github.com/bundler/bundler/pull/4064

The thing is that you will be able to add something like this `ruby "~>2.2"` in the Gemfile. How this buildpack is going to handle this case? What ruby version will choose?

thanks

schneems:
We will need to see, depending on how it works we still might make you explicitly state a version. 
On one hand getting the latest security patches is good, on the other hand having your site accidentally break because there was a regression and you didn't test locally or even know there was a minor version released is a bad experience. "It worked yesterday and we didn't change anything" is a really bad experience.


---
Richard Schneeman
http://www.schneems.com




On Sat, Jan 23, 2016 at 8:07 PM -0800, "Juan Ignacio Donoso" <notifications@github.com> wrote:












The bundler team has merged a PR that let you use version operators for semver in the ruby method in the Gemfile. This is going to be released in the next version of bundler. bundler/bundler#4064



The thing is that you will be able to add something like this ruby "~>2.2" in the Gemfile. How this buildpack is going to handle this case? What ruby version will choose?



thanks

jtarchie:
@schneems, I'm from the Cloud Foundry team that did the original PR for the feature in bundler.

Our use case, as described above, was to allow users to get the latest security patches of Ruby. This feature is great as the Cloud Foundry fork of the buildpack supports only a subset of Ruby versions -- usually the latest two patches of a minor.

It would be awesome if this could be pulled into the code. Our team is willing to help with upstream changes to help you better support this feature.

ericboehs:
@schneems Bundler does add an exact version in the `Gemfile.lock`. That'd probably be the correct place to check.

```shell
grep -A1 'RUBY VERSION' Gemfile.lock
```

schneems:
I wrote that comment in a taxi on my iphone. Thanks for the sanity check. I also ran locally and confirm it will work with the buildpack.

jtarchie:
The `Gemfile.lock` will only contain Ruby version used in development at that time. This means the `ruby` directive in `Gemfile` could specify it is alright with a newer version of Ruby that the buildpack does support.

For example:

`Gemfile` contains:

```
ruby "> 2.2.0"
```

`Gemfile.lock` contains:

```
RUBY VERSION
[...]

PR #483, opened 2016-05-04
Bump bundler version to 1.12.1

vincentwoo:
Bundler 1.12 offers a bunch of important enhancements and fixes:
http://bundler.io/blog/2016/04/28/the-new-index-format-fastly-and-bundler-1-12.html

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -14,7 +14,7 @@ class LanguagePack::Ruby < LanguagePack::Base
   NAME                 = "ruby"
   LIBYAML_VERSION      = "0.1.6"
   LIBYAML_PATH         = "libyaml-#{LIBYAML_VERSION}"
-  BUNDLER_VERSION      = "1.11.2"
+  BUNDLER_VERSION      = "1.12.1"
   BUNDLER_GEM_PATH     = "bundler-#{BUNDLER_VERSION}"
   RBX_BASE_URL         = "http://binaries.rubini.us/heroku"
   NODE_BP_PATH         = "vendor/node/bin"


vincentwoo:
tests failed with something weird that I am pretty sure have nothing to do with this change:

```
== cloning 'git://github.com/sharpstone/node_multi.git' into '/home/travis/build/heroku/heroku-buildpack-ruby/repos/multibuildpack/node_multi'
if [ `git config --get user.email` ]; then echo 'already set'; else `git config --global user.email 'buildpack@example.com'`; fi
if [ `git config --get user.name` ];  then echo 'already set'; else `git config --global user.name  'BuildpackTester'`      ; fi
echo '
Host heroku.com
    StrictHostKeyChecking no
    CheckHostIP no
    UserKnownHostsFile=/dev/null
    IdentityFile ~/.ssh/id_rsa
Host github.com
    StrictHostKeyChecking no
' >> ~/.ssh/config
ssh-keygen -f ~/.ssh/id_rsa -t rsa -N ''
Generating public/private rsa key pair.
Your identification has been saved in /home/travis/.ssh/id_rsa.
Your public key has been saved in /home/travis/.ssh/id_rsa.pub.
The key fingerprint is:
9c:43:29:54:24:11:5e:e1:f8:ae:d7:93:6f:b1:21:80 travis@testing-worker-linux-docker-01d832f9-3360-linux-10
The key's randomart image is:
+--[ RSA 2048]----+
|      ===.       |
|     o = .       |
|      +.+        |
|      E=..       |
|        S.       |
|       . .. o    |
|        .. o +   |
|       .. + o    |
|      ..   +.    |
+-----------------+
yes | heroku keys:add
heroku-cli: Installing Toolbelt v4... 2.65MB/2.65MB
heroku-cli: Updating to 5.1.4-ccc1af5... done
Enter your Heroku credentials.
Email: Password (typing will be hidden): stty: standard input: Inappropriate ioctl for device
stty: standard input: Inappropriate ioctl for device
 ▸    Authentication failed.
 ▸    Email or password is not valid.
 ▸    Check your credentials on https://dashboard.heroku.com
rake aborted!
Command failed with status (2): [yes | heroku keys:add...]
```

schneems:
Hey @vincentwoo , to deploy bundler we have to first install it on a heroku dyno, capture the expanded output directory tar it up and then put it in S3. Then we set permissions, test it and then finally the very last step is to change that version constant value. Changing the value without uploading a version of bundler to S3 does nothing, unfortunately.

Why do you want 1.12.1? We typically are very conservative with our bundler upgrades as they tend to be one of the less stable pieces of our infrastructure, this means we upgrade it less frequently and only after significant need is demonstrated and testing is performed. Explaining why you want/need it could help.

vincentwoo:
1.12.1 (they're actually up to 1.12.3 now) is the first release that fixed an outstanding bundler bug that would cause frivolous changes in Gemfile.lock. I can use this version on the client, but found Heroku's warnings about my version being ahead of the server version to be mildly annoying.

I totally get not wanting to upgrade, but the other thing I'd say is that 1.12 is also probably one of the biggest changes in a long time so you should probably hop on it sooner or later.

schneems:
>  but the other thing I'd say is that 1.12 is also probably one of the biggest changes in a long time so you should probably hop on it sooner or later.

Can you say more?

vincentwoo:
The big architectural change was the new format they're using for the index, but I was just going by the size of the changelog: https://github.com/bundler/bundler/blob/master/CHANGELOG.md

ericboehs:
1.12 also adds version operator support for specifying ruby versions in the Gemfile.

I don't know if upgrading to 1.12 means that we support 1.12's features but builds definitely fail when they hit `ruby '~> 2.3.0'`.

See also https://github.com/heroku/heroku-buildpack-ruby/issues/452.

schneems:
Looks like our tooling would work with the version operator, we just need a new version of bundler. I'll start warming up our bundler version upgrade engines.

jtarchie:
This is just a comment to cc the following users: @jvshahid @flavorjones @dannyzen

PR #487, opened 2016-05-26
Removed hard-coded error about Ruby 2.1

jmccartie:
```
irb(main):012:0> version = "ruby-2.3.0-p0"
=> "ruby-2.3.0-p0"
irb(main):013:0> version.match(/(\d\.\d)/)[0]
=> "2.3"
```

New string will read: 

> Note: Only the most recent version of Ruby 2.3 is supported on Cedar-14

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -403,9 +403,9 @@ ERROR
       message = <<ERROR
 An error occurred while installing Ruby #{ruby_version.version}
 For supported Ruby versions see https://devcenter.heroku.com/articles/ruby-support#supported-runtimes
-Note: Only the most recent version of Ruby #{ruby_version.version.match(/(\d\.\d)/)[0]} is supported on Cedar-14
 #{error.message}
 ERROR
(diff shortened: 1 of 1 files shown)


schneems:
The reason we have the 2.1 message in there is it's the only version that had cedar-10 binaries that wouldn't compile on cedar 14. It's some older random versions like 2.1.2 or something, I don't even know which ones are missing. We should maybe instead detect and see if they're trying to install a version that starts with 2.1 before adding that error message.

jmccartie:
@schneems So that line of the error should only print out if It's `~> 2.1`?

schneems:
Exactly

jmccartie:
@schneems Like-ah dees?

schneems:
Sorry, forgot to mention earlier. While I was mucking with the new ruby version specifier stuff I needed to work on the errors, so I went ahead and added that https://github.com/heroku/heroku-buildpack-ruby/commit/0ec19283b6e601037439e74745b19b5168bddcb0#diff-3fc66e13e389ff4d15c7ca3ddb86464eR408

Thanks for the idea, sorry for not giving you credit via commit 😁 

I still value you 🎉❤️🙌

jmccartie:
@schneems I didn't feel valued. But then I saw the emojis. So we're cool. 👬

PR #488, opened 2016-05-27
Revert "Revert "Bump bundler 1.12.5""

schneems:
This reverts commit 088683acf9f6edde85179061480ec1850956cc2f.

This PR bumps up the version of bundler to 1.12.5 which allows us to use Ruby version specifiers.

Due to change in behavior in bundler https://github.com/bundler/bundler/issues/4621 we must now detect the when patchlevel is not needed and strip it everywhere we were previously using a version.

Right now there is no way to know when a bundle platform failed with an invalid output, it doesn't do Ruby version validations so if you put "2.3.x" in your Gemfile, then it will return `ruby 2.3.x` back out at you which isn't a valid version and we can't download it. 

I'm going to update the "specify a ruby version" docs to include the ruby version specifiers. I've made this error when a build fails a bit cleaner. We're now only warning about Ruby 2.1 when they're trying to use Ruby 2.1. Also i'm identifying the curl output as "debug information" so hopefully fewer people will think that there is some really strange buildpack error and instead focus on our well written message with links. 

I would like to call out that for failed downloads people should try to re-deploy before declaring it a lost cause, but that's a bit too verbose for writing in an encyclopedic way inside the error output. Think most people will retry anyway. I'm also bumping up the retries to 5, because why not? It might mean that people deploying bad versions of Ruby get delayed by an extra 2 seconds, but hopefully it will cut down on the number of tickets that are solved by people re-running the deploy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master
 
+* Bump bundler to 1.12.5 [Bundler changfelog](https://github.com/bundler/bundler/blob/master/CHANGELOG.md#1123-2016-05-06). Allows for use of Ruby version operators.
+
 ## v146 (03/23/2016)
 
 * Warn when `.bundle/config` is checked in (#471)
--- a/hatchet.json
+++ b/hatchet.json
@@ -14,10 +14,12 @@
   ],
   "bundler": [
     "sharpstone/bad_gemfile_on_platform",
+    "sharpstone/problem_gemfile_version",
     "sharpstone/git_gemspec",
     "sharpstone/no_lockfile",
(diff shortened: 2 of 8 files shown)


schneems:
cc @hone @dmathieu this is a risky merge. 

Going to wait till monday to get it on master, then i'll try not to deploy until Thursday.

dmathieu:
How about the failing test?

schneems:
All tests pass. The "travis-ci/pr" tests never passes due to the way we test (we deploy based off the branch that you're on, so when travis tries to merge and deploy to master then buildpacks:set fail since this code isn't on master).

dmathieu:
LGTM then :)

PR #491, opened 2016-07-12
set HEROKU_DYNO_TYPE to 1X/2X/PM/PL

nruth:
Does this look like something that can be added?

It helps with custom puma configs that work with web process auto schedulers/scalers that change dyno type based on time of day or measured load. Otherwise we have to base it on the set heroku memory limit or looking at `ulimit -u` again, which are already handled in this buildpack and it seems brittle / duplication to do it again in the app's puma config.

If the idea is reasonable it needs refactoring. It should be done outside of set_default_web_concurrency, and seems like it could be used by some of the other methods instead of bash case statements that look at `ulimit -u` again.


schneems:
I don't love this approach. The dyno naming is something we already have changed once it's no longer `1X` dyno it's `Standard-1X`. I don't want to break people's apps who rely on this name and I don't want to emit wrong/confusing names.

nruth:
Is there another approach?

schneems:
You can look directly at `HEROKU_RAM_LIMIT_MB` to make those judgements. That would be your best bet.

nruth:
Is there another place that I can suggest this as a platform feature, as it seems more like a heroku platform request than a ruby/buildpack specific feature?

Having things like https://elements.heroku.com/addons/process-scheduler in the market-place cries out for this kind of thing, and building production server setups relying on settings that could well be changed or removed without official support or awareness of weird custom setups just seems like asking for trouble.

PR #496, opened 2016-09-22
Bundler v1.13.1

schneems:


jkutner:
👍

gmoore:
Hi @schneems this appears to run in contrast to Heroku's documented Bundler version (1.11) here: https://devcenter.heroku.com/articles/ruby-support#libraries 

Are the docs going to be updated?

schneems:
This version isn't deployed. We only document the currently deployed version. When (if) we deploy this version, we'll update the docs.

gmoore:
@schneems Thanks. I didn't realize that putting the buildpack URL into .buildpacks locked us to master and not a release version. We're going to try taking it out.

schneems:
Yes, we see that from time to time from apps that were around in the `.buildpacks` days. We consider master "stable" and do our best to keep it bug free, but it does see changes before they are deployed. For larger applications I recommend using the officially deployed version.

PR #498, opened 2016-09-28
Announce when we detect Rake tasks

schneems:
Right now people are putting infinite scripts in their rakefile and it looks like it's getting stuck in `bundle clean` when it should look like it's getting stuck while detecting rake tasks.


hone:
:+1:

PR #508, opened 2016-11-25
Initialize env before detecting the app type

dmathieu:
Detecting the app type requires bundler, which makes a curl call.
Initializing the env after means we cannot override CURL_CONNECT_TIMEOUT or CURL_TIMEOUT.


dmathieu:
It would appear this repo doesn't like PRs from forks very much. I'll reopen from a local branch.

PR #509, opened 2016-11-25
Initialize env before detecting the app type

dmathieu:
Detecting the app type requires bundler, which makes a curl call.
Initializing the env after means we cannot override CURL_CONNECT_TIMEOUT or CURL_TIMEOUT.


schneems:
You are peer reviewed. Can't deploy yet though. Thanks!

PR #510, opened 2016-11-28
Use ShellHelper#env in language pack fetcher

dmathieu:
This complements #509.

Custom env vars aren't set into the global `ENV` variable.
Without this change, we can't set custom CURL timeouts in curl requests.


schneems:
Funny, that the people who explicitly asked for this feature, never actually noticed that it didn't work.

dmathieu:
It would work if they didn't filter the environment variables in their builds (which would be a security issue).

PR #513, opened 2016-12-08
Fix initialize_env fails if env path contains a directory

mmacvicar:
The method initialize_env assumes that the env path argument contains only files, this is causing an exception in a deis/kubernetes environment. This is very dependent on the environment where the buildpack is used and it probably only affects Deis setups. Still it seems reasonable to check if each path is a file before trying to read the contents.

### DIFF
--- a/lib/language_pack/shell_helpers.rb
+++ b/lib/language_pack/shell_helpers.rb
@@ -33,9 +33,10 @@ module LanguagePack
     def self.initialize_env(path)
       env_dir = Pathname.new("#{path}")
       if env_dir.exist? && env_dir.directory?
-        env_dir.each_child do |file|
-          key   = file.basename.to_s
-          value = file.read.strip
+        env_dir.each_child do |child|
+          next unless child.file?
+          key   = child.basename.to_s
+          value = child.read.strip
           user_env_hash[key] = value unless blacklist?(key)
         end
       end


schneems:
What do you have that isn't a file in that directory?

mmacvicar:
K8s creates symlinks to files which are the actual variables and also leaves there a symlink named ..data pointing to a folder, looks like something internal related to how configMaps work. There is already a PR for slugbuilder that literally moves the env folder to a different place ignoring the ephemeral ..data. This just to make sure this buildpack does not fail. Fixing it on the buildpack makes sense as it is easier and makes initialize_env more robust, that fix on slugbuilder just tries to go around the issue. On another hand, fixing it on slugbuilder would avoid the need to upgrade the buildpacks.

The slack fork seems to also have implemented the same fix, but they did not send the PR.
https://github.com/heroku/heroku-buildpack-ruby/compare/master...slack:env-from-files

Here is the related issue in Deis builder:
https://github.com/deis/builder/issues/454

It is also opened in slug builder
https://github.com/deis/slugbuilder/issues/124

schneems:
I'm not super +1 on fixing this here if there is somewhere else that fixes the cause of the non-files in this folder. While it might be a trivial amount of time per each file check, that extra file stat isn't free (in terms of performance) when you multiply it by all N files for each config var by N Ruby apps running on Heroku it's perhaps not a huge amount of time, but it's not nothing either.

bacongobbler:
We're fixing this upstream. Heroku is right, Kubernetes' setup of volumes is wonky. We're working around that so this isn't necessary. :)

schneems:
> Heroku is right

Happy to speak for heroku 😎

Going to close for now, thanks for bringing it up for discussion! Ping again if we need to re-open discussion.

PR #515, opened 2016-12-16
bootstrap buildpack with vendored ruby

hone:
This makes the buildpack not reliant on the ruby on the stack image, so we can control what version we're using and what version of rubygems is used during `bundle install`.


hone:
@schneems would love a peer review/check over when you get a chance

lib/language_pack/installers/ruby_installer.rb
```diff
@@ -0,0 +1,30 @@
...
+module LanguagePack::Installers::RubyInstaller
+  include LanguagePack::ShellHelpers
+
+  DEFAULT_BIN_DIR = "bin"
+
+  def self.installer(ruby_version)
+    if ruby_version.rbx?
+      LanguagePack::Installers::RbxInstaller
+    else
+      LanguagePack::Installers::MriInstaller
```
schneems: For naming the MRI handles JRuby as well. Maybe it makes more sense to call it something more generic? `HerokuRubyInstaller` (or other), the differentiation is between what we support via our bucket and what we don't versus the interpreter?

bin/compile
```diff
@@ -1,31 +1,17 @@
...
 
-# sync output
-$stdout.sync = true
+BIN_DIR=$(cd $(dirname $0); pwd)
+ROOT_DIR=$(dirname $BIN_DIR)
 
-$:.unshift File.expand_path("../../lib", __FILE__)
-require "language_pack"
-require "language_pack/shell_helpers"
+ruby_dir="$ROOT_DIR/vendor/ruby/$STACK"
```
schneems: Pragmatically i've never actually seen a ticket where this has bit us in the butt so i'm fine moving forward with this. Ideologically I'm not in love with it. At the end of the day i'm fine with this location.

schneems:
This is needed for tests due to many gems having this check in the gemspec:

```
  if spec.respond_to?(:metadata)
    spec.metadata['allowed_push_host'] = "TODO: Set to 'http://mygemserver.com'"
  else
    raise "RubyGems 2.0 or newer is required to protect against " \
      "public gem pushes."
  end
```

Due to this we need a more recent version of Rubygems than what is provided to us. This PR gives us the ability to rev our version independently.

PR #526, opened 2017-01-18
RAILS_GROUPS should only be used for rake execution and not set in .profile.d

### LINKED ISSUE #525: v151: Adding ENV['RAILS_GROUPS'] = 'assets' Breaks Implementation
Issue #516 made [a change](https://github.com/heroku/heroku-buildpack-ruby/pull/516/commits/bcc74ad47a502cdad4bcfc434fd0ef435e01164d) to the Rails 3 Language Pack that merges in the environment variable 'RAILS_GROUPS' set to 'assets' by default. Prior versions of the buildpack did not add this environment variable.

We had a section of our code that relied on the response to `Rails.groups.include?('assets')` which resulted in our app breaking after this new buildpack was implemented since all old builds returned `false` and all new builds on Heroku return `true`.

What was the reasoning behind adding this environment variable by default? Is it possible to fix this latest version to remove that default env var?

hone:
This fixes #525.

#516 accidentally wrote `RAILS_GROUPS` in rails3+ to `.profile.d`, which was not intended behavior.

### DIFF
--- a/lib/language_pack/rails3.rb
+++ b/lib/language_pack/rails3.rb
@@ -19,12 +19,6 @@ class LanguagePack::Rails3 < LanguagePack::Rails2
     "Ruby/Rails"
   end
 
-  def default_env_vars
-    super.merge({
-      "RAILS_GROUPS" => "assets"
-    })
-  end
-
   def default_process_types
     instrument "rails3.default_process_types" do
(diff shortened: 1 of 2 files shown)


schneems:
Thanks!

PR #530, opened 2017-01-24
New year rollovers are hard. =)

jfilip:
I saw that the dates were incorrect for January releases while reading the Changelog.


hone:
:+1: Thanks!

PR #531, opened 2017-01-29
Bundler 1.14.3

kimyu92:
Bundler to 1.14.3 [changelog](https://github.com/bundler/bundler/blob/v1.14.3/CHANGELOG.md)


schneems:
We have to explicitly build it, it's a more involved process than just setting a config var. Is there a feature in 1.14.3 that you need?

PR #535, opened 2017-02-17
Limit the creation of postgres databases to apps using one of the pg adapters

dmathieu:
And extend that to rails apps, so we don't provision a database needlessly for them either.
Closes heroku/heroku-buildpack-ruby/issues/532

### DIFF
--- a/lib/language_pack/rails2.rb
+++ b/lib/language_pack/rails2.rb
@@ -79,12 +79,6 @@ private
     end
   end
 
-  # most rails apps need a database
-  # @return [Array] shared database addon
-  def add_dev_database_addon
-    ['heroku-postgresql']
-  end
-
   # sets up the profile.d script for this buildpack
   def setup_profiled
     super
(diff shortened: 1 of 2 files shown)


dmathieu:
@hone @schneems as far as I can see, it appears the tests are failing because hatchet doesn't wait until the app is fully deployed and restarted to make requests to it, making it call a blank app.

schneems:
Can you re-submit this PR from a branch of the buildpack? The tests only fire if they're coming from within the same repo (they use encrypted env vars). 

Where did you get that list from? Any chance we're missing some?

hone:
@schneems looks like it's already a [branch from the repro](https://github.com/heroku/heroku-buildpack-ruby/tree/no_pg_database_unless_required).

dmathieu:
@hone I'm discussing this with Richard directly on slack :)
@schneems I just rebased this against master.

hone:
Oops. Sorry, carry on. :)

PR #538, opened 2017-02-22
Update heroku_hatchet in Gemfile.lock

jkutner:
Hatchet 1.4.3 includes better support for Git repos over HTTPS.


jkutner:
Looks like this doesn't work on Travis.

schneems:
🤔

jkutner:
I think they way I'm running locally is not the same as the Travis config. I'll see if I can align them.

PR #540, opened 2017-02-25
replace s3pository with nodebin

hone:
upgrade node to latest Node LTS for Rails 5.1
s3pository is being replaced with nodebin
can't push to cedar-10 stack anymore, so don't need "legacy"
node 0.12 is ancient at this point, we should pin to LTS now

/cc @hunterloftis


schneems:
Awesome start, tests are failing. Looks like due to rate limiting. Restarting tests.

hone:
added `yarn` support if we detect the `webpacker` gem.

lib/language_pack/helpers/nodebin.rb
```diff
@@ -0,0 +1,24 @@
+require 'json'
+
+class LanguagePack::Nodebin
```
schneems: The other helpers are name-spaced `LanguagePack::Helpers`

lib/language_pack/helpers/nodebin.rb
```diff
@@ -0,0 +1,24 @@
...
+      JSON.parse(response.body)
+    end
+  end
+
+  def self.node(q)
+    query("node/linux-x64/#{q}")
+  end
+
+  def self.node_lts
+    node("latest?range=6.x")
```
schneems: Will we need to manually rev this for future versions of node?

lib/language_pack/helpers/node_installer.rb
```diff
@@ -1,44 +1,31 @@
...
-    else
-      node_bin = "#{MODERN_BINARY_PATH}/bin/node"
-      @fetchers[:modern].fetch_untar("#{MODERN_BINARY_PATH}.tar.gz", "#{MODERN_BINARY_PATH}/bin/node")
-      FileUtils.mv(node_bin, ".")
-      FileUtils.rm_rf(MODERN_BINARY_PATH)
+    Dir.mktmpdir do |dir|
+      node_bin = "#{binary_path}/bin/node"
+
+      Dir.chdir(dir) do
+        @fetcher.fetch_untar(@url, node_bin)
```
schneems: Why is the mktmpdir dance needed here?

lib/language_pack/ruby.rb
```diff
@@ -456,11 +464,21 @@ def install_binaries
...
   #   Example: https://s3.amazonaws.com/language-pack-ruby/node-0.4.7.tgz, where name is "node-0.4.7"
   def install_binary(name)
+    topic "Installing #{name}"
     bin_dir = "bin"
     FileUtils.mkdir_p bin_dir
     Dir.chdir(bin_dir) do |dir|
       if name.match(/^node\-/)
         @node_installer.install
+        node_bin_path = File.absolute_path(".")
+        ENV["PATH"] = "#{ENV["PATH"]}:#{node_bin_path}"
```
schneems: It's weird that this is needed. `bin/` should already be on the path.

spec/hatchet/rails51_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "spec_helper"
+
+describe "Rails 5.1" do
+  it "works with webpacker + yarn (js friends)" do
+    Hatchet::Runner.new("rails51_webpacker").deploy do |app, heroku|
+      expect(app.output).to include("Installing yarn")
+      expect(app.output).to include("yarn install")
+      expect(app.run("rails -v")).to match("")
```
schneems: We supposed to be expecting a rails version here?

Maybe we should check status of `yarn -v`?

lib/language_pack/helpers/nodebin.rb
hone: happy to make that change.

lib/language_pack/helpers/nodebin.rb
hone: yeah, I'm asking @hunterloftis for a `/lts` route or param. It'd be nice not having to keep up to date with it. /cc @jmorrell

lib/language_pack/helpers/node_installer.rb
hone: it's not looking at it now.

spec/hatchet/rails51_spec.rb
hone: sure. I just copied the rails5 test I think.

lib/language_pack/ruby.rb
hone: we need this b/c we need the absolute path during `yarn install`. some node modules using `node-gyp` will change the CWD and try to execute `node` and it won't be there.

lib/language_pack/helpers/nodebin.rb
hunterloftis: I'm down with the idea & if @jmorrell wants to implement 👍  but I wouldn't sweat the maintenance burden too much (once a year updates).

hone: I'd be happy to submit a PR too if you're down with it :) I just imagine if say the `Go` buildpack needs to start bundling yarn + node, it would be nice to have a path for it.

PR #544, opened 2017-03-10
Rails 4.x Improved Asset Pipeline Cleanup

matthewbjones:
Makes two modifications to how the `rake:precompile` step of the Heroku
slug compilation process behaves.

The first is to specify the argument `keep=0` to the `rake:cleanup` task,
which instructs the `rake:cleanup` task to keep around zero previously
compiled/fingerprinted files. The effect of this is that `public/assets/*`
will only contain the 'current' version of each asset file, without
keeping the previous three versions (the default).

The second change is to remove the `tmp/cache/assets` directory after this
directory has been cached by the Heroku build cache.

Both of these changes can drastically reduce the resulting Heroku
slug filesize.

### DIFF
--- a/lib/language_pack/rails4.rb
+++ b/lib/language_pack/rails4.rb
@@ -1,3 +1,4 @@
+require "fileutils"
 require "language_pack"
 require "language_pack/rails3"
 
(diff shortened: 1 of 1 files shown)


schneems:
> The first is to specify the argument keep=0 to the rake:cleanup task, which instructs the rake:cleanup task to keep around zero previously

This would break so many apps. We can't do that.

> The second change is to remove the tmp/cache/assets directory after this directory has been cached by the Heroku build cache.

We've talked about this before, but we can't be sure that you're not compiling dynamically in runtime. It's probably okay. However it should already be limited to 50mb of size by our LRU cache cleaner. 50mb would be nice to clean up, but it's not a problem for 99% of apps on Heroku.

Issue #543, opened 2017-03-06
Heroku CI with db/structure.sql


ombr:
Hi,

I am trying to use heroku ci, with a rails app with:

```
config.active_record.schema_format = :sql
```

so I need to run `rake db:structure:load` or `rake db:migrate` instead of `rake db:schema:load`

So I configured my app.json with:
```
  "scripts": {
    “test-setup”: “bundle exec rake db:structure:load db:migrate”,
    "test": "bundle exec rspec spec"
  },
```

But inside this buildpack the rake db:schema:load is called in `prepare_tests` is there any way to override it ?

Thanks.

Luc

schneems:
Thanks for the ticket, we've got some ideas here. In the short term if you really need to get this working you could re-define that task in your Rakefile

```ruby
task "db:schema:load" do
  Rake::Task["db:structure:load"].invoke
end
```

ombr:
Thanks, I will do that.

Oh and by the way I also had to change my `.rubocop.yml` to exclude
`lib/tasks/heroku_db_test_purge.rake` because of the double quotes. I don't know if it is an issue for you. but I did not expect the buildpack to add code to my app.

Have a nice day.

raphaelcosta:
@schneems Would not be better to remove this prepare phase from Heroku buildpack and everyone handle that using “test-setup” script?

schneems:
Ideally we want things to "just work" by default. That's a goal anyway. Give us a little bit to smooth over the sharp edges. If it's still a burden after that we can re-consider.

rlgreen91:
Was this ever decided one way or another?

Tolsto:
@schneems Any new thoughts here? We currently have to maintain a fork of this buildpack that disables the `prepare_tests` method since we have a heavy customized database setup for parallel_rspec in our test-setup script.

ClayShentrup:
> ```ruby
> task "db:schema:load" do
>   Rake::Task["db:structure:load"].invoke
> end
> ```

Lifesaver. Thanks. I just spent too many hours digging into this.

PR #546, opened 2017-03-15
Use schema.rb or structure.sql if exists

raphaelcosta:


schneems:
Sorry for the delay in following up on your issue. We can't use that method wholesale as it's not present in all versions of Rails that we support. It's on our radar though. We're a bit busy with Yarn support for Rails 5.1.

raphaelcosta:
Thanks @schneems ! I will think how to do that in a better way.

schneems:
I appreciate the commit, but you don't have to do the leg work. You can get your app working with a hack I posted in #543. We can get it working.

PR #547, opened 2017-03-15
Hardcoded nodebin

schneems:
Hardcoding nodebin results so we can ship before security review. Commits from @hone from #540


hone:
@schneems we should use s3pository instead of s3 directly

hunterloftis:
nodebin will be done w/ security review late next week. After that I'd just ask @jmorrell when any remediations are finished.

schneems:
Tests are green

josegonzalez:
Has nodebin passed security review? If so, using it to have a single source of truth for nodejs would be awesome.

PR #548, opened 2017-03-20
Do not assume lib/tasks exists

rmontgomery429:
This blows up when you don't have a `lib/tasks` folder.


schneems:
Good idea, i missed this when it came in. Thanks ❤️

Issue #551, opened 2017-03-24
Heroku CI: undefined symbol: rb_data_typed_object_zalloc


joshwlewis:
Some users will see this when Heroku CI is running their tests:

```
-----> Running Ruby buildpack tests...
/tmp/buildpacks/f6d48d8a14fccbb19c0c0402fca224929d18e9ee042b3e204bc5992612e990d3b3a0fb1f9627b1a3bae11e9fa20dffc96e136bd734a2f3cf92a0d05bedd42cfe/vendor/ruby/cedar-14/bin/ruby: symbol lookup error: /app/vendor/bundle/ruby/2.3.0/gems/json-1.8.6/lib/json/ext/generator.so: undefined symbol: rb_data_typed_object_zalloc
-----> Ruby buildpack tests failed with exit status 127
```

After a bunch of digging, it looks like these are the conditions needed to reproduce:

1. The app is not using the same ruby version that comes with the Heroku stack
2. The app has `json` in it's Gemfile
3. The app's `app.json` does not include a `scripts.test` entry

I've created a sample app that meets those conditions. You can find it at https://github.com/joshwlewis/heroku-ci-json-example. I've also setup a pipeline that showcases the failing test [here](https://dashboard.heroku.com/pipelines/6b0f6d21-ae7a-494c-87fc-9ce6ea4bbc16/tests/2).

I've traced down the error to [this line](https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/helpers/nodebin.rb#L1). 

My working theory:

1. During `test-setup` the buildpack runs `require "json"`, this does not error, because it uses the default `json` library packaged with the stack Ruby.
2. `test-setup` installs `json` (with native extensions) with the user-specified ruby version.
3. `test` (running against the stack Ruby) attempts to `require "json"`, but instead of grabbing the Ruby vendored version, it tries to load the bundled one (which was compiled against the user's specified Ruby version).

joshwlewis:
Note that users can work around this by specifying the test script in`app.json`. For example:

```json
{
  "scripts": {
    "test": "bundle exec rspec"
  }
}
```

schneems:
Thanks we can also replicate the error on a regular deploy if two ruby buildpacks are specified

```ruby
# Gemfile
source "https://rubygems.org"

ruby "2.3.3"
gem "json", "1.8.6"
```

```
# Fails
$ heroku buildpacks
[...]

PR #552, opened 2017-03-24
Fix Double Rainbow

schneems:
A "double rainbow" deploy is when someone is using the same buildpack twice. Normally this is by accident. There are failure conditions defined in #511 the app has to have a `json` gem and a version of Ruby that is different from the version of Ruby being used by the buildpacks.

When this happens, this exception will occur when the second buildpack tries to execute:

```
remote: /tmp/tmp.Se0nOtPeri/bin/ruby: symbol lookup error: /tmp/build_c85cc44940028913e25caa54b9ff2142/vendor/bundle/ruby/2.3.0/gems/json-1.8.6/lib/json/ext/generator.so: undefined symbol: rb_data_typed_object_zalloc
```

If you look at the output, the failure is happening extremely early, this indicated that it was happening before any significant code had been run, otherwise we would have expected to see

```
remote: -----> Compiling Ruby
```

The exception only occurred when using `v155` of the buildpack as the second buildpack (via `heroku buildpacks:add`). 

Here's what was happening: The first buildpack runs and installs the app's `json` gem. The buildpacks are designed so that they make their contents available for the next buildpack. I.e. if you install `node` via a buildpack, then the next buildpack to execute is expected to have `node` on the `PATH`. The same is true of ruby and any gems installed.

So now the second buildpack executes. To do this it's own source code is loaded into memory, when it does this it hits a `require "json"` call that was introduced in `v155` and tries find the `json` library. While this library ships with Ruby it is also a gem, if the gem is present on the system it will prefer the gem. So it loads the `json` gem from a different version of Ruby that is currently executing, it can't because the symbols are different, it errors out.

We can avoid this completely by not loading the `json` library, instead we can vendor in https://github.com/kr/okjson.

### DIFF
--- a/lib/language_pack/helpers/nodebin.rb
+++ b/lib/language_pack/helpers/nodebin.rb
@@ -1,4 +1,4 @@
-require 'json'
+require 'vendor/okjson'
 
 class LanguagePack::Helpers::Nodebin
   URL = "https://nodebin.herokai.com/v1/"
@@ -6,7 +6,7 @@ class LanguagePack::Helpers::Nodebin
   def self.query(q)
     response = Net::HTTP.get_response(URI("#{URL}/#{q}"))
     if response.code == '200'
-      JSON.parse(response.body)
+      OkJson.decode(response.body)
     end
   end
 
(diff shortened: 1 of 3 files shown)


joshwlewis:
Tested, it does fix the example app mentioned in #551.

joshwlewis:
I agree with the sentiment here... ugh. Weird stuff is weird. Thanks for figuring it out.

tt:
Rather than using a different gem should the buildpack try to isolate itself from changes in the build environment? Could it set the appropriate `GEM_PATH` to only look where it wants to look?

schneems:
@tt here's an implementation https://github.com/heroku/heroku-buildpack-ruby/pull/553

PR #555, opened 2017-04-04
Fix Double Ruby for CI & DRY

schneems:
We can DRY up the bash scripts a bit by putting common logic into a function. You get access to this function by sourcing `bin/support/bash_functions.sh`.

Currently only one function is offered:

```
heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$ROOT_DIR"
```

This will download a version of Ruby for the buildpack and set the appropriate env vars so that it can be run.

Now everywhere we were previously doing this manually we can re-use this bash function:

```
source "$BIN_DIR/support/bash_functions.sh"
heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$ROOT_DIR"
```

This left us with a problem since `bin/test-compile` is what set's up an app's `PATH` and `GEM_PATH`, however `bin/test` which is called after, also installs a verison of Ruby for the buildpack and to get this working correctly it mutates the `PATH` and unsets the `GEM_PATH`. To get the application code working correctly we can manually set the two values.

```ruby
LanguagePack::ShellHelpers.user_env_hash["PATH"] = "#{build_dir}/bin:#{ENV["PATH"]}"
LanguagePack::ShellHelpers.user_env_hash["GEM_PATH"] = LanguagePack::Ruby.slug_vendor_base
```

To get this code to work we have to remove support for Ruby 1.8.7 from the buildpack. This is fine since 1.8.7 is EOL for many years now and not supported on any stacks that Heroku currently maintains.

--- 

Fix tests on Travis and Circle

---

Skip running tests on "merge" commits by auto exiting 0


schneems:
The goal of this PR is to close #551

bin/support/bash_functions.sh
```diff
@@ -0,0 +1,33 @@
...
+# can be installed.
+#
+# Returns the directory where ruby was downloaded
+heroku_buildpack_ruby_install_ruby()
+{
+  local bin_dir=$1
+  local root_dir=$2
+  heroku_buildpack_ruby_dir="$root_dir/vendor/ruby/$STACK"
+
+  if [ ! -d "$root_dir/vendor/ruby/$STACK" ]; then
```
hone: can we replace this with `$heroku_buildpack_ruby_dir`?

bin/support/bash_functions.sh
```diff
@@ -0,0 +1,33 @@
...
+# and setting appropriate env vars for execution
+#
+# Example:
+#
+#   heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$ROOT_DIR"
+#
+# Takes two arguments, the first is the location of the buildpack's
+# `bin` directory. This is where the `download_ruby` script can be
+# found. The second argument is the root directory where Ruby
+# can be installed.
```
hone: Let's make sure to mention `$STACK` since we're assuming it's set.

bin/support/bash_functions.sh
```diff
@@ -0,0 +1,33 @@
+# This function will install a version of Ruby onto the
+# system for the buidlpack to use. It coordinates download
+# and setting appropriate env vars for execution
+#
+# Example:
+#
+#   heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$ROOT_DIR"
```
hone: I think `$APP_DIR` is more descriptive than `$ROOT_DIR`

bin/support/ruby_test
```diff
@@ -20,13 +24,19 @@ def execute_command(command)
...
-# point bundler wrapper to app's bundler gem
+
+# The `ruby_test-compile` program installs a version of Ruby for the
+# user's application. It needs the propper `PATH`, where ever Ruby is installed
+# we always add a symlink to the `bin/ruby` file so that is always valid.
+# We calculate the gem path the same way we do when compiling.
+LanguagePack::ShellHelpers.user_env_hash["PATH"] = "#{build_dir}/bin:#{ENV["PATH"]}"
+LanguagePack::ShellHelpers.user_env_hash["GEM_PATH"] = LanguagePack::Ruby.slug_vendor_base
+
 bundler_path = "#{build_dir}/vendor/bundle/ruby/2.3.0/gems/"
```
hone: we aren't using this anymore it looks like?

circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   ruby:
-    version: 2.0.0
+    version: 2.4.0
```
hone: This should be `2.4.1` since Travis is set to this.

spec/hatchet/ci_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe "CI" do
+  it "Does not cause the double ruby rainbow bug" do
+    Hatchet::Runner.new("heroku-ci-json-example").run_ci do |test_run|
+      # puts test_run.output
```
hone: remove this comment

bin/support/bash_functions.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+# Takes two arguments, the first is the location of the buildpack's
+# `bin` directory. This is where the `download_ruby` script can be
+# found. The second argument is the root directory where Ruby
+# can be installed.
+#
+# Returns the directory where ruby was downloaded
```
hone: we aren't actually returning anything b/c we don't `echo` at the end.

bin/support/bash_functions.sh
hone: actually n/m

PR #559, opened 2017-04-20
Blacklist JAVA_OPTS and JAVA_TOOL_OPTIONS during build

jkutner:
Like `JRUBY_OPTS`, these should be blacklisted during build to prevent errors when running `bundle` or other `ruby` commands.

### DIFF
--- a/lib/language_pack/shell_helpers.rb
+++ b/lib/language_pack/shell_helpers.rb
@@ -27,7 +27,7 @@ module LanguagePack
     end
 
     def self.blacklist?(key)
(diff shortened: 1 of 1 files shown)


jkutner:
These env vars are causing errors when we invoke a JRuby process during the build and try to capture the output.

Basically, anything that causes a JRuby process to output more than just the directory name string to stdout will break [this call to `run_no_pipe`](https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/ruby.rb#L164-L165).

jkutner:
https://support.heroku.com/tickets/470342
https://support.heroku.com/tickets/469718

PR #560, opened 2017-05-03
Ask on yarnpkg rather than yarn

strzibny:
This is because later on we actually run `yarnpkg` and not `yarn`.
So the install step can be skipped if `yarn` is available but
fail later if `yarnpkg` is not.

I run into this on dokku where in my setup the first check worked


schneems:
I'm not familiar with the distinction between the tools. I don't totally follow why yarn would be on a system but not yarnpkg.

PR #564, opened 2017-05-12
Get tests working on heroku-16

schneems:
Heroku-16 stack does not support any ruby versions prior to 2.2.


schneems:
The logic is the same for both stacks for now. The only reason i made that a cedar-14 app was because the name of the app had `193` in it, also the body of the test had a bit more logic than some of the other ones. I say we add a test if we ever work on that code, and definitely before cedar-14 goes away.

PR #565, opened 2017-05-18
move rails db schema checking to only be done for rails apps

hone:
stick to `db:schema:load` and `db:migrate` for general ruby apps if they exist


neilmiddleton:
This broke most of our CI suites that are Ruby/Rails but have no db present

schneems:
Why does your app with no database have a `db:migrate` rake task? 

If you don't have `activerecord` in your gemfile you won't have that task and it won't get executed by this code.

neilmiddleton:
I don't have a `rake db:migrate`

PR #571, opened 2017-05-22
Fix errors when rake tasks missing in CI

schneems:
There are two modes of errors that were causing failures. 

1) `db:test:purge` or other task does not exist it, yet we try to clear it.

Solution: check the task exists before trying to clear

```
if Rake::Task.task_defined?('db:test:purge')
```

2) Rails apps aren't required to have a database or activerecord.

When this happens the `db:schema:load_if_ruby` and `db:structure:load_if_sql` will fail which will cause the testpack to look up schema format and fail when it's not set.

Solution: Exit early if no `db:migrate` task is present.

```
return [] if db_migrate.not_defined?
```

### DIFF
--- a/hatchet.json
+++ b/hatchet.json
@@ -80,6 +80,7 @@
     "sharpstone/rails5_sql_schema_format",
     "sharpstone/rails_31_ruby_schema_format",
     "sharpstone/rails_31_sql_schema_format",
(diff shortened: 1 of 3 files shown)


neilmiddleton:
lgtm

schneems:
thanks for the compliance @neilmiddleton going to deploy shortly.

PR #575, opened 2017-05-30
Ruby 2.1 is no longer supported.

schneems:
Remove this extra message


hone:
@schneems we should make sure that devcenter is also up to date.

schneems:
Oldest reported version is 2.2.7 https://devcenter.heroku.com/articles/ruby-support#supported-runtimes

PR #577, opened 2017-06-02
install node if webpacker gem is present

leshik:


lib/language_pack/ruby.rb
```diff
@@ -799,7 +799,7 @@ def pg_adapters
   # @note execjs will blow up if no JS RUNTIME is detected and is loaded.
   # @return [Array] the node.js binary path if we need it or an empty Array
   def add_node_js_binary
-    bundler.has_gem?('execjs') && node_not_preinstalled? ? [@node_installer.binary_path] : []
+    bundler.has_gem?('execjs') || bundler.has_gem?('webpacker') && node_not_preinstalled? ? [@node_installer.binary_path] : []
```
schneems: Move this to be multiple lines, having a ternary operator with logic is not ideal

```
if bundler.has_gem?('execjs') || bundler.has_gem?('webpacker') && node_not_preinstalled? ?
  [@node_installer.binary_path]
else
  []
end
```

leshik:
> I don't think we would support the rest of this patch. But if you want to trim it down so it's just the logic for adding node when webpacker is present I would be 👍

Ok, I changed it. But what about caching `node_modules` and `public/packs` dirs?

leshik:
@schneems What if we also cache node_modules dir during assets:precompile? I've made corresponding changes, please review.

schneems:
If you want to caching use the node buildpack + ruby buildpack. We will detect that node is already installed and not reinstall.

leshik:
Yeah, I know. But if there is already all this nodejs logic in this buildpack, why not to cache node_modules (similar to tmp/cache or bundler)?

schneems:
We like to move extremely conservatively. Adding caching in to asset generation with Rails 4.1 doubled support ticket load. We are seein enoug webpack tickets with unexplainable behavior, we don't need to add a cache to that equation for now.

It's something to consider adding eventually but not today.

PR #579, opened 2017-06-02
Update Bundler to 1.15.1

calebthompson:
This update includes several bugfixes:

- `bundle lock --update GEM` will fail gracefully when the gem is not in the lockfile
- `bundle init --gemspec` will fail gracefully when the gemspec is invalid
- `bundle install --force` works when the gemfile contains git gems
- `bundle env` will print well-formed markdown when there are no settings

Full details: https://github.com/bundler/bundler/compare/v1.15.0...v1.15.1


schneems:
We have to stage a version of bundler on S3 before changes to the gemfile take effect https://github.com/heroku/languages-team/blob/master/languages/ruby/bundler.md

PR #584, opened 2017-06-16
Update `heroku_hatchet` to use V3 of the API

schneems:
V2 of the Heroku API is being deprecated. Hatchet 3.0.0 uses V3 of the Heroku API.


jkutner:
LGTM

PR #586, opened 2017-06-20
Warn if ruby/rails/bundle binstub shebangs are bad

calebthompson:
Binstubs, which RubyGems and Bundler generate to prepare the environment
before dispatching calls to original executables, can sometimes include
shebangs that worked on developer machines, but won't work on Heroku.
Often, this is caused by ruby binaries being named other than `ruby` (or
`ruby.exe` on Windows).

While it can happen for any binstub, we're mostly concerned about
`rake`, `rails`, and `bundler` stubs. Check each of these to see if they
are `!#/usr/bin/env ruby` or `!#/usr/bin/env ruby.exe` and issue a
warning if this is not the case.

### DIFF
--- a/lib/language_pack/installers/heroku_ruby_installer.rb
+++ b/lib/language_pack/installers/heroku_ruby_installer.rb
@@ -22,5 +22,20 @@ class LanguagePack::Installers::HerokuRubyInstaller
       @fetcher.fetch_untar(file)
     end
   end
+
+  def setup_binstubs(*)
+    super
+    Dir["#{DEFAULT_BIN_DIR}/{rake,bundle,rails}"].select do |binstub|
+      begin
+        if File.file?(binstub)
+          shebang = File.open(binstub, &:readline)
+          if !shebang.match %r{\A#!/usr/bin/env ruby(.exe)?\z}
+            warn("Binstub #{binstub} contains shebang #{shebang}. This may cause issues if the program specified is unavailable.", inline: true)
+          end
+        end
+      rescue EOFError
+      end
+    end
+  end
 end
 
--- a/spec/installers/heroku_ruby_installer_spec.rb
+++ b/spec/installers/heroku_ruby_installer_spec.rb
@@ -48,4 +48,41 @@ describe LanguagePack::Installers::HerokuRubyInstaller do
     end
 
   end
+
+  describe "#setup_binstubs" do
+    it "warns about malformed shebangs" do
+      Dir.mktmpdir do |dir|
+        Dir.chdir(dir) do
+          Dir.mkdir("bin")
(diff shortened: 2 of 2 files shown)


lib/language_pack/installers/heroku_ruby_installer.rb
```diff
@@ -22,5 +22,20 @@ def fetch_unpack(ruby_version, install_dir, build = false)
       @fetcher.fetch_untar(file)
     end
   end
+
+  def setup_binstubs(*)
```
schneems: Is this the right file to add this logic? We use this to installer to download the version of Ruby that the buildpack bootstraps with. 

It gets called here https://github.com/heroku/heroku-buildpack-ruby/blob/a46dcce3f35881dea142d609e7e42bf30ebc7549/bin/support/download_ruby#L12 

and that gets called from this script https://github.com/heroku/heroku-buildpack-ruby/blob/a46dcce3f35881dea142d609e7e42bf30ebc7549/bin/support/bash_functions.sh#L35 which is executed in a few places

https://github.com/heroku/heroku-buildpack-ruby/search?utf8=%E2%9C%93&q=heroku_buildpack_ruby_install_ruby&type=

I don't think we're inside of the right directory either.

lib/language_pack/installers/heroku_ruby_installer.rb
```diff
@@ -22,5 +22,20 @@ def fetch_unpack(ruby_version, install_dir, build = false)
...
+    super
+    Dir["#{DEFAULT_BIN_DIR}/{rake,bundle,rails}"].select do |binstub|
+      begin
+        if File.file?(binstub)
+          shebang = File.open(binstub, &:readline)
+          if !shebang.match %r{\A#!/usr/bin/env ruby(.exe)?\z}
+            warn("Binstub #{binstub} contains shebang #{shebang}. This may cause issues if the program specified is unavailable.", inline: true)
```
schneems: I would like a link to that devcenter article you wrote here. If you give me the link I can edit it and give some feedback as well.

schneems:
This is the main flow for the compile action https://github.com/heroku/heroku-buildpack-ruby/blob/cebdf6889c86fde4a22d58665e23c2b22e6a288b/lib/language_pack/ruby.rb#L91-L114

If we can't find a good place to shoe-horn in another check, one option is to add a new method like `warn_bad_binstubs`. 

The trick is we want to do this before any errors could crop up that might arise from the thing we're warning about. In this case we can inspect the contents from disk any time, so perhaps doing it before we install  ruby would be good because we use `bundle platform --update` call to determine their ruby version, and that can fail if they're using a mangled binstub.

PR #588, opened 2017-06-20
Remove dead code in language_pack/ruby.rb

### LINKED ISSUE #587: Deadcode in ruby.rb after #515
After #515, the following lines are unused:
https://github.com/heroku/heroku-buildpack-ruby/blob/4da52dad0da02d3bb53a0b5b2f07e55e8b956fec/lib/language_pack/ruby.rb#L42-L43
https://github.com/heroku/heroku-buildpack-ruby/blob/4da52dad0da02d3bb53a0b5b2f07e55e8b956fec/lib/language_pack/ruby.rb#L21

...which give the incorrect impression that Ruby binaries are being downloaded over HTTP.

edmorley:
These are unused after #515.

Fixes #587.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -18,7 +18,6 @@ class LanguagePack::Ruby < LanguagePack::Base
   LIBYAML_PATH         = "libyaml-#{LIBYAML_VERSION}"
   BUNDLER_VERSION      = "1.15.1"
   BUNDLER_GEM_PATH     = "bundler-#{BUNDLER_VERSION}"
-  RBX_BASE_URL         = "http://binaries.rubini.us/heroku"
   NODE_BP_PATH         = "vendor/node/bin"
 
   # detects if this is a valid Ruby app
@@ -39,8 +38,6 @@ class LanguagePack::Ruby < LanguagePack::Base
 
   def initialize(build_path, cache_path=nil)
     super(build_path, cache_path)
-    @fetchers[:mri]    = LanguagePack::Fetcher.new(VENDOR_URL, @stack)
-    @fetchers[:rbx]    = LanguagePack::Fetcher.new(RBX_BASE_URL, @stack)
     @node_installer    = LanguagePack::Helpers::NodeInstaller.new
     @yarn_installer    = LanguagePack::Helpers::YarnInstaller.new
     @jvm_installer     = LanguagePack::Helpers::JvmInstaller.new(slug_vendor_jvm, @stack)


edmorley:
The Travis run failed for an unrelated reason (PRs from forks do not have access to secure Travis environment variables, see also: heroku/heroku-buildpack-clojure#52, heroku/heroku-buildpack-jvm-common#28).

schneems:
I appreciate the PR, thanks for forking and opening this PR. We tend to be pretty paranoid about removing code. Our tests are not comprehensive.  I'm going to close this for now. I'll try to get your other https changes in soon though.

edmorley:
Hi - thank you for the reply, though I must admit it's left me scratching my head a little..? :-)

If this were a library that had external consumers, and so impossible to track down which APIs were used, I would totally agree about perhaps punting on the change. However this isn't a library, and this is definitely dead code - I grepped the codebase and tracked down (and linked to above) the PR where it stopped being used. (I wasn't relying on the tests to determine whether it was still in use.)

If it's a case of "not enough time to review right now" - then perhaps leaving the PR open would make more sense?

ripgrep results:
```
heroku-buildpack-ruby $ rg RBX_BASE_URL
lib\language_pack\ruby.rb
21:  RBX_BASE_URL         = "http://binaries.rubini.us/heroku"
43:    @fetchers[:rbx]    = LanguagePack::Fetcher.new(RBX_BASE_URL, @stack)
```

```
heroku-buildpack-ruby $ rg -w 'mri'
lib\language_pack\ruby.rb
42:    @fetchers[:mri]    = LanguagePack::Fetcher.new(VENDOR_URL, @stack)

spec\installers\heroku_ruby_installer_spec.rb
9:    it "should fetch and unpack mri" do

spec\helpers\ruby_version_spec.rb
89:  it "detects non mri engines" do
```

```
heroku-buildpack-ruby $ rg -w 'rbx'
CHANGELOG.md
203:* rbx is now stack aware
583:* switchover to rubinius hosted rbx binaries
956:* fix native extension building for rbx 2.0.0dev
963:* rbx 2.0.0dev support
1026:* rbx 1.2.4 support

Rakefile
80:  rubies   = document.elements.to_a("//Contents/Key").map {|node| node.text }.select {|text| text.match(/^(ruby|rbx|jruby)-\\\\d+\\\\.\\\\d+\\\\.\\\\d+(-p\\\\d+)?/) }

lib\language_pack\ruby_version.rb
43:      if rbx?
72:    # determine if we're using rbx
74:    def rbx?
75:      engine == :rbx

lib\language_pack\ruby.rb
43:    @fetchers[:rbx]    = LanguagePack::Fetcher.new(RBX_BASE_URL, @stack)

spec\installers\rbx_installer_spec.rb
5:  let(:ruby_version) { LanguagePack::RubyVersion.new("ruby-2.3.1-rbx-3.69") }
9:    it "should fetch and unpack rbx" do

lib\language_pack\installers\ruby_installer.rb
10:    if ruby_version.rbx?
```

schneems:
Unfortunately there's no way to set up travis to run for arbitrary pull requests and guarantee safety since we have to use encrypted env vars. That's why all your tests fail.

I agree that not merging this change in is overly paranoid yes. It's also my job to provide stability, which is my number one concern.

It's hard to explain if you've not maintained a high volume buildpack, but people use these things in the weirdest ways imaginable. Some of the "safest" and most "bullet proof" changes i've made have introduced some of the largest failures. Working with mature buildpacks is an exercise in patience and legacy code.

In reality it's more of a "not right now" response. I also unfortunately don't like leaving hanging issues open as it creates a lot of clutter. There's really too many right now, and I should go through and clean them up.

I think this is a really good candidate for code removal, but right now it's not a huge priority (for me personally). I'll open it back up for a bit. Maybe try to get some other eyes on it.

PR #593, opened 2017-06-27
Set JAVA_HOME env var for JRuby apps

jkutner:
This change sets the `JAVA_HOME` environment variable at build time, which enables the use of Rake tasks like [JBundler](https://rubygems.org/gems/jbundler).


jkutner:
@schneems this is ready to merge :)

hone:
@schneems is OOO, but I can merge. :)

PR #599, opened 2017-07-11
[CI Skip] heroku-buildpack-ruby v165

calebthompson:
- Increase Node version to 6.11.1
- Set the `JAVA_HOME` environment variable at build time, which enables
  the use of Rake tasks like `JBundler`.


hone:
Let's add a changelog with this PR. :)

calebthompson:
@hone CHANGELOG updated.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Master
 
+## v165 (07/11/2017)
+
+* Set `$JAVA_HOME` for JRuby apps (#593)
```
hone: "during build"

PR #602, opened 2017-07-17
Bundler 1.15.2

schneems:
Includes this fix https://github.com/bundler/bundler/pull/5707.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master
 
+* Update Bundler to 1.15.1 (#602)
+
 ## v166 (07/11/2017)
 
 * no changes from v165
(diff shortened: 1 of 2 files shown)


calebthompson:
Oh speaking of since this will probably go out as a discrete change should we add it to CHANGELOG.md?

schneems:
Yup, good call. Added.

-- 
Richard Schneeman
https://www.schneems.com


On July 17, 2017 at 4:55:36 PM, Caleb Thompson (notifications@github.com) wrote:

Oh speaking of since this will probably go out as a discrete change should we add it to CHANGELOG.md?

calebthompson:
@schneems Compliance is failing because the last commit is ci-skip, so it's "untested".

PR #608, opened 2017-07-31
install node when using execjs OR webpacker

hone:
With `webpacker` now a part of Rails 5.1, you can now get in a case where you're using `webpacker` but not using `execjs` pulled in by `uglifier`. When this happens you'll get this error message:

```
-----> Installing yarn-0.22.0
-----> Detecting rake tasks
       Running: rake assets:precompile
       /usr/bin/env: ‘node’: No such file or directory
       rake aborted!
```

This patch checks for both `execjs` or `webpacker` to see if `node` should be installed.

### DIFF
--- a/hatchet.json
+++ b/hatchet.json
@@ -74,6 +74,9 @@
   "multibuildpack": [
     "sharpstone/node_multi"
   ],
+  "node": [
+    "sharpstone/webpacker_no_execjs"
+  ],
   "ci": [
     "sharpstone/rails5_ruby_schema_format",
     "sharpstone/heroku-ci-json-example",
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -800,7 +800,11 @@ params = CGI.parse(uri.query || "")
   # @note execjs will blow up if no JS RUNTIME is detected and is loaded.
   # @return [Array] the node.js binary path if we need it or an empty Array
   def add_node_js_binary
-    bundler.has_gem?('execjs') && node_not_preinstalled? ? [@node_installer.binary_path] : []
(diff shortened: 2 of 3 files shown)


lib/language_pack/ruby.rb
```diff
@@ -800,7 +800,7 @@ def pg_adapters
   # @note execjs will blow up if no JS RUNTIME is detected and is loaded.
   # @return [Array] the node.js binary path if we need it or an empty Array
   def add_node_js_binary
-    bundler.has_gem?('execjs') && node_not_preinstalled? ? [@node_installer.binary_path] : []
+    (bundler.has_gem?('execjs') || bundler.has_gem?('webpacker')) && node_not_preinstalled? ? [@node_installer.binary_path] : []
```
schneems: Lets make this an if statement instead of a ternary. It's a bit long.

hatchet.json
```diff
@@ -74,6 +74,9 @@
   "multibuildpack": [
     "sharpstone/node_multi"
   ],
+  "node": [
+    "sharpstone/webpacker_no_execjs"
```
hone: https://github.com/sharpstone/webpacker_no_execjs

lib/language_pack/ruby.rb
hone: could also just turn `(bundler.has_gem?('execjs') || bundler.has_gem?('webpacker'))` into a method like `gem_requires_nodejs?`

calebthompson: even better

hatchet.json
calebthompson: thanks

hone:
@schneems feel free to merge when you get a chance to re-review.

schneems:
This is good, will merge in the morning when I can babysit

schneems:
Merged. Thanks!

PR #609, opened 2017-08-02
Bundler 1.15.3

paulomcnally:
From v1.15.2 to v1.15.3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master
 
+* Update Bundler to 1.15.3 (#609)
+
 ## v167 (07/25/2016)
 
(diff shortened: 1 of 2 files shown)


schneems:
Why do you want 1.15.3? Are there critical things you need?

paulomcnally:
No, it's just that in my project I use that version and I saw an alert in the deploy log.

schneems:
Bundler is pretty difficult to upgrade site wide so we generally do it very cautiously. There are lots of features of bundler that Heroku deploys do not use, so it does not make sense to risk downtime due to upgrading for features that will not be used. 

If you're hitting a bug or need a feature in a newer version of bundler please let me know.

PR #614, opened 2017-08-15
Added explicit case for JVM version 8 or 1.8 in system.properties

jkutner:
This catches cases where `system.properties` is configured to the same value as the default JDK version. It fixes the following:

https://support.heroku.com/tickets/502207
https://twitter.com/fkchang2000/status/897238534125305861


schneems:
aren't you supposed to be reviewing a 600 line pull request of mine 😜

schneems:
Change seems fine. Let's wait until morning to merge so we can catch any failures (if they come).

PR #620, opened 2017-09-13
[close rails/webpacker#739] Upgrade Yarn to 1.0.2

schneems:
Linking rails/webpacker#739

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Make installs more robust against temporary network issues by retrying failed
   downloads in `LanguagePack::Fetcher#fetch_untar`, which installs Rubies (#585)
 
+* Yarn version upgraded to v1.0.2
+
 ## v167 (07/25/2016)
 
 * Update Bundler to 1.15.2 (#602)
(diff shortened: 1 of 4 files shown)


FreekingDean:
Is there a reason to hardcode the yarn version, instead of pulling it from the packages.json as with the nodejs buildpack?

shaundr:
What are we waiting on here?

wolfpakz:
Any updates? FWIW, Yarn is now at 1.1.0.

schneems:
> Is there a reason to hardcode the yarn version, instead of pulling it from the packages.json as with the nodejs buildpack?

The support for webpacker in the Ruby buildpack is minimal. If you want full featured support you should use `heroku/nodejs` buildpack and then `heroku/ruby` buildpack via your `heroku buildpacks` command.

In general we usually don't allow arbitrary dependency manager versions because it is extremely difficult to support. For example it's much easier to release a version, verify it works and if any bugs or edge cases become known we can document them or upgrade/downgrade to a version without that bug. When you can specify your own version then it's virtually impossible for us to provide any kind of meaningful support as unexpected behavior may be expected with your version. That's why we only provide a single version of bundler etc.

If you want you can always specify your own version via forking the buildpack. It's a pretty big undertaking but if your app is truly blocked on a version of a dependency manager tool we're using at least it's an escape valve. In all I would recommend using `heroku/nodejs` buildpack instead of relying on this functionality in the Ruby buildpack. In the future we may simply execute that buildpack instead of hard coding our own functionality in.

PR #621, opened 2017-09-13
Support gems.rb/gems.locked

### LINKED ISSUE #534: "gems.rb" and "gems.locked" doesn't work
This is the prefered file name instead of `Gemfile` and `Gemfile.lock`.

calebthompson:
Resolves #534

### DIFF
--- a/lib/language_pack/no_lockfile.rb
+++ b/lib/language_pack/no_lockfile.rb
@@ -3,7 +3,7 @@ require "language_pack/base"
 
 class LanguagePack::NoLockfile < LanguagePack::Base
   def self.use?
-    !File.exists?("Gemfile.lock")
+    ["gems.locked", "Gemfile.lock"].none? { |lockfile| !File.exists?(lockfile) }
   end
 
   def name
@@ -11,6 +11,6 @@ class LanguagePack::NoLockfile < LanguagePack::Base
   end
 
   def compile
-    error "Gemfile.lock required. Please check it in."
+    error "Gemfile.lock or gems.locked required. Please check it in."
   end
 end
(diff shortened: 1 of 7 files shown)


lib/language_pack/no_lockfile.rb
```diff
@@ -3,14 +3,14 @@
 
 class LanguagePack::NoLockfile < LanguagePack::Base
   def self.use?
-    !File.exists?("Gemfile.lock")
+    ["gems.locked", "Gemfile.lock"].none? { |lockfile| !File.exists?(lockfile) }
```
knu: I think`any? { |lockfile| File.exists?(lockfile) }` will be more readable.

lib/language_pack/ruby.rb
```diff
@@ -25,12 +25,26 @@ class LanguagePack::Ruby < LanguagePack::Base
   # @return [Boolean] true if it's a Ruby app
   def self.use?
     instrument "ruby.use" do
-      File.exist?("Gemfile")
+      !gemfile.nil?
     end
   end
 
+  def self.gemfile
+    ["./gems.rb", "./Gemfile"].detect { |gemfile| File.exist?(gemfile) }
```
knu: I think `detect` should be `find` instead.

lib/language_pack/ruby.rb
```diff
@@ -25,12 +25,26 @@ class LanguagePack::Ruby < LanguagePack::Base
...
+  end
+
+  def self.lockfile
+    if lockfile = ["./gems.locked", "./Gemfile.lock"].detect { |gemfile| File.exist?(gemfile) }
```
knu: Here too.

PR #628, opened 2017-10-11
Ruby 2.5.0 and vendored Bundler

hone:
Make the Ruby buildpack work with Ruby vendored Bundler.

With Bundler being vendored in Ruby, we have no choice but to use the Ruby vendored Bundler if it's >= than the one we package in the Ruby buildpack since Ruby comes first in the `GEM_PATH` and `require 'bundler'` will always pull the latest. This still allows us the ability to update Bundler when we need to. Forcing the version of Bundler with `bundle _1.15.2_` won't be propagated during runtime. Having a different version used during build and runtime is not a good idea.

This patch works also works around the [binstub `Kernel#load` issue](https://bugs.ruby-lang.org/issues/13997) in preview1.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -540,6 +540,10 @@ WARNING
     "vendor/bundle/bin"
   end
 
+  def bundler_path
+    @bundler_path ||= "#{slug_vendor_base}/gems/#{BUNDLER_GEM_PATH}"
+  end
+
   # runs bundler to install the dependencies
   def build_bundler(default_bundle_without)
     instrument 'ruby.build_bundler' do
@@ -576,7 +580,15 @@ WARNING
           bundle_command += " --deployment"
         end
 
-        topic("Installing dependencies using bundler #{bundler.version}")
+        # If Ruby's bundler is >= buildpack version, it will win. Get the
+        # version of bundler actually being used
+        bundler_version =
+          if ruby_version.ruby_version >= "2.5.0"
+            `#{bundler_path}/exe/#{bundle_bin} -v`.downcase.chomp
+          else
+            "bundler #{bundler.version}"
+          end
+        topic("Installing dependencies using #{bundler_version}")
         load_bundler_cache
 
         bundler_output = ""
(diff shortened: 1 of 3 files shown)


lib/language_pack/ruby.rb
```diff
@@ -603,6 +603,12 @@ def build_bundler(default_bundle_without)
           }
           env_vars["BUNDLER_LIB_PATH"] = "#{bundler_path}" if ruby_version.ruby_version == "1.8.7"
           puts "Running: #{bundle_command}"
+          # bundler binstub bug in Ruby 2.5.0-preview1
+          # https://bugs.ruby-lang.org/issues/13997
+          if ruby_version.ruby_version == "2.5.0"
```
schneems: Should probably make this a ">="

hone: That's fair, I was hoping we can remove this when ruby 2.5.0 releases. Honestly, it also works for every version of ruby, so maybe we just scratch the `if`.

lib/language_pack/ruby.rb
```diff
@@ -576,7 +580,15 @@ def build_bundler(default_bundle_without)
...
-        topic("Installing dependencies using bundler #{bundler.version}")
+        # If Ruby's bundler is >= buildpack version, it will win. Get the
+        # version of bundler actually being used
+        bundler_version =
+          if ruby_version.ruby_version >= "2.5.0"
+            `#{bundler_path}/exe/#{bundle_bin} -v`.chomp
+          else
+            bundler.version
+          end
+        topic("Installing dependencies using bundler #{bundler_version}")
```
schneems:
```
$ bundle -v
Bundler version 1.15.4
```

This will give us

```
Installing dependencies using bundler Bundler version 1.15.4
```

hone: good catch, thanks will update.

calebthompson:
We should add a test for it

schneems:
I added a test

hone:
@schneems we still need to have a check for if `bundle` can't be found for some reason.

PR #631, opened 2017-10-30
Use full path when setting JAVA_HOME during bundle install

jkutner:
Using a relative path causes problems with some Gems that try to run the java command, such as the naether Gem. See ticket 523738. Added a spec for jruby with the naether Gem. This confirms the use of a full path for the jvm vendor directory.


jkutner:
Commit 8a2c44745a47f68e7fbf14b38e91b638e1685369 fixes a test that broken due to a JDK update (which happens out of cycle with the buildpack updates)

PR #638, opened 2017-12-04
Bump bundler to 1.16.0

mad-raz:
Update bundler to the latest version


hone:
Unfortunately, Bundler 1.16.0 is riddled with issues. For instance, it breaks inline Gemfile, which breaks the Rails test suite. We'll be waiting for 1.16.1 before we consider upgrading it in the buildpack.

PR #640, opened 2017-12-14
Updated Ruby 2.5.0 Support

hone:
This updates the buildpack to support Ruby 2.5.0's vendored bundler with Rails. Ruby 2.5.0 on trunk currently vendors the newly minted Bundler 1.16.1, but it breaks [the rails test suite](https://travis-ci.org/rails/rails/builds/319911098) and [rails binstubs](https://github.com/rails/webpacker/issues/998#issuecomment-353461343). Part of the buildpack's value is to alleviate users of this pain, so this patch creates a shim around Bundler so the buildpack can control the version of Bundler used. This patch lets users take advantage of Ruby 2.5.0 without the current problems with the vendored version of Bundler.

* reverts commit 26d0eddf7dcfc079e5604f0cfc07466e88a8fa49. This isn't needed anymore since https://bugs.ruby-lang.org/issues/13997 is now fixed.
* create a shim around Bundler, so every version of Ruby uses the same version of Bundler (`1.15.2`)

### DIFF
--- a/lib/language_pack/installers/ruby_installer.rb
+++ b/lib/language_pack/installers/ruby_installer.rb
@@ -24,6 +24,8 @@ module LanguagePack::Installers::RubyInstaller
     run("ln -s ruby #{install_dir}/bin/ruby.exe")
 
     Dir["#{install_dir}/bin/*"].each do |vendor_bin|
+      # for Ruby 2.5.0+ don't symlink the Bundler bin so our shim works
+      next if vendor_bin.include?("bundle")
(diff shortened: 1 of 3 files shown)


hsbt:
@hone https://travis-ci.org/rails/rails/builds/319911098 is not a bundler-1.16.1 problem. It's a Travis and rubygems issues. I did investigate its problem: https://github.com/rubygems/rubygems/pull/2124#issuecomment-353595931

And rails/rails is works with bundler-1.16.1 now. https://travis-ci.org/rails/rails/builds/320254712

PR #649, opened 2018-01-03
Only set JAVA_HOME for Bundler when using JRuby

jkutner:
The `JAVA_HOME` env var is absolute when using `heroku/jvm` buildpack, which causes this env var to be the invalid dir `/app/app/.jdk`.We should only set the env var when using JRuby.

A follow up to this PR should make the `@jvm_installer.java_home` absolute. But it's derived from `slug_vendor_jvm`, which is relative (along with all the other `vendor` references). I'm not sure if we want to make that one different.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,6 +9,8 @@ machine:
 
 test:
   pre:
+    - wget -qO- https://cli-assets.heroku.com/install-ubuntu.sh | sh
+    - heroku --version
     - bundle exec rake hatchet:setup_travis
   override:
     - bundle exec parallel_rspec -n 11 spec/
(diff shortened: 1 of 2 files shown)


watsonian:
We've disabled a regression test that's dependent on this being merged. Cross-referencing our issue to re-enable the test here so we can see when this is closed easily. https://github.com/heroku/cedar-ops/issues/521

watsonian:
What's the status on this? We have a direwolf test disabled waiting on this and it's been 3 weeks.

hone:
@watsonian sorry about that. do you need this deployed to enable direwolves or do you pull from github?

watsonian:
I think we'll need it deployed. No need to rush it out, but would love if you could post here once it's deployed. How long would we be looking at if it's not rushed out?

hone:
@watsonian I was planning on doing it on Monday if that works for you guys. I can bump this issue when it happens.

watsonian:
Perfect, thanks!

calebthompson:
@watsonian this has been released. It was on master fo ra while but if you needed it you can now use the released buildpack instead.

PR #652, opened 2018-01-17
now that Ruby 2.5.0 is out, update the default ruby to 2.4.3

hone:
also related to heroku/prodsec#244


josler:
This would be great to address vulnerabilities in 2.3.4!

schneems:
It is currently  2.4.4

PR #653, opened 2018-01-19
Execute Rake in context of current bundle

tt:
I have a Ruby 2.4.3 app that depends on Rake 10.4.2 and uses minitest.

Trying to use Heroku CI yields this error:

    -----> Running test: bin/rake test
           rake aborted!
           Gem::LoadError: You have already activated rake 12.0.0, but your Gemfile requires rake 10.4.2. Prepending `bundle exec` to your command may solve this.
           /app/vendor/bundle/ruby/2.4.0/gems/bundler-1.15.2/lib/bundler/runtime.rb:317:in `check_for_activated_spec!'
           /app/vendor/bundle/ruby/2.4.0/gems/bundler-1.15.2/lib/bundler/runtime.rb:32:in `block in setup'
           /app/vendor/bundle/ruby/2.4.0/gems/bundler-1.15.2/lib/bundler/runtime.rb:27:in `map'
           /app/vendor/bundle/ruby/2.4.0/gems/bundler-1.15.2/lib/bundler/runtime.rb:27:in `setup'
           /app/vendor/bundle/ruby/2.4.0/gems/bundler-1.15.2/lib/bundler.rb:101:in `setup'
           /app/vendor/bundle/ruby/2.4.0/gems/bundler-1.15.2/lib/bundler/setup.rb:19:in `<top (required)>'
           /app/Rakefile:1:in `<top (required)>'
           LoadError: cannot load such file -- bundler/setup
           /app/Rakefile:1:in `<top (required)>'
           (See full trace by running task with --trace)

This is due to Ruby 2.4.3 shipping with Rake 12.0.0.

The same error can be reproduced outside Heroku CI by running either `heroku run bin/rake --tasks` or `heroku run rake --tasks`. Inserting `bundle exec` resolves the issue.

(I wasn't sure whether to update both cases but it seems necessary as long as Ruby includes a Rake version.)

### DIFF
--- a/bin/support/ruby_test
+++ b/bin/support/ruby_test
@@ -53,9 +53,9 @@ execute_test(
   elsif File.exist?("bin/rails") && bundler.has_gem?("railties") && bundler.gem_version("railties") >= Gem::Version.new("5.x")
     "bin/rails test"
   elsif File.exist?("bin/rake")
-    "bin/rake test"
+    "bundle exec bin/rake test"
   else
-    "rake test"
+    "bundle exec rake test"
   end
 )
 


tt:
Assuming the buildpack always installs a stub, will it ever hit the else block? If not, adding `bundle exec` there should probably be avoided.

tt:
Another option is only inserting it if `bundler.has_gem?("rake")`.

tt:
@heroku/ruby, can you review this?

hone:
@tt the binstubs built by the buildpack should be bundler aware. What does your `bin/rake` look like?

tt:
This is the content of /app/bin/rake:

    #!/bin/sh
    # -*- ruby -*-
    # This file was generated by RubyGems.
    #
    # The application 'rake' is installed as part of a gem, and
    # this file is here to facilitate running it.
    #
    _=_\
    =begin
    bindir="${0%/*}"
    exec "$bindir/ruby" "-x" "$0" "$@"
    =end
    #!/usr/bin/env ruby
    
    require 'rubygems'
    
    version = ">= 0.a"
    
    if ARGV.first
      str = ARGV.first
      str = str.dup.force_encoding("BINARY") if str.respond_to? :force_encoding
      if str =~ /\A_(.*)_\z/ and Gem::Version.correct?($1) then
        version = $1
        ARGV.shift
      end
    end
    
    if Gem.respond_to?(:activate_bin_path)
    load Gem.activate_bin_path('rake', 'rake', version)
    else
    gem "rake", version
    load Gem.bin_path("rake", "rake", version)
    end

Even if this is supposed to be Bundler aware, the error we're getting is real.

PR #717, opened 2018-03-19
Fix buildpack logging with prefix

schneems:
- Add prefix for Ruby buildpack
- Ensure log prefix ends with a dot.

### DIFF
--- a/lib/language_pack/base.rb
+++ b/lib/language_pack/base.rb
@@ -10,6 +10,7 @@ require "language_pack/fetcher"
 require "language_pack/instrument"
 
 Encoding.default_external = Encoding::UTF_8 if defined?(Encoding)
+ENV["BPLOG_PREFIX"] = "buildpack.ruby"
 
 # abstract class that all the Ruby based Language Packs inherit from
 class LanguagePack::Base
--- a/spec/helpers/shell_spec.rb
+++ b/spec/helpers/shell_spec.rb
@@ -21,7 +21,7 @@ describe "ShellHelpers" do
         Tempfile.open("logfile.log") do |f|
           ENV["BUILDPACK_LOG_FILE"] = f.path
           FakeShell.new.mcount "foo"
-          expect(File.read(f.path)).to match("count#foo=1")
+          expect(File.read(f.path)).to match("count#buildpack.ruby.foo=1")
         end
       ensure
         ENV["BUILDPACK_LOG_FILE"] = original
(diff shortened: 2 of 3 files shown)


lib/language_pack/base.rb
```diff
@@ -10,6 +10,7 @@
 require "language_pack/instrument"
 
 Encoding.default_external = Encoding::UTF_8 if defined?(Encoding)
+ENV["BPLOG_PREFIX"] ||= "buildpack.ruby"
```
jmorrell: IIUC this won't set the prefix if it's already set, but is that what we want? Won't this be set if another buildpack has run before this?

spec/helpers/shell_spec.rb
```diff
@@ -21,7 +21,7 @@ class FakeShell
         Tempfile.open("logfile.log") do |f|
           ENV["BUILDPACK_LOG_FILE"] = f.path
           FakeShell.new.mcount "foo"
-          expect(File.read(f.path)).to match("count#foo=1")
+          expect(File.read(f.path)).to match("count#buildpack.ruby.foo=1")
```
jmorrell: Oh it's used for testing... Hmmm 🤔

lib/language_pack/base.rb
dzuelke: No because buildpacks run with isolated environments. Only the export feature can be used to forward e.g. `$PATH` for following buildpacks.

schneems: Unless you're putting it in the export folder/file directory thing that gets run between multi-buildpack builds it shouldn't be picked up by another buildpacks value.

That being said i'm fine we we want to hardcode this value to over-write what's already there I could do that, I just didn't know if there was ever a future where it could be modified by codon/etc.

spec/helpers/shell_spec.rb
schneems: This is the log file, not the prefix, if I hardcode the prefix this test would still pass.

jmorrell: 🙈 That seems obvious now

lib/language_pack/base.rb
jmorrell: Why would you want this to be configurable?

schneems: 🤷‍♂️ I originally thought the value was being set via codon/cytokine-thing-i-cannot-spell until this PR.

schneems: I updated to not be `||=`

PR #718, opened 2018-03-19
Make note of common error and fix.

schneems:


lib/language_pack/ruby.rb
```diff
@@ -920,6 +920,14 @@ def precompile_fail(output)
       msg << "Attempted to access a nonexistent database:\n"
       msg << "https://devcenter.heroku.com/articles/pre-provision-database\n"
     end
+
+    sprockets_version = bundler.gem_version('sprockets')
+    if output.match(/Sprockets::FileNotFound/) && (sprockets_version < Gem::Version.new('4.0.0.beta7') && sprockets_version > Gem::Version.new('4.0.0.beta4'))
+      msg << "If you have this file in your project\n"
```
jmorrell: It might be unclear what `this file` refers to in this message? Can you pull out the file name?

jmorrell: I would also suggest adding an `mcount` here :)

schneems: I don't want this to get too complicated I think this makes more sense in context

```
Sprockets::FileNotFound: could not find file: /tmp/dir_1/codetriage-62a1c/app/assets/images/logo.svg
# ...
If you have this file in your project
# ...
```

jmorrell: It's up to you. I think that adding some whitespace and calling attention to this as suggestion coming from Heroku and not sprockets would help users see and understand this message.

```
-------
Sprockets could not find file: /tmp/dir_1/codetriage-62a1c/app/assets/images/logo.svg

If you have this file in your project...
```

schneems: I added an mcount, also standardized all the mcount names to not add "ruby" as a duplicate.

lib/language_pack/base.rb
```diff
@@ -93,7 +93,7 @@ def compile
         puts @deprecations.join("\n")
       end
     end
-    mcount "ruby.success"
+    mcount "success"
```
jmorrell: I've been meaning to do this and compare the numbers against cytokine's librato success metric

Issue #722, opened 2018-03-23
Review apps failing due to invalid byte sequence in UTF-8 encoding error


heridev:
At this moment for me, review apps and manual deployments are failing and I'm surprised that nobody has raised this issue so far.

All tickets I have reported to Heroku about the same issue:
![46d12f4f6e2e7b4b615f980b1783cfd3 _image 202018-03-22 20at 201 22 19 20pm](https://user-images.githubusercontent.com/1863670/37805790-fa1ae8b6-2e01-11e8-85c6-dc9db011a06e.png)

And the past days is been failing with the same error:
![04652b19c13a46192e48b92e1b7fca48 _image 202018-03-22 20at 201 25 48 20pm](https://user-images.githubusercontent.com/1863670/37805805-080cb148-2e02-11e8-90d0-1bc3b5bdb29e.png)

```
       │ │ │   └── whet.extend@0.9.9
       │ │ ├─┬ postcss-unique-selectors@2.0.2
 !
 !     invalid byte sequence in UTF-8
 !
/app/tmp/buildpacks/f6d48d8a14fccbb19c0c0402fca224929d18e9ee042b3e204bc5992612e990d3b3a0fb1f9627b1a3bae11e9fa20dffc96e136bd734a2f3cf92a0d05bedd42cfe/lib/language_pack/shell_helpers.rb:135:in `split': invalid byte sequence in UTF-8 (ArgumentError)
```
or some months ago with a different dependency while precompiling assets (Friday 22nd September 2017):
```
      ??? ??? ????????? xtend@4.0.1  deduped
 !
 !     invalid byte sequence in UTF-8
 !
/app/tmp/buildpacks/f6d48d8a14fccbb19c0c0402fca224929d18e9ee042b3e204bc5992612e990d3b3a0fb1f9627b1a3bae11e9fa20dffc96e136bd734a2f3cf92a0d05bedd42cfe/lib/language_pack/shell_helpers.rb:124:in `split': invalid byte sequence in UTF-8 (ArgumentError)
```

So as you see, this is been happening from time to time, and after looking at the error, we can see that the line with the problem is:
https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/shell_helpers.rb#L135

Because of an invalid encoding.

If we add a `rescue` in that method:
```
    def puts(message)
      begin

      message.to_s.split("\n").each do |line|
        super "       #{line.strip}"
      end

      rescue Exception => e
[...]

PR #723, opened 2018-03-23
Prevents exception when logging invalid utf8 encoding messages

heridev:
At this moment Heroku review apps and manual deployments are failing and there are several tickets reported to Heroku about the same issue, in order to fix the issue a new rescue clause is added to the ShellHelper#puts method, and the right Rspec tests were added to show the error about
encoding.

closes https://github.com/heroku/heroku-buildpack-ruby/issues/722

46d12f4f6e2e7b4b615f980b1783cfd3 _image 202018-03-22 20at 201 22 19 20pm

And the past days is been failing with the same error:
04652b19c13a46192e48b92e1b7fca48 _image 202018-03-22 20at 201 25 48 20pm

```
       │ │ │   └── whet.extend@0.9.9
       │ │ ├─┬ postcss-unique-selectors@2.0.2
 !
 !     invalid byte sequence in UTF-8
 !
/app/tmp/buildpacks/f6d48d8a14fccbb19c0c0402fca224929d18e9ee042b3e204bc5992612e990d3b3a0fb1f9627b1a3bae11e9fa20dffc96e136bd734a2f3cf92a0d05bedd42cfe/lib/language_pack/shell_helpers.rb:135:in `split': invalid byte sequence in UTF-8 (ArgumentError)
```


schneems:
We use private env vars for tests. I’ll have to pull it into a commit within the buildpack to get tests to run. 

You can run them locally if you want though the Test suite creates Heroku apps against your account.

schneems:
I opened #724 to run tests. I also figured out how to get rid of the original UTF-8 error.

heridev:
Great @schneems thank you, looking forward to getting the PR merged, seems like the build is green

PR #724, opened 2018-03-23
Modify #723 to not raise an exception on a UTF8 string

schneems:
Instead of catching the exception and ignoring it, we can re-write this method to not call any methods on the original string that would cause it to raise an “invalid byte sequence”. This is done by using `each_line` instead of splitting the string. We can also avoid the `strip` by switching the `puts` to a `print` and ensuring that the printed string ends in a newline.

Just in case the original error is still caught and a better error message raised. I also added a counter to determine if the issue is actually fixed.


heridev:
👍

heridev:
screen shot 2018-03-23 at 12 49 12 pm
👍

PR #732, opened 2018-04-04
Update Default Ruby to not be an insecure version

schneems:


hone:
We should move to 2.4 soon.

schneems:
Agreed, that will take a little more work. This should do in the short term.

PR #740, opened 2018-04-16
Use nodebin s3 bucket to download node and yarn

jmorrell:


hunterloftis:
Could do the same thing for yarn: https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/helpers/nodebin.rb#L25

PR #750, opened 2018-04-24
Use Ruby 2.4.4 on Heroku 18

schneems:


lib/language_pack/ruby.rb
```diff
@@ -518,6 +518,12 @@ def load_default_cache
   # install libyaml into the LP to be referenced for psych compilation
   # @param [String] tmpdir to store the libyaml files
   def install_libyaml(dir)
+    case stack
+    when "cedar-14", "heroku-16"
+    else
```
hone: looks fine, on heroku-18
```
ii  libyaml-dev:amd64                  0.1.7-2ubuntu3                    amd64        Fast YAML 1.1 parser and emitter library (development)
```

tt: Heroku-16 has version 0.1.6. I see we're installing version 0.1.7 in the buildpack; is there a reason for that or is it just so we wouldn't have to backport security fixes ourselves? It would be great to use the stack image package consistently.

schneems: I don’t think we cache libyaml. When you upgrade stacks you should get the stack image.

The packages SE center article does not list libyaml for cedar-14. Maybe that should be updated?

tt: Oh, you obviously need the headers to. My bad. Yeah, those are not there.

tt: @schneems, do we cache which buildpack version last touched the cache? If so, we can use that to figure out when to switch to the stack image package.

schneems: We already detect when a stack changes and clear the cache https://github.com/heroku/heroku-buildpack-ruby/blob/901f2490b2e13bf41c4a827072580930850e774c/lib/language_pack/ruby.rb#L963-L971

PR #755, opened 2018-05-09
Disable listing check for disabling cops

schneems:
For a backstory see https://github.com/bbatsov/rubocop/issues/4459

You can configure rubocop to error out when anyone disables rubocop errors. Which is fine, except the Ruby buildpack needs to do that. We needed a way to disable that check, so this feature was added https://github.com/bbatsov/rubocop/pull/4938.

This PR is using that feature to hopefully allow anyone to deploy and not get errors from rubocop.

### DIFF
--- a/lib/language_pack/test/rails2.rb
+++ b/lib/language_pack/test/rails2.rb
@@ -36,6 +36,7 @@ class LanguagePack::Rails2
   def clear_db_test_tasks
     FileUtils::mkdir_p 'lib/tasks'
     File.open("lib/tasks/heroku_clear_tasks.rake", "w") do |file|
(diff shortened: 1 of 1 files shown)


litonico:
Is this patch still on the way to being merged? I just ran in to this very problem while adding rubocop to our Heroku CI setup.

In the mean time, I'll disable Lint/UnneededCopDisableDirective in our .rubocop.yml.

Thank you!

schneems:
Revert in https://github.com/heroku/heroku-buildpack-ruby/pull/922 due to comment left on https://github.com/heroku/heroku-buildpack-ruby/commit/1dc39bb0ca86bb5e3dd3287609e471f639952f95

PR #762, opened 2018-05-30
Speed up travis tests

schneems:
Cache the output of a bundle install and cache previously installed hatchet repos.

https://docs.travis-ci.com/user/caching/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 language: ruby
+cache:
+  bundler: true
+  directories:
+    - repos
 rvm:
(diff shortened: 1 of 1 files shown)


hone:
we should clean up these commits :)

schneems:
This is still a work in progress. Even though it's using the cache, using a system version of ruby instead of installing a new one, not re-installing bundler, the overall run time doesn't seem to be any faster.

I have an open support ticket with travis about this. For max speed it i'm seriously considering getting our tests running on HerokuCI.

PR #770, opened 2018-06-04
Allow for Rails Runner timeout

schneems:
It is assumed that all Rails apps can boot within 60 seconds since this is the timeout for binding to a processes. While testing this feature to a subset of Heroku users (on master branch) an issue came up where it appeared that this call would “hang” and prevent the build from succeeding. This should never happen.

To prevent an indefinite hang, a timeout is added to the `rails runner call`. If it exceeds 65 seconds the task will fail.

When this happens a warning will be emitted along with steps on how to debug (by adding `HEROKU_DEBUG_RAILS_RUNNER=1`) to config.

Tests are added to this feature to simulate actually calling a `rails` executable.

### DIFF
--- a/lib/language_pack/helpers/rails_runner.rb
+++ b/lib/language_pack/helpers/rails_runner.rb
@@ -118,13 +118,17 @@ class LanguagePack::Helpers::RailsRunner
       @success = $?.success?
 
       unless @success
-        warn("Detecting rails configuration failed\nset HEROKU_DEBUG_RAILS_RUNNER=1 to debug")
+        message = String.new("Detecting rails configuration failed\n")
+        message << "set HEROKU_DEBUG_RAILS_RUNNER=1 to debug" unless @debug
+        warn(message)
         mcount("warn.rails.runner.fail")
       end
       return out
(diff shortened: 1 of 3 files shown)


lib/language_pack/helpers/rails_runner.rb
```diff
@@ -98,16 +99,34 @@ def command
...
+
+    def execute_command!
+      out = String.new
+      Timeout.timeout(@timeout) do
+        pipe(command, user_env: true, silent: true, buffer: out)
+      end
+      @success = $?.success?
+
+      unless @success
+        warn("Detecting rails configuration failed\nset HEROKU_DEBUG_RAILS_RUNNER=1 to debug")
```
calebthompson: It looks like this will output even if you're debugging?

schneems: Good call. I changed it so i'm only modifying the message if debug is not being used.

schneems:
Looking back at this I accidentally got rid of the ability to debug the output of the runner except in the case where there is a failure. That should be added back in.

PR #784, opened 2018-06-25
Run tests on Heroku CI

schneems:


app.json
```diff
@@ -0,0 +1,29 @@
...
+    "test": {
+      "env": {
+          "HATCHET_RETRIES":    "3",
+          "IS_RUNNING_ON_CI":   "true",
+          "HATCHET_APP_LIMIT":  "80",
+          "BUILDPACK_LOG_FILE": "tmp/buildpack.log"
+      },
+      "formation": {
+        "test": {
+          "size":     "performance-m",
```
joshwlewis: You should give "performance-l" a whirl. I bet you'll cut it down to under 8 minutes.

schneems: Giving it a shot. Part of the issue with splitting things via knapsack is that it only does file based splitting, some of my files are really small and others are huge. Also each test is run in serial instead of in parallel (inside of each node). It looks like knapsack has a "pro" mode that integrates with `parallel_tests` which might be worth trying. 

The bulk of buildpack time is spent waiting for `git push heroku master` since it actually deploys to production to run tests. It makes parallelizing ideal since most time is spent in IO. If I can nodes to be better partitioned (via test and not file) and get each to run tests in parallel then We can likely get this to sub-five minutes.

schneems: 8m 22s, actually better than I would have guessed.

schneems: It could be really cool to open a port and connect them between dynos privately so that a tool like this could be used https://github.com/tmm1/test-queue that would let test jobs queue. I'm wondering if anything like this exists for a data store such as redis.

schneems: There is https://twitter.com/rafaelfranca/status/1011389040808128512 though i'm not sure if it runs multiple workers on each node. Still investigating.

joshwlewis: Yeah, that sounds interesting. Why would you prefer a test queue over knapsack, though? Adding heroku-redis for your CI runs would certainly add some complication and provisioning time (up to 3 minutes for heroku-redis).

schneems: A queue will auto load balance for me. Also I need the load to be distributed by Test and not Test file. Right now some of my Test nodes are done in 30 seconds while others take 8 minutes. If tests were distributed by Test and not by file then they should all finish at about the same time which means I’ll have higher utilization of resources and therefore faster tests.

3 minutes is a really long time! I wonder if there is a way to have a persisted redis store attached to another app and then I manually set the config vars. I’m not sure if this library expects redis to be dedicated to only one test instanciation but I’m guessing that’s not the case.

schneems:
@joshwlewis moving the conversation out of that thread because it's hidden to everyone else. I used `ci-queue` gem and now build is down to 4 minutes https://dashboard.heroku.com/pipelines/ac057663-170b-4bdd-99d0-87560eb3a570/tests/30?node=1

I could likely get even faster if they allowed me to parallelize again from within a node, but for now going from 8~9 => 4 is a pretty big win. Definitely better than the ~13-16 I was seeing before with other CI providers.

schneems:
The next test run was 4m 55s 🏃‍♂️💨💨💨💨💨

joshwlewis:
Nice! Under 5 minutes is amazing!

PR #787, opened 2018-06-30
 Add binary cache mechanism

miyucy:
In my local machine, Download from s3 is too slow. So I want to add caching mechanism.


schneems:
The cache on Heroku also comes from S3.

Maybe you can tell me more about your use case.

PR #788, opened 2018-07-03
Colorize buildpack failures

schneems:
Example output for Errors:

Example output for Warnings:

### DIFF
--- a/lib/language_pack/base.rb
+++ b/lib/language_pack/base.rb
@@ -81,7 +81,7 @@ class LanguagePack::Base
   def compile
     write_release_yaml
     instrument 'base.compile' do
-      Kernel.puts ""
+      Kernel.puts "\e[1m\e[33m" # Bold yellow
       warnings.each do |warning|
         Kernel.puts "###### WARNING:\n"
         puts warning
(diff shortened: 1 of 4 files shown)


lib/language_pack/base.rb
```diff
@@ -91,6 +91,7 @@ def compile
         topic "DEPRECATIONS:"
         puts @deprecations.join("\n")
       end
+      Kernel.puts "\e[0m"
```
sigmavirus24: Do we want a comment to indicate this is "default"?

lib/language_pack/shell_helpers.rb
```diff
@@ -62,6 +62,22 @@ def self.initialize_env(path)
...
 
+    # Should only be used once in the top level script
+    # for example $ bin/support/ruby_compile
+    def self.display_error_and_exit(e)
+      Kernel.puts "\e[1m\e[31m" # Bold Red
+      Kernel.puts " !"
+      e.message.split("\n").each do |line|
+        Kernel.puts " !     #{line.strip}"
+      end
+      Kernel.puts " !\e[0m"
```
sigmavirus24: Same question here

schneems:
Worth mentioning that I tested this on windows 8 and it works

schneems:
@dmathieu is there a reason you deleted the branch?

PR #803, opened 2018-09-02
Cache Webpacker builds between deploys

ericboehs:
## Goal
Save 5-10 minutes per deploy by using Webpacker's existing caching to skip running Webpack on builds where no changes are made.

 ## Approach
Use Heroku's deploy cache to cache Webpacker's default **public output** and **cache** paths.

 ## Context
Currently all deploys using Webpacker, unless very specifically configured, will do a full recompile on every deploy even if there are no files changed. This is because the digest file and packs directory is not cached between deploys.

[By default](https://github.com/rails/webpacker/blob/3-x-stable/lib/install/config/webpacker.yml#L6-L7), Webpacker stores [a compilation digest](https://github.com/rails/webpacker/blob/3-x-stable/lib/webpacker/compiler.rb#L88) in `tmp/cache/webpacker` and its packs in `public/packs`. [Webpacker iterates](https://github.com/rails/webpacker/blob/3-x-stable/lib/webpacker/compiler.rb#L50) over the watched paths to compute the digest based on the mtime of the files. If none of the files have changed and the packs directory is in place, running Webpack is skipped, saving a lot of time (e.g. my company's Rails app takes 8 minutes to run webpack).

 ## ~~Caveats~~
~~The environment variable `CI` needs to be set as the mtime does not seem to be preserved correctly across deploys. If `ENV['CI']` is set, Webpacker will use the file contents to compute the cache digest file instead of the files' mtimes.~~

~~I'm not familiar enough with the Heroku infrastructure to understand why the mtime isn't being preserved. [I did notice](https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/cache.rb#L53) that `cp`ing is done via `-a` which [preserves timestamps](http://www.man7.org/linux/man-pages/man1/cp.1.html) via `--preserve=all` but that doesn't seem to be enough. Perhaps there is an issue with the way Webpacker is computing the mtime digest.~~

~~I believe this is a tangential issue that should be resolved independently.~~

The CI environment variable issue has been resolved with https://github.com/rails/webpacker/commit/95c00e23b98397706bece763b3f45421821efcb7. (Thanks to @grk for pointing this out.)

 ## Further Notes
* This could technically be solved without modifying either Webpacker or the Heroku build pack (by changing the pack output and cache paths) but I thought this would be the best solution as it fixes all existing configurations provided they used the default settings. This would also make the default "just work" instead of requiring end-users to read documentation and configure their Rails apps specifically for Heroku.

* I've also tested this works correctly with version 4.0.0.pre.pre.2 of Webpacker. (Since the name of the digest file changed, it will require two deploys to realize its effect.)

* I don't mean for this pull request to be the canonical place to fix this issue but merely an issue report with a solution attached.

 ## Related
* https://github.com/rails/webpacker/issues/1347
* https://github.com/rails/webpacker/issues/1439#issuecomment-417713222
* https://github.com/rails/webpacker/issues/1611

### DIFF
--- a/lib/language_pack/rails4.rb
+++ b/lib/language_pack/rails4.rb
@@ -66,6 +66,14 @@ WARNING
     "tmp/cache/assets"
   end
 
+  def public_packs_folder
+    "public/packs"
+  end
+
+  def default_webpacker_cache
+    "tmp/cache/webpacker"
+  end
+
   def run_assets_precompile_rake_task
     instrument "rails4.run_assets_precompile_rake_task" do
       log("assets_precompile") do
@@ -80,7 +88,9 @@ WARNING
         topic("Preparing app for Rails asset pipeline")
 
         @cache.load_without_overwrite public_assets_folder
+        @cache.load_without_overwrite public_packs_folder
         @cache.load default_assets_cache
+        @cache.load default_webpacker_cache
 
         precompile.invoke(env: rake_env)
 
@@ -93,7 +103,9 @@ WARNING
 
           cleanup_assets_cache
           @cache.store public_assets_folder
+          @cache.store public_packs_folder
           @cache.store default_assets_cache
+          @cache.store default_webpacker_cache
         else
           precompile_fail(precompile.output)
         end


ericboehs:
@schneems @hone Either of you had a chance to peek at this? I think it will help your server costs quite a bit. :)

dreyks:
this would be a nice place to increase ASSETS_CACHE_LIMIT see #817

swrobel:
I hate to be the fly in the ointment, but I have a [forked buildpack](https://github.com/swrobel/heroku-buildpack-ruby/commit/ab1cf559e58c63082b6ebb07803696316dcd2004) that does this exactly and used it for a while. My slugs grew gigantic. The problem is [that there's no equivalent to `rake assets:clean` for webpacker](https://github.com/rails/webpacker/issues/1410#issuecomment-380984734). Until that's fixed, I don't see how this can be merged (and trust me, I really want to see this merged).

ericboehs:
@swrobel Interesting.

It looks like webpack has a plugin to clean the output directory called [clean-webpack-plugin](https://www.npmjs.com/package/clean-webpack-plugin). This would be the preferred way to clean the output directory.

I spent ~10 minutes and came up with a similar solution using common binaries:
```bash
pushd public; comm -3 <(awk '{print $2}' assets/packs/manifest.json | sed '/^\s*$/d' | tr -d '",' | sed 's/^/./' | sort) <(find assets/packs -type f | sed 's/^/.\//' | sort | grep -v 'manifest.json') | xargs rm; popd
```

It's a bit terse but it uses `comm` to select the old pack files and `rm` them using `xargs`.

`comm` compares lines between two common files and select/filter matches. I use `<()` to redirect the command outputs to file descriptors using process substitution. The two commands being compared: the first outputs the list of files that webpack says are in use via webpack's manifest.json; the second command list all the files in public/assets/packs (except manifest.json). It's a bit of a kludge though.

Running this command locally reduced my packs directory size by 40% (124MB).

grk:
As of https://github.com/rails/webpacker/commit/95c00e23b98397706bece763b3f45421821efcb7 the `ENV["CI"]` check no longer applies, webpacker uses file contents for all checks.

schneems:
See https://github.com/heroku/heroku-buildpack-ruby/pull/876#issuecomment-494476388

PR #805, opened 2018-09-06
Bundler 1.16.4

schneems:


Edouard-chin:
Hi @schneems , thanks for opening this PR ! Was wondering if there was something I could do to help ship this. We are mainly looking for the bundler upgrade to fix Sigterm exception  when puma gets restarted https://github.com/puma/puma/issues/1438 which is fixed in bundler 1.16.0 https://github.com/bundler/bundler/pull/6092

Happy to target this branch on one of our app to try it out if you'd like.

PR #806, opened 2018-09-11
Hide worker dyno type when no jobs:work rake task

### LINKED ISSUE #800: Don't add a default "worker" process if they don't have that task
If someone does not have the rake task "jobs:work" then do not add that as a default worker task:

```
 bundle exec rake jobs:work
```

JacobEvelyn:
Fixes #800

### DIFF
--- a/lib/language_pack/rails2.rb
+++ b/lib/language_pack/rails2.rb
@@ -48,11 +48,16 @@ class LanguagePack::Rails2 < LanguagePack::Ruby
         "bundle exec thin start -e $RAILS_ENV -p $PORT" :
         "bundle exec ruby script/server -p $PORT"
 
-      super.merge({
+      process_types = {
         "web" => web_process,
-        "worker" => "bundle exec rake jobs:work",
         "console" => "bundle exec script/console"
-      })
+      }
+
+      if has_worker_rake_task?
+        process_types.merge!("worker" => "bundle exec rake #{worker_rake_task}")
+      end
+
+      super.merge(process_types)
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/language_pack/rails2.rb
```diff
@@ -76,6 +81,14 @@ def best_practice_warnings
 
 private
 
+  def has_worker_rake_task?
```
JacobEvelyn: I couldn't find a reliable way to get the rake tasks through Ruby directly, though there probably is a way. Let me know if I'm missing something here!

lib/language_pack/rails2.rb
```diff
@@ -48,11 +48,16 @@ def default_process_types
...
 
-      super.merge({
+      process_types = {
         "web" => web_process,
-        "worker" => "bundle exec rake jobs:work",
         "console" => "bundle exec script/console"
-      })
+      }
+
+      if has_worker_rake_task?
```
JacobEvelyn: I wasn't sure if this is the sort of thing I should write a test for, and if so how to do that.

JacobEvelyn:
Superseded by #810, so I'm closing this out. Thanks @schneems!

schneems:
Thanks again for this PR. I just merged mine into master. A little about timelines. You can go ahead and use it today if you set your buildpack via the `heroku buildpack` and pointing it at `https://github.com/heroku/heroku-buildpack-ruby` instead of `heroku/ruby`.

We generally let changes sit on master for awhile (about a week) and we have a deploy embargo coming up due to dreamforce. Sitting on master allows us a form of gradual rollout as some percentage of users on the platform do deploy from master. This helps us detect any edge conditions before being rolled out to everyone.

If there's no reported issues then after the dreamforce deploy embargo then i'll roll it out to the whole platform and then (hopefully) none of your apps will ever have this issue again! Let me know if you have any questions on timeline or anything related to that process.

JacobEvelyn:
Great, thanks so much for all the info!

PR #809, opened 2018-09-13
Better failure message for CI apps using default stack

schneems:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Master
 
+## v192 (9/14/2018)
+
+* Add error messages when using unsupported Ruby versions on the Heroku 18 stack (https://github.com/heroku/heroku-buildpack-ruby/pull/809)
```
tt: "Heroku-18"

lib/language_pack/ruby.rb
```diff
@@ -380,6 +380,41 @@ def install_ruby
 
     true
   rescue LanguagePack::Fetcher::FetchError => error
+    if stack == "heroku-18" && ruby_version.version_for_download.match?(/ruby-2\.(2|3)/)
+      message = <<ERROR
+An error occurred while installing #{ruby_version.version_for_download}
+
+This version of Ruby is not available on Heroku 18. The minimum supported version
```
tt: "Heroku-18"

lib/language_pack/ruby.rb
```diff
@@ -380,6 +380,41 @@ def install_ruby
...
+of Ruby on the Heroku 18 stack can found at:
```
tt: "Heroku-18"

lib/language_pack/ruby.rb
```diff
@@ -380,6 +380,41 @@ def install_ruby
...
+
+  https://devcenter.heroku.com/articles/ruby-support#supported-runtimes
+
+ERROR
+
+      ci_message = <<ERROR
+
+If you did not intend to build your app for CI on the Heroku 18 stack
```
tt: "Heroku-18"

lib/language_pack/ruby.rb
```diff
@@ -380,6 +380,41 @@ def install_ruby
...
+More information about this change in behavior can be found at:
+  https://kb.heroku.com/3Y1HEXGJ/why-doesn-t-ruby-2-3-7-work-in-my-ci-tests
+
+ERROR
+
+      if env("CI")
+        mcount "fail.bad_version_fetch.heroku-18.ci"
+        message << ci_message
+      else
+        mcount "fail.bad_version_fetch.heroku-18"
```
tt: Maybe instead of this add a generic `mcount "fail.bad_version_fetch.#{stack}.#{ruby_version.version_for_download}"` after this block. That will give us data for future stacks without having to go back and change this.

schneems: The way we're measuring failures doesn't like reporting the same failure twice under the same key i.e. `fail`. Otherwise you can end up with failure rates that are higher than 100% we're doing that for version numbers.

PR #810, opened 2018-09-14
[close #800] Don't add a default "worker" process if they don't have that task

### LINKED ISSUE #800: Don't add a default "worker" process if they don't have that task

schneems:
This PR first detects if `jobs:work` exist before adding it as a default task.

Similar to #806 but it uses the existing RakeRunner object and also adds testing.

Special thanks to @JacobEvelyn for the original PR and the motivation.


schneems:
Needs https://github.com/heroku/heroku-buildpack-ruby/pull/809 merged

PR #831, opened 2018-12-03
Update included Node to latest LTS version

jmorrell:


schneems:
Thanks!

luccasmaso:
@schneems Is there any reason from node version not be infered from node engine version inside package.json? Adding both ruby and node buildpacks to my app in Heroku causes yarn being called twice using a rails app with Webpacker (where the second time is called by `rails assets:precompile`). 

I actually found out that https://github.com/rails/webpacker/commit/386fd850a31502df503c7b427761adedddabaaa0
resolves the problem of `yarn install` invocation when assets precompile run. The problem is that only this last `yarn install` generates a manifest file for Webpacker

`Webpacker can't find application in /app/public/packs/manifest.json. Possible causes:`

Thanks

PR #845, opened 2018-12-21
1.17.3

schneems:


jkutner:
@schneems the test uses `jruby-1.7.16.1`, which I don't think we should seek to support. But I have a fix ready for `jruby-1.7.27`

schneems:
Here's the commit i'm testing against for hatchet https://github.com/sharpstone/ruby_193_jruby_1_7_27/commit/289fb2c94ebfc69e77f95b6c9640da6e009505bd

PR #850, opened 2019-02-05
Bundler 2 support

schneems:
Detect bundler version

## Problem 1

Bundler 2.x cannot be used with a Gemfile.lock that specifies bundler 1.x:

```
BUNDLED WITH
   1.16.4
```

If you try, then Bundler 2.x will look for an installed bundler 1.x version and try to use it. If no such version exists on the system then an error is thrown:

```
can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
```

- https://github.com/rubygems/rubygems/pull/2426
- https://github.com/rubygems/rubygems/pull/2515

## Problem 2

Likewise a version of bundler 1.x cannot be used on a project where a `BUNDLED WITH` specifies 2.x:

```
$ bundle -v
1.17.3
$ cat Gemfile.lock | grep -A 1 BUNDLED WITH
BUNDLED WITH
  2.0.1
$ bundle
Traceback (most recent call last):
        2: from /Users/rschneeman/.gem/ruby/2.6.0/bin/bundle:23:in `<main>'
        1: from /Users/rschneeman/.rubies/ruby-2.6.0/lib/ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
/Users/rschneeman/.rubies/ruby-2.6.0/lib/ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.0.0) required by your /private/tmp/default_ruby/Gemfile.lock. (Gem::GemNotFoundException)
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:2.0.0`
```

This is due to a bug in Rubygems 3.x https://github.com/rubygems/rubygems/issues/2592

## Proposed Solution

This PR implements a solution where we have a different "blessed" version of bundler for each major version.

The way this is implemented is to read in a Gemfile.lock and to read in the `BUNDLED WITH` value. Then the major version is pulled out and converted into a "blessed" version.

This allows for multiple major versions of bundler to be used on Heroku without sacrificing stability.

### DIFF
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -1,16 +1,56 @@
+# frozen_string_literal: true
+
 require 'language_pack/fetcher'
 
 # This class is responsible for installing and maintaining a
 # reference to bundler. It contains access to bundler internals
 # that are used to introspect a project such as detecting presence
 # of gems and their versions.
+#
+# Example:
+#
+#   bundler = LanguagePack::Helpers::BundlerWrapper.new
+#   bundler.install
+#   bundler.version                 => "1.15.2"
+#   bundler.dir_name                => "bundler-1.15.2"
+#   bundler.has_gem?("railties")    => true
+#   bundler.gem_version("railties") => "5.2.2"
+#
+# Also used to determine the version of Ruby that a project is using
+# based on `bundle platform --ruby`
+#
+#   bundler.ruby_version # => "ruby-2.5.1"
+#
 class LanguagePack::Helpers::BundlerWrapper
   include LanguagePack::ShellHelpers
 
+  BLESSED_BUNDLER_VERSIONS = {}
+  BLESSED_BUNDLER_VERSIONS["1"] = "1.15.2"
+  BLESSED_BUNDLER_VERSIONS["2"] = "2.0.1"
+  private_constant :BLESSED_BUNDLER_VERSIONS
+
   class GemfileParseError < BuildpackError
     def initialize(error)
-      msg = "There was an error parsing your Gemfile, we cannot continue\n"
+      msg = String.new("There was an error parsing your Gemfile, we cannot continue\n")
       msg << error
+      super msg
+    end
+  end
+
+  class UnsupportedBundlerVersion < BuildpackError
+    def initialize(version_hash, major)
+      msg = String.new("Your Gemfile.lock indicates you need bundler `#{major}.x`\n")
+      msg << "which is not currently supported. You can deploy with bundler version:\n"
+      version_hash.keys.each do |v|
+        msg << "  - `#{v}.x`\n"
+      end
+      msg << "\nTo use another version of bundler, update your `Gemfile.lock` to poin\n"
+      msg << "to as supported version. For example:\n"
+      msg << "\n"
+      msg << "```\n"
+      msg << "BUNDLED WITH\n"
+      msg << "   #{version_hash["1"]}"
+      msg << "```"
       super msg
     end
   end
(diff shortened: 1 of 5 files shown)


lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -1,37 +1,89 @@
...
+      msg << "which is not currently supported. You can deploy with bundler version:\n"
+      version_hash.keys.each do |v|
+        msg << "  - `#{v}.x`\n"
+      end
+      msg << "\nTo use another version of bundler, update your `Gemfile.lock` to point\n"
+      msg << "to a supported version. For example:\n"
+      msg << "\n"
+      msg << "```\n"
+      msg << "BUNDLED WITH\n"
+      msg << "   #{version_hash["1"]}\n"
```
hone: why index 1 instead of 0?

lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -40,9 +92,9 @@ def install
 
   def clean
     ENV['BUNDLE_GEMFILE'] = @orig_bundle_gemfile
-    FileUtils.remove_entry_secure(@bundler_tmp) if Dir.exist?(@bundler_tmp)
+    @bundler_tmp.rmtree if @bundler_tmp.directory?
 
-    if LanguagePack::Ruby::BUNDLER_VERSION  == "1.7.12"
+    if version  == "1.7.12"
```
hone: should we extract this into a constant somewhere?

lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -130,4 +186,27 @@ def parse_gemfile_lock
       Bundler::LockfileParser.new(gemfile_contents)
     end
   end
+
+  def major_bundler_version
+    # https://rubular.com/r/uuRpai9IheL68d
+    bundler_version_match = @gemfile_lock_path.read.match(/^BUNDLED WITH$(\r?\n)   (?<major>\d*)\.\d*\.\d*/m)
```
hone: Should these be `+` and not `*`?

hone: Also, scraping `BUNDLED_WITH` seems awful :(

lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -130,4 +186,27 @@ def parse_gemfile_lock
...
+
+    if bundler_version_match
+      bundler_version_match[:major]
+    else
+      "1"
```
hone: Should we actually fail instead of defaulting?

lib/language_pack/helpers/bundler_wrapper.rb
schneems: This is the major version. So this will be "1.15.2"

lib/language_pack/helpers/bundler_wrapper.rb
schneems: Based on this code I don't think it's possible to ever actually trigger this case. Since we hard code a version and we're not using "1.7.12" it won't ever fire.

Let me know if you agree and I can remove it.

lib/language_pack/helpers/bundler_wrapper.rb
schneems: People can still deploy without having a `BUNDLED WITH` in their Gemfile.lock. For example in some of our hatchet test apps https://github.com/sharpstone/default_ruby/blob/master/Gemfile.lock.

lib/language_pack/helpers/bundler_wrapper.rb
schneems: > Should these be `+`

Great call.

> Also, scraping `BUNDLED_WITH` seems awful :(

Agreed, though It's the least awful thing I could come up with :(

PR #854, opened 2019-02-13
Ignore invalid byte encodings when detecting rails config

schneems:
Previously if a customer had an invalid byte sequence emitted during boot up, it would raise an error:

```
invalid byte sequence in UTF-8
```

Which would prevent configuration detection from functioning correctly. 

When detecting rails configuration, we know that the input we are trying to detect for success will be UTF-8. As a result, we can ignore any other encoded characters.

This PR prevents an invalid byte sequence from blocking rails configuration detection.

Reference internal support ticket number: 681812

Here's a blog post on invalid byte sequences if you're interested https://thoughtbot.com/blog/fight-back-utf-8-invalid-byte-sequences.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Master
 
 
+* Ignore invalid byte encodings when detecting rails config (https://github.com/heroku/heroku-buildpack-ruby/pull/854)
+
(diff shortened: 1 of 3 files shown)


schneems:
This has been deployed

PR #858, opened 2019-02-27
Build an export file in CI, too

joshwlewis:
Without an export file, subsequent testpacks can't use Ruby. 

This [example repo](https://github.com/joshwlewis/heroku-ruby-then-node-build-script) will not compile on CI or pass tests when using the `heroku/ruby` buildpack. The example repo expects `bundle` and `ruby` to be available on the `$PATH` during the node buildpack `bin/test-compile` and `bin/test` stages. With this change, the example repo builds and passes tests on CI.


schneems:
This has been deployed

PR #861, opened 2019-03-08
Set memory default for Node builds

### LINKED ISSUE #859: Set memory default for Node builds
In the Node buildpack we set an ENV var that works around Node's default 1.5 GB memory limit and let's it address all 2.5GB of the memory in the build dynos: https://github.com/heroku/heroku-buildpack-nodejs/blob/master/lib/environment.sh#L32

However if a user is using Ruby and webpacker to run their build, this does not persist for the Ruby buildpack execution.

This means that there are occasionally users who hit the limit and their build will fail:

```
remote:        <--- Last few GCs --->        
remote:        [7875:0x34cd670]   486451 ms: Mark-sweep 1331.0 (1440.7) -> 1317.8 (1442.2) MB, 1564.5 / 0.0 ms  (average mu = 0.140, current mu = 0.049) allocation failure scavenge might not succeed        
remote:        [7875:0x34cd670]   488071 ms: Mark-sweep 1332.5 (1442.2) -> 1319.4 (1443.7) MB, 1534.7 / 0.0 ms  (average mu = 0.098, current mu = 0.053) allocation failure scavenge might not succeed        
remote:        <--- JS stacktrace ---> 
```

Currently the workaround is setting the ENV var manually:

```
$ heroku config:set NODE_OPTIONS="--max_old_space_size=2560" -a $APP_NAME
```

but this is a simple enough thing that I feel it would be valuable to add it to the Ruby buildpack

jmorrell:
Fixes #859 

My attempt at trying to address #859. I'm mostly guessing how to do this and happy to revise.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -58,13 +58,6 @@ class LanguagePack::Ruby < LanguagePack::Base
     instrument "ruby.default_config_vars" do
       vars = {
         "LANG" => env("LANG") || "en_US.UTF-8",
-        # By default Node can address 1.5GB of memory, a limitation it inherits from
-        # the underlying v8 engine. This can occasionally cause issues during frontend
-        # builds where memory use can exceed this threshold.
-        #
-        # This passes an argument to all Node processes during the build, so that they
-        # can take advantage of all available memory on the build dynos.
-        "NODE_OPTIONS" => env("NODE_OPTIONS") || "--max_old_space_size=2560"
       }
 
       ruby_version.jruby? ? vars.merge({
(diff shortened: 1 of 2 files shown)


lib/language_pack/ruby.rb
```diff
@@ -302,6 +302,14 @@ def setup_language_pack_environment
       ENV["PATH"] += ":#{node_preinstall_bin_path}" if node_js_installed?
       ENV["PATH"] += ":#{yarn_preinstall_bin_path}" if !yarn_not_preinstalled?
 
+      # By default Node can address 1.5GB of memory, a limitation it inherits from
+      # the underlying v8 engine. This can occasionally cause issues during frontend
+      # builds where memory use can exceed this threshold.
+      #
+      # This passes an argument to all Node processes during the build, so that they
+      # can take advantage of all available memory on the build dynos.
+      ENV["NODE_OPTIONS"] ||= "--max_old_space_size=2560"
```
schneems: Is this a user set-able option? If so we need to use lowercase `env()` function instead of `ENV`.

PR #864, opened 2019-03-26
Warn when app is not using the latest Ruby version

schneems:
Developers should be using the latest Ruby version for their minor 
series. For example currently developers using 2.4.x should be using 
2.4.7, if not then this PR will give them a warning:

```
There is a more recent Ruby version available for you to use:

2.4.8

The latest version will include security and bug fixes, we always recommend
running the latest version of your minor release.

See https://devcenter.heroku.com/articles/ruby-versions for all available versions.
```

Since it is relatively expensive to check each version serially, we can 
instead queue up 5 version checks at a time using threads. If the last 
version check shows that it is a valid version, then it will enqueue another 
round of 5 checks in parallel.

We use similar logic for detecting EOL versions. For example when 
someone is using 2.2 then we traverse the s3
repository, 2.3.0, 2.4.0, 2.5.0, 2.6.0, 2.7.0, etc. In this case there
are more than 3 more recent versions so 2.2 is for sure EOL and we
recommend that people update to the 3rd most recent version in the
above case it would be 2.5.0.

We also have to account for the case that there is a major version bump,
so in addition to the minor version check we will check to see if 3.0.0
is released. If it has been then we will also check for 3.1.0, 3.2.0,
3.3.0, etc.

There is a problem with relying on deployed Ruby versions on S3 to
indicate EOL support since pre-released binaries are not marked. For
example 2.7.0 is currently in "preview" mode, however the binary is at
2.7.0. So when someone using 2.4.x ruby executes the EOL check it will
see that 2.5, 2.6, and 2.7 are out. Because of this we have 2 different
EOL warnings, one for "maybe" eol that warns the version they are on
either is already EOL or is about to become EOL at the next release.

For example in the future when 2.7.0 is released and someone is using
2.5.0 then they will start getting a warning when the 2.8.0 (or 3.0.0)
preview version is released even though 2.5.0 would still technically be
supported until the full release at Christmas. This is reasonable as it
gives people a heads up that they need to upgrade soon. They would keep
getting this "maybe" warning until the preview of 2.9.0 (or 3.1.0) would
be released. At that time they would start to receive the more severe
EOL message.


lib/language_pack/helpers/outdated_ruby_version.rb
```diff
@@ -0,0 +1,150 @@
...
+    check_eol_versions_minor
+    self
+  end
+
+  def join
+    return false if current_ruby_version.patchlevel_is_significant?
+    return false if current_ruby_version.rbx?
+    return false if current_ruby_version.jruby?
+
+    @threads.each(&:join)
```
jkutner: how much does doing this in parallel actually save? It creates a lot to think about (and a lot that could go wrong)

lib/language_pack/ruby.rb
```diff
@@ -108,9 +108,13 @@ def compile
       end
       config_detect
       best_practice_warnings
+      warn_outdated_ruby
       cleanup
       super
     end
+  rescue => e
+    warn_outdated_ruby
```
jkutner: `warn_outdated_ruby` twice?

lib/language_pack/helpers/outdated_ruby_version.rb
schneems: It's a non-trivial amount of time if we're doing it in serial. Here's a benchmark script you can run on heroku:

```
$ heroku run bash -a heroku-buildpack-ruby
~$ irb -I ./lib

require 'language_pack'
ENV["STACK"] = "heroku-18"
pack = LanguagePack::Ruby.new(".", "/dev/null")
fetcher = pack.instance_variable_get("@fetchers")[:mri]

exists_time = Benchmark.measure do
  fetcher.exists?("ruby-2.4.0.tgz")
end.real

does_not_exist_time = Benchmark.measure do
  fetcher.exists?("ruby-2.4.0asdfasdfasdf.tgz")
end.real

raw_time  = Benchmark.measure do
`bash -c set -o pipefail; curl -L --fail --retry 5 --retry-delay 1 --connect-timeout 3 --max-time 30 -I https://s3-external-1.amazonaws.com/heroku-buildpack-ruby/heroku-18/ruby-2.4.0.tgz`
end.real

puts "Time to check binary that exists:         #{exists_time}"
puts "Time to check binary that DOES NOT exist: #{does_not_exist_time}"
```


```
Time to check binary that exists:         0.1962961310055107
Time to check binary that DOES NOT exist: 0.2097533739870414
```

So if we're checking 5+ urls then we're adding a extra second to build time. These results are way worse locally since S3 may be far away, the worst I saw was 1.5 seconds for the case where the binary does not exist. Technically that shouldn't be a concern, but fast tests locally is pretty nice.

lib/language_pack/ruby.rb
schneems: I could maybe add a comment here. But in the event that a build fails the result might be because you're using old Ruby versions and upgrading them might help to solve your issue. The `warn_outdated_ruby` method is memoized so that it will only fire once.

Essentially if your project makes it to the first `warn_outdated_ruby` invocation and then later gets an error, it won't fire again. If it gets an error before the normal warning location then it should still fire 

```ruby
    @warn_outdated ||= begin
       @outdated_version_check.join

        warn_outdated_minor
       warn_outdated_eol
       true
     end
```

jkutner: 👍

lib/language_pack/helpers/outdated_ruby_version.rb
jkutner: I would personally probably eat the seconds, and do it serial. But I also don't see anything wrong here. It's read-only right?

schneems: > It's read-only right? 

These threads do write to data structures, but they're doing so in threadsafe ways with MRI and i'm purposefully not firing this logic on jruby. My last class at georgia tech involved doing a lot of parallel computation in C, i'm reasonably confident in the threadsafety here and we have tests that exercise all of this logic.

lib/language_pack/helpers/outdated_ruby_version.rb
```diff
@@ -0,0 +1,150 @@
...
+#   #=> "ruby-2.2.10"
+class LanguagePack::Helpers::OutdatedRubyVersion
+  DEFAULT_RANGE = 1..5
+  attr_reader :current_ruby_version
+
+  def initialize(current_ruby_version: , fetcher:)
+    @current_ruby_version = current_ruby_version
+    @fetcher      = fetcher
+
+    @minor_versions = [current_ruby_version]
```
hone: I believe we should be using [`Concurrent::Array`](http://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Array.html) here since it's being accessed in the thread itself. Last I checked Ruby's memory model wasn't well defined, so it may not be thread safe to share memory in the thread.

lib/language_pack/helpers/outdated_ruby_version.rb
jkutner: I didn't realize the threads were sharing memory. I think this might need some changing. Another option better than raw Threads and fork-join is [Concurrent::Future](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Future.html)

lib/language_pack/helpers/outdated_ruby_version.rb
schneems: Each single operation on the array is atomic in MRI due to the GIL as long as we're not relying on something like `+=` on the shared object we're fine.

The two operations we're using are single MRI instructions:

```
@eol_versions << version
@minor_versions[i] = version
```

The first one `@eol_versions << version` is atomic due to the GIL and the order of entries doesn't matter since we later sort it.

The last one `@minor_versions[i] = version` is protected against races since each thread will have a unique integer so no two threads will try to write to the same memory.

jkutner: I still think it's better to use higher level constructs over lower level ones like `Thread`. As things change and people add new things to this, it will ensure this code continues to work.

schneems: We actually don't have the capacity to use native ruby gems with the buildpack, we would have to vendor and maintain all of Concurrent Ruby. We've previously done this with `lpxc.rb` though this is a much smaller project.

This is what the diff would look like using concurrent Ruby:

```diff
diff --git a/lib/language_pack/helpers/outdated_ruby_version.rb b/lib/language_pack/helpers/outdated_ruby_version.rb
index 268f37b..2323e76 100644
--- a/lib/language_pack/helpers/outdated_ruby_version.rb
+++ b/lib/language_pack/helpers/outdated_ruby_version.rb
@@ -23,7 +23,7 @@ class LanguagePack::Helpers::OutdatedRubyVersion

     @minor_versions = [current_ruby_version]
     @eol_versions = []
-    @threads = []
+    @futures = []
   end

   # Enqueues checks in the background
@@ -41,7 +41,7 @@ class LanguagePack::Helpers::OutdatedRubyVersion
     return false if current_ruby_version.rbx?
     return false if current_ruby_version.jruby?

-    @threads.each(&:join)
+    @futures.each(&:value!)

     true
   end
@@ -96,7 +96,7 @@ class LanguagePack::Helpers::OutdatedRubyVersion
   # value in the series is found
   private def check_minor_versions(range: DEFAULT_RANGE, base_version: current_ruby_version, &block)
     range.each do |i|
-      @threads << Thread.new do
+      @futures << Concurrent::Future.execute do
         version = base_version.next_logical_version(i)
         next if !@fetcher.exists?("#{version}.tgz")

@@ -116,7 +116,7 @@ class LanguagePack::Helpers::OutdatedRubyVersion
   #   - 2.8.0
   private def check_eol_versions_minor(range: DEFAULT_RANGE, base_version: current_ruby_version)
     range.each do |i|
-      @threads << Thread.new do
+      @futures << Concurrent::Future.execute do
         version = base_version.next_minor_version(i)
         next if !@fetcher.exists?("#{version}.tgz")

@@ -137,7 +137,7 @@ class LanguagePack::Helpers::OutdatedRubyVersion
   #   - 3.2.0
   #   - 3.3.0
   private def check_eol_versions_major
-    @threads << Thread.new do
+    @futures << Concurrent::Future.execute do
       version = current_ruby_version.next_major_version(1)

       next if !@fetcher.exists?("#{version}.tgz")
```

The only thing that futures buy us here is that they use a thread pool rather than spawning up default threads. However I would rather not use a thread pool here, we're not using any resource limited features like database connections and the number of threads has a natural upper bound (since iteration stops when a version that does not exist is reached). I'm not sure what the default future thread pool count is but we could look it up.

When I experimented with webrick https://schneems.com/2017/08/01/is-webrick-webscale/ I saw it spike up to 101 concurrent spawned threads. 

The other thing Concurrent Ruby Future could potentially give us is if we needed to peek in on the status of an individual job i.e. `pending?`.

lib/language_pack/helpers/outdated_ruby_version.rb
schneems: Moving convo over to https://github.com/heroku/heroku-buildpack-ruby/pull/864#discussion_r329653044

lib/language_pack/helpers/outdated_ruby_version.rb
hone: Can we do a fan out/fan in strategy instead like a map/reduce? `Thread#value` returns the value of the block from the `Thread`, so we can construct the array data structure sequentially once the work is computed in parallel. While, your code actually looks ok, I think this strategy is a good pattern to reduce the cognitive overload on someone reading/touching this code in the future.

jkutner: a `Future` also doesn't required shared memory

jkutner: You can use a [SimpleExecutorService](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/SimpleExecutorService.html) with the `Future` if you don't want a pool

schneems: > a Future also doesn't require shared memory

What do you mean? You mean you want the values to be stored in the future instead of a shared array? That's similar to the change Terence requested, I've updated the PR to do that instead.

> Can we do a fan out/fan in strategy instead like a map/reduce?

Done, ready for another round of reviews.

hone:
This looks much clearer to me. I wonder if we can take this out once we automate binary builds for ruby, so that file isn't out of date.

PR #874, opened 2019-05-14
Set JRUBY max heap similarly to JVM common buildpack logic

partounian:
A rather simple calculation, but better than the current method, because it takes in consideration the total memory rather than putting a random value per `ulimit -a`.

 Most notably this should fix mem issues with people using P-M dynos, and improve ram util for P-L.

Tested command on a P-L one-off instance.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -219,34 +219,48 @@ WARNING
   # default JAVA_OPTS
   # return [String] string of JAVA_OPTS
   def default_java_opts
-    "-Xss512k -XX:+UseCompressedOops -Dfile.encoding=UTF-8"
+    '-Dfile.encoding=UTF-8'
+  end
+
+  def set_jvm_stack_size
+    <<-EOF
+    JVM_STACK_SIZE=512
+    EOF
   end
 
   def set_jvm_max_heap
     <<-EOF
 case $(ulimit -u) in
-256)   # 1X Dyno
-  JVM_MAX_HEAP=384
+512)   # 2X, private-s: memory.limit_in_bytes=1073741824
+  JVM_MAX_HEAP=671
   ;;
-512)   # 2X Dyno
-  JVM_MAX_HEAP=768
-  ;;
-16384) # IX Dyno
+16384) # perf-m, private-m: memory.limit_in_bytes=2684354560
   JVM_MAX_HEAP=2048
   ;;
-32768) # PX Dyno
-  JVM_MAX_HEAP=5120
+32768) # perf-l, private-l: memory.limit_in_bytes=15032385536
+  JVM_MAX_HEAP=12288
+  ;;
+*) # Free, Hobby, 1X: memory.limit_in_bytes=536870912
+  JVM_MAX_HEAP=300
   ;;
 esac
-EOF
+    EOF
+  end
+
+  def set_java_stack_size
+    <<-EOF
+if ! [[ "${JAVA_OPTS}" == *-Xss* ]]; then
+  export JAVA_OPTS="${JAVA_OPTS} ${JAVA_STACK:--Xss${JVM_STACK_SIZE:-512}k}"
+fi
+    EOF
   end
 
   def set_java_mem
     <<-EOF
 if ! [[ "${JAVA_OPTS}" == *-Xmx* ]]; then
-  export JAVA_MEM=${JAVA_MEM:--Xmx${JVM_MAX_HEAP:-384}m}
+  export JAVA_OPTS="${JAVA_OPTS} ${JAVA_MEM:--Xmx${JVM_MAX_HEAP:-300}m}"
 fi
-EOF
+    EOF
   end
 
   def set_default_web_concurrency
@@ -268,8 +282,6 @@ case $(ulimit -u) in
   export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-6144}
   export WEB_CONCURRENCY=${WEB_CONCURRENCY:-16}
   ;;
-*)
-  ;;
 esac
 EOF
   end
@@ -286,6 +298,12 @@ EOF
     "-Xmx${JVM_MAX_HEAP:-384}m"
   end
 
+  # default Java Xss
+  # return [String] string of Java Xss
+  def default_java_stack_mem
+    "-Xss${JVM_STACK_SIZE:-512}k"
+  end
+
   # sets up the environment variables for the build process
   def setup_language_pack_environment
     instrument 'ruby.setup_language_pack_environment' do
@@ -294,6 +312,10 @@ EOF
         ENV["JAVA_MEM"] = run(<<-SHELL).chomp
 #{set_jvm_max_heap}
 echo #{default_java_mem}
+SHELL
+        ENV["JAVA_STACK"] = run(<<-SHELL).chomp
+#{set_jvm_stack_size}
+echo #{default_java_stack_mem}
 SHELL
         ENV["JRUBY_OPTS"] = env('JRUBY_BUILD_OPTS') || env('JRUBY_OPTS')
         ENV["JAVA_HOME"] = @jvm_installer.java_home
@@ -326,6 +348,8 @@ SHELL
       if ruby_version.jruby?
         add_to_export set_jvm_max_heap
         add_to_export set_java_mem
+        add_to_export set_jvm_stack_size
+        add_to_export set_java_stack_size
         set_export_default "JAVA_OPTS",  default_java_opts
         set_export_default "JRUBY_OPTS", default_jruby_opts
       end
@@ -348,6 +372,8 @@ SHELL
       if ruby_version.jruby?
         add_to_profiled set_jvm_max_heap
         add_to_profiled set_java_mem
+        add_to_profiled set_jvm_stack_size
+        add_to_profiled set_java_stack_size
         set_env_default "JAVA_OPTS", default_java_opts
         set_env_default "JRUBY_OPTS", default_jruby_opts
       end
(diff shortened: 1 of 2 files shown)


partounian:
Can't figure out but I keep getting this error when deploying with my fork of this buildpack
```
remote: Compressing source files... done.
remote: Building source:
remote:
remote: -----> JVM Common app detected
remote: -----> Installing JDK 1.8... done
remote: -----> Compiling Ruby/Rails
remote: -----> Using Ruby version: ruby-2.2.3-jruby-9.0.5.0
remote: -----> Using pre-installed JDK
remote: Error: Could not find or load main class expr:
remote:
remote:  !
remote:  !     Problem detecting bundler vendor directory:
remote:  !
remote:
remote:  !     Push failed
```

schneems:
cc @jkutner

jkutner:
`ulimit -u` isn't random on Heroku and we can't depend on `` in all environments. However, we should probably update this to match the [jvm-common buildpack](https://github.com/heroku/heroku-buildpack-jvm-common/blob/master/opt/jvmcommon.sh#L8-L19)

```
  limit=$(ulimit -u)
  case $limit in
  512)   # 2X, private-s: memory.limit_in_bytes=1073741824
    echo "$opts -Xmx671m -XX:CICompilerCount=2"
    ;;
  16384) # perf-m, private-m: memory.limit_in_bytes=2684354560
    echo "$opts -Xmx2g"
    ;;
  32768) # perf-l, private-l: memory.limit_in_bytes=15032385536
    echo "$opts -Xmx12g"
    ;;
  *) # Free, Hobby, 1X: memory.limit_in_bytes=536870912
    echo "$opts -Xmx300m -Xss512k -XX:CICompilerCount=2"
    ;;
esac
```

lib/language_pack/ruby.rb
```diff
@@ -224,21 +224,8 @@ def default_java_opts
...
-  ;;
-16384) # IX Dyno
-  JVM_MAX_HEAP=2048
-  ;;
-32768) # PX Dyno
-  JVM_MAX_HEAP=5120
-  ;;
-esac
-EOF
+    JVM_MAX_HEAP=$(expr $(awk '/MemTotal/ {print $2}' /proc/meminfo) / 1024 / 4 \* 3)
```
jkutner: we should update this to match https://github.com/heroku/heroku-buildpack-jvm-common/blob/master/opt/jvmcommon.sh#L8-L19

partounian:
@jkutner My mistake, I thought I had run `ulimit -u` on one-off P-M and P-L with the same result. Tried it now and I can see I was wrong.

I've updated the PR to match your suggestion, and tried some bonus stuff for setting a stack size. Forgive me if I've done something dumb, this is one of my first PRs to a bigger project.

lib/language_pack/ruby.rb
```diff
@@ -348,6 +372,8 @@ def setup_profiled
       if ruby_version.jruby?
         add_to_profiled set_jvm_max_heap
         add_to_profiled set_java_mem
+        add_to_export set_jvm_stack_size
+        add_to_export set_java_stack_size
```
jkutner: I think these should be `add_to_profiled`. The `export` script is for multi-buildpack support.

partounian:
Please review the updates @jkutner

lib/language_pack/ruby.rb
```diff
@@ -219,34 +219,46 @@ def ruby_version
...
   ;;
 esac
-EOF
+    EOF
+  end
+
+  def set_java_stack_size
+    <<-EOF
+if ! [[ "${JAVA_OPTS}" == *-Xss* ]]; then
+  export JAVA_STACK=${JAVA_STACK:--Xss${JVM_STACK_SIZE:-512}k}
```
jkutner: should this be `export JAVA_OPTS=${JAVA_OPTS:--Xss${JVM_STACK_SIZE:-512}k}`?

I don't see where `JAVA_STACK` (and thus the -Xss` option) gets added to `JAVA_OPTS` or the command.

partounian: Updated, sorry this file bamboozles me, probably need to refactor it

jkutner:
@partounian thanks for working through this!

schneems:
Thanks for all the work here! I'll try to merge in next week during a week day, people deploy off of master and I treat merging as a deployment, I don't deploy on fridays ;)

partounian:
No problem! I'm glad to be of help, and super exciting to have a PR merged
into heroku.

My current env vars are unruly

Please excuse any typos as I'm sending this message from my phone.

schneems:
Our systems won't let us merge in untested code and our encrypted vars are limited to only branches of the same repo. I pulled your commit into https://github.com/heroku/heroku-buildpack-ruby/pull/874 for testing, that will be what I merge in.

schneems:
Getting some failures on the branch

```
  1) Ruby Versions on cedar-14 should deploy jruby with the naether gem
     Failure/Error:
       app.deploy do |app|
         expect(app.output).to match("Installing naether")
         expect(app.output).not_to include("An error occurred while installing naether")
       end
     Hatchet::App::FailedDeploy:
       Could not deploy 'hatchet-t-74199b411b' (jruby_naether) using 'Hatchet::GitApp' at path: './repos/jruby/jruby_naether'
        if this was expected add `allow_failure: true` to your deploy hash.
       output:
       Buildpack: nil
       Repo: https://git.heroku.com/hatchet-t-74199b411b.git
       remote: Compressing source files... done.        
       remote: Building source:        
       remote: 
       remote: -----> Compiling Ruby        
       remote: -----> Using Ruby version: ruby-2.3.3-jruby-9.1.12.0        
       remote: -----> Installing JVM: openjdk-8        
       remote: Error: Could not find or load main class bash:        
       remote:         
       remote:  !        
       remote:  !     Problem detecting bundler vendor directory:        
       remote:  !        
       remote: 
       remote:  !     Push failed        
       remote: Verifying deploy...        
       remote: 
       remote: !	Push rejected to hatchet-t-74199b411b.        
       remote: 
       To https://git.heroku.com/hatchet-t-74199b411b.git
        ! [remote rejected] master -> master (pre-receive hook declined)
       error: failed to push some refs to 'https://git.heroku.com/hatchet-t-74199b411b.git'
     # ./vendor/bundle/ruby/2.5.0/gems/heroku_hatchet-4.0.12/lib/hatchet/git_app.rb:11:in `push_without_retry!'
     # /vendor/bundle/ruby/2.5.0/gems/heroku_hatchet-4.0.12/lib/hatchet/app.rb:239:in `block in push'
     # ./vendor/bundle/ruby/2.5.0/gems/rrrretry-1.0.0/lib/rrrretry.rb:41:in `retry'
     # ./vendor/bundle/ruby/2.5.0/gems/heroku_hatchet-4.0.12/lib/hatchet/app.rb:237:in `push'
     # ./vendor/bundle/ruby/2.5.0/gems/heroku_hatchet-4.0.12/lib/hatchet/app.rb:228:in `block in deploy'
     # ./vendor/bundle/ruby/2.5.0/gems/heroku_hatchet-4.0.12/lib/hatchet/app.rb:215:in `block (
```

This looks like the main issue?

```
Error: Could not find or load main class bash:   
```

jkutner:
Looks similar to the error posted earlier `Error: Could not find or load main class expr`. @partounian do you recall how that was fixed?

partounian:
@jkutner sorry I'm not sure what is causing that

lib/language_pack/ruby.rb
```diff
@@ -219,34 +219,46 @@ def ruby_version
   # default JAVA_OPTS
   # return [String] string of JAVA_OPTS
   def default_java_opts
-    "-Xss512k -XX:+UseCompressedOops -Dfile.encoding=UTF-8"
+    '-Dfile.encoding=UTF-8'
+  end
+
+  def set_jvm_stack_size
+    ENV['JVM_STACK_SIZE'] = env('JVM_STACK_SIZE') || 512
```
schneems: Here's the bug. This method `set_jvm_stack_size` returns back a string of "512"

Then later this is called here:

```
        ENV["JAVA_STACK"] = run(<<-SHELL).chomp
#{set_jvm_stack_size}
echo #{default_java_stack_mem}
```

it ends up being a raw `512` on that line which bash interprets as a variable, one that does not exist. It causes this `JAVA_STACK` to be set to:

```
bash: 512: command not found
-Xss512k
```

In my debugging branch, i changed those `run` invocations to `run!` so they will error early, but it didn't trigger. Is that because there's no `set -e` at the  top of this script @jkutner ?

@partounian what do you want to do to fix this up? Is the goal simply to set that environment variable? If so, we could run it outside of the shell script, alternatively we could return a nil?

schneems: Also FWIW you should be able to run this test locally and see it fail:

```
$ rspec ./spec/hatchet/rubies_spec.rb:72
```

schneems: I think the return of `set_jvm_stack_size` needs to be a valid bash script as it's later added to profile.d

```
add_to_profiled set_jvm_stack_size
```

lib/language_pack/ruby.rb
```diff
@@ -294,6 +310,10 @@ def setup_language_pack_environment
         ENV["JAVA_MEM"] = run(<<-SHELL).chomp
 #{set_jvm_max_heap}
 echo #{default_java_mem}
+SHELL
+        ENV["JAVA_STACK"] = run(<<-SHELL).chomp
+#{set_jvm_stack_size}
```
jkutner: should this be `set_java_stack_size`? that's the name of the method on 248

partounian:
Seems to work now, ran/running on our prod

jkutner:
I wasn't able to figure out the `Error: Could not find or load main class bash` problem on CI, so i created #902, which is a bit simpler. If you need to override Xss you can still set it explicitly.

PR #875, opened 2019-05-14
Use last line of bundler platform --ruby

halogenandtoast:
For determining the ruby version when `bundler platform --ruby` is used,
no longer take into account all output from the command. Instead use the
final line which should container the ruby version.

This addresses issues where Gemfiles may contain debugging information
or processes that produce output on STDOUT, and are then unable to to
build due to using an inccrect ruby version name.


schneems:
This is perfect, thanks! Looks like tests are being flappy. I restarted them.

schneems:
Tests are passing. I'll merge after saku embargo is lifted.

schneems:
🎉 🎉 🎉  thanks again 🎉 🎉 🎉

Issue #451, opened 2016-01-21
Rails asset cache not persisting across builds


ericboehs:
As of [sprockets-rails 3.0.0](https://github.com/rails/sprockets-rails/commit/80164cb0b3eae7dc8a6efab7553a7a8ab4e20c10#diff-b295f4f07d2e863f15e53eb2047cfd1f) (released on Dec 17th), the default asset cache is stored in `tmp/cache/sprockets` rather than `tmp/cache/assets`.

From what I can tell, all Rails builds that are using sprockets-rails 3 are not caching assets between deploys.

To fix this the `default_assets_cache` needs to be changed to `"tmp/cache/sprockets"` in [rails4.rb#L66](https://github.com/heroku/heroku-buildpack-ruby/blob/613defce8ec18b6416d527d2183db93cf8702b28/lib/language_pack/rails4.rb#L66).

I'd submit a PR but I'm not sure if there are other places that need to be fixed or how to go about running the test suite. And for the fact that we'll probably want to conditionally check which version of sprockets-rails they're using.

schneems:
Good catch, thanks! How did you find that? I can't believe I (or anyone else) never noticed this before. Sprockets 3 has been out for a pretty long time. I think we should detect sprockets version and return a different cache location based on version.

ericboehs:
We recently ported a project's front end over to rails. When our deploys started taking 6 minutes instead of 30 seconds I had to figure this out.

After lots of debugging (3 hours actually), I found this solution.

Originally I thought it had to do with the buildpack not caching our browserify-rails stuff. But it turns out we just updated to sprockets 3 when browserify-rails came out. I also thought it had to do with us using multiple build packs. Neither of which were the problem.

After forking the ruby buildpack and `puts`ing everything I could think of, I finally figured it out after getting it to reproduce locally. Removing the `tmp` directory before running the precompile broke it. I thought, that must be what's happening on heroku! The tmp directory isn't being persisted.

schneems:
It's not sprockets. It is something else. Probably `sprockets-rails`

Using the same app and only updating the sprockets version:

```
$ rm -rf tmp/cache/
$ RAILS_ENV=production bundle exec rake assets:precompile
$ bundle show sprockets
/Users/richardschneeman/.gem/ruby/2.3.0/gems/sprockets-2.12.4
$ ls tmp/cache/assets/production/sprockets/
# ...

$ rm -rf tmp/cache/
[...]

PR #876, opened 2019-05-17
Cache Webpacker builds between deploys, redux

qrush:
Continuing from #803. I was seeing very slow builds on our Webpack/er 4 setup:

`remote:        Asset precompilation completed (362.53s)`

Now with @ericboehs's patch:

`remote:        Asset precompilation completed (96.99s)`

I'll take that speedup over nothing!

💥 ⚡️

### DIFF
--- a/lib/language_pack/rails4.rb
+++ b/lib/language_pack/rails4.rb
@@ -66,6 +66,14 @@ WARNING
     "tmp/cache/assets"
   end
 
+  def public_packs_folder
+    "public/packs"
+  end
+
+  def default_webpacker_cache
+    "tmp/cache/webpacker"
+  end
+
   def cleanup
     super
     return if assets_compile_enabled?
@@ -87,7 +95,9 @@ WARNING
         topic("Preparing app for Rails asset pipeline")
 
         @cache.load_without_overwrite public_assets_folder
+        @cache.load_without_overwrite public_packs_folder
         @cache.load default_assets_cache
+        @cache.load default_webpacker_cache
 
         precompile.invoke(env: rake_env)
 
@@ -100,7 +110,9 @@ WARNING
 
           cleanup_assets_cache
           @cache.store public_assets_folder
+          @cache.store public_packs_folder
           @cache.store default_assets_cache
+          @cache.store default_webpacker_cache
         else
           precompile_fail(precompile.output)
         end


ericboehs:
Two things to note:

1. Webpacker can be configured to use the existing paths that Heroku caches 
2. Webpacker does not clean old assets so the slug size will balloon. Additional code is needed to clear the cache after it grows over a certain size.

Also, this looks the same as my patch.

qrush:
@ericboehs yep I just rebased it - you still have the commit author on this.

That's a good tip re existing paths - how are you configuring that in your env?

ericboehs:
Well, @swrobel [added cache cleanup](https://github.com/swrobel/heroku-buildpack-ruby/commit/ab1cf559e58c63082b6ebb07803696316dcd2004) to his buildpack. That would probably be a good idea.


You can also manually run `heroku repo:purge_cache`.

Personally I've implemented https://github.com/rails/webpacker/pull/1744. I need to find some time to add tests so that it can be merged.

schneems:
Hello 👋 . Here's some history that may not seem related at first, but it gets there eventually.

Sprockets introduced caching fairly early no, Heroku did not support it though because it was not 

- https://github.com/heroku/heroku-buildpack-ruby/issues/451#issuecomment-254888758

Here's an example of people tacking on their own caching solution https://blog.alexmaccaw.com/faster-deploys. It got so bad at one point that I messaged Alex and asked him to take down the blog post. While it made deploys faster, the resulting volume of support tickets from people who were seeing caching bugs was rather large.

Essentially: Even though sprockets supported caching out of the box, the experience was not fully baked so we didn't add default support for it until Rails 4.0 was released.

Even then, since i've taken over maintenance of the sprockets gem, the number one thing I spend my time doing is diagnosing and fixing caching bugs. Sprockets is faster than webpacker, partially because it has caching baked in, however that came at an extremely steep cost. The trajectory of the project has ground to a stand-still in part due to the complexity and difficulty of maintaining a cache-aware plugin eco-system and codebase.

One of the biggest issues is that sprockets does not fully own all the code it executes since it has a plugin architecture. It is difficult enough to get all the caching and cache invalidation code in sprockets working correctly, but to add on to that, each and every processor plugin that you use also has to be cache aware and do all the right things for your app to function.

Here's a recent bug that I chased for over a year and finally fixed https://blog.heroku.com/debugging-year-old-sprockets-bug. It turns out that it was the result of me fixing another caching issue. It's caching issues all the way down.

Even recently there have been major bugs found and some features are very difficult to support in a cache aware way such as ERB templates which can change based on things that don't involve the mtime of the asset such as environment variables or other data-stores.

This is all to say, caching is really hard, asset caching is especially hard, and without full community support from not just the tools (webpack and webpacker), but also all the extensions/plugins in the ecosystem then, I don't think it's a good idea to opt people into this experience that is likely to break in unexpected ways by default.

I'm happy to explore ways that we can better integrate with webpack(er). I would defer to the nodejs buildpack and their expertise with webpack before considering adding any kind of caching support cc @jmorrell.

ericboehs:
I think this is a reasonable response. I'm using the vanilla ruby buildpack and I'm able to cache webpack with the aforementioned patch to webpacker. Once the webpack(er) tooling is solidified with caching support then we can consider a seamless integration for users.

jmorrell:
Unless something has changed recently, my understanding was that the main issue with caching webpack artifacts was that they relied on absolute paths when serializing. This breaks during Heroku's build step which occurs in a different random directory for each build.

ericboehs:
I'm able to cache compiled packs that don't change between deploys. So, for example, if I push a simple db migration that doesn't touch my js, the packs from the previous deploy will be reused. If, however, I do make a change that touches my js, a full repack is required and there is no time savings.

Maybe you're referring to the latter scenario? I'm not familiar with using this sort of caching but would absolutely love to speed up my full-repack builds.

schneems:
I'm unfamiliar with the internals of webpack caching, but to be truly effective you can't just cache the generated assets, you need to cache intermediate files, for example, if one file is built up of 10 dependencies, and one of those dependencies changes, you don't want to rebuild all 10 of them, just the one that changed. In sprockets, this is called fragment caching.

For several years, there was a bug in sprockets that prevented fragments from being re-used due to the absolute path issue that Jeremy is referencing. I wrote about it at length for sprockets here https://blog.heroku.com/speeding-up-sprockets.

> the main issue with caching webpack artifacts was that they relied on absolute paths when serializing

The biggest issue for me personally, is the lack of unified management experience. For sprockets includes that we have an `assets:clean` task that handles deletion and removal of stale fragments.

> Webpacker does not clean old assets so the slug size will balloon. Additional code is needed to clear the cache after it grows over a certain size.

That being said

> Webpacker can be configured to use the existing paths that Heroku caches

If you configure webpacker to use the existing cache directory that sprockets uses, then we should clean up fragments for you, we have a task that truncates the cache at 50mb.

PR #881, opened 2019-05-24
fix vendored ruby is not in PATH for CNB v3

neerfri:
This fixes a problem where the `ruby` binary that is vendored in the `heroku_ruby` layer is not used by default.
This causes many native gems to fail with `"incompatible library version"` error.
Tto reproduce the problem run `pack build` in a ruby project that has `mysql2` gem in it's Gemfile. Then run `docker run --rm my-built-image bundle exec ruby -e "require 'mysql2'"`

The reason for this error is that ruby's [`dln_incompatible_library_p`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/dln.c#L1252-L1256) ensures that the MRI binary used and the dynamic library loaded have the exact same `ruby_xmalloc`.

~The solution implemented here "exports" the relevant parts of the `PATH` (those starting with `<layer-dir>`) to the `env.launch/PATH` file so it is appended properly as specific by the [CNB spec](https://github.com/buildpack/spec/blob/37ba4d9/buildpack.md#provided-by-the-buildpacks)~

The change adds the same path that is used during install (in `#setup_ruby_install_env`) to the `PATH` written to `profile.d`

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -9,7 +9,7 @@ BIN_DIR=$(cd $(dirname $0); pwd)
 BUILDPACK_DIR=$(dirname $BIN_DIR)
 
 # legacy buildpack uses $STACK 
-STACK=$PACK_STACK_ID
+STACK=$CNB_STACK_ID
 
 source "$BIN_DIR/support/bash_functions.sh"
 heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$BUILDPACK_DIR"
--- a/bin/build
+++ b/bin/build
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+LAYERS_DIR=$1
+PLATFORM_DIR=$2
+ENV_DIR="$PLATFORM_DIR/env"
+PLAN=$3
+APP_DIR=$(pwd)
+BIN_DIR=$(cd $(dirname $0); pwd)
+BUILDPACK_DIR=$(dirname $BIN_DIR)
+
+# legacy buildpack uses $STACK 
+STACK=$PACK_STACK_ID
+
+source "$BIN_DIR/support/bash_functions.sh"
+heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$BUILDPACK_DIR"
+
+$heroku_buildpack_ruby_dir/bin/ruby $BIN_DIR/support/ruby_build $APP_DIR $LAYERS_DIR $PLATFORM_DIR $PLAN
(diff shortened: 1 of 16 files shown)


schneems:
Thanks for the PR!

/cc @hone wdyt? It might take a bit of time, he's traveling now. Terence wrote all that integration, so he's the best to review this.

neerfri:
@schneems thanks for your response.

I think I may have found a more clean way to do this using the `profile.d` feature since this is the one currently setting the `PATH`. see https://github.com/heroku/heroku-buildpack-ruby/blob/41e1a97901cba675ad12344d1245389aec44912a/lib/language_pack/ruby.rb#L400-L402

Working on the change now. I'll push it here

neerfri:
TL;DR; This should NOT be merged for now.

While trying to come up with simple steps to reproduce I found that the buildpack symlinks the `ruby` binary to `bin/ruby`. So everything works well without this PR, just not in my app :trollface: 

In my app, for some cosmic reason, these symlinks are not present in the final image, although they are present when the buildpack finishes (I've added a `ls -al bin` in `bin/build` to verify).
I'm trying to figure out why. Looks like a deep rabbit hole.

As a side thought, Does it even make sense to symlink between layers? shouldn't they be somewhat "independent"

neerfri:
I've tracked the problem I had with the `bin/ruby` symlink. Obviously it was a silly-right-under-your-node kind of problem, I've used docker-compose to run the image with all it's dependencies and that docker-compose file mounted the `bin` folder for some reason.

The reason this mount existed is for developers to be able to develop locally (in desktop IDE) while the tests & server are running in a container (as well as the DBs). This makes for an easier on boarding of developers and easier upgrades of stack elements. In this case syncing `bin` is really not so important so it's not a blocker for me.


I think it would be wise to still merge this PR since it makes the right ruby binary to be used at all times. For example when changing the current working directory:
(with`heroku-buildpack-ruby-app` being an image built using the ruby buildpack)

```
$ docker run --rm heroku-buildpack-ruby-app "which -a ruby"
bin/ruby
/usr/bin/ruby
$ docker run --rm heroku-buildpack-ruby-app "cd / && which -a ruby"`
/usr/bin/ruby
```

neerfri:
@schneems, following your rebase of the `cnb` branch I've rebased this PR too and squashed everything to 1 commit.

As I wrote above I believe it's a good addition to the ruby buildpack since users of the buildpack probably expect their ruby binary (and version) to be the one they installed in the buildpack and not the system default ruby. Currently if any code changes `PWD` it will break.

PR #885, opened 2019-05-24
Schneems/standard 2

schneems:


schneems:
Annnnd, forgot that cedar-14 still exists which is running on Ruby 1.9.3

schneems:
Ruby buildpack is updated to no longer depend on the stack image, it now uses bash to bootstrap instead of Ruby

schneems:
This will need to wait until #888 which is quite large

PR #890, opened 2019-06-06
Log test order

schneems:
Feature introduced in https://github.com/Shopify/ci-queue/pull/99

This will output a failing test order if the suite fails.


schneems:
Naturally my test runs are passing now...

PR #894, opened 2019-06-19
Rev Bundler 2 to 2.0.2

schneems:
Due to https://github.com/bundler/bundler/issues/7205

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Set memory default for Node builds (https://github.com/heroku/heroku-buildpack-ruby/pull/861)
 * Default Ruby version is now 2.5.5, was previously 2.5.3 (https://github.com/heroku/heroku-buildpack-ruby/pull/863)
 * Default Node version is now 10.15.3 and default Yarn version is now 1.16.0 (https://github.com/heroku/heroku-buildpack-ruby/pull/884)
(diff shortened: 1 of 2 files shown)


schneems:
I added docs for this behavior in number 4 of the "known issues" here https://devcenter.heroku.com/articles/bundler-version#bundler-2-0-1

Here is an inline copy:

----

4) A Gemfile.lock that specifies bundler 2.0.2 does not work with bundler 2.0.1

If you attempt to deploy an app that uses bundler 2.0.2 onto the Heroku platform with bundler 2.0.1 you may get this error:

```
`find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
```

This is due to a bug in the Rubygems bundler version checking code. It was fixed in 2.7.10 and 3.0.4 however those versions have not been bundled with a new release of a Ruby version and no patch for older rubybgems versions has been provided.

* https://blog.rubygems.org/2019/06/14/3.0.4-released.html
* https://blog.rubygems.org/2019/06/14/2.7.10-released.html

The short term fix is to set your Gemfile.lock to use bundler version 2.0.1. In the medium term, the version of bundler has been changed to 2.0.2 and merged into master (but not yet deployed) https://github.com/heroku/heroku-buildpack-ruby/pull/894.

PR #907, opened 2019-08-19
Add support to Rails 6 by adding a new Language Pack

rafaelfranca:
The current Rails 5 language pack work for Rails 6 applications but it could not be used because of the version constraints. This was making Rails 6 apps to be handled as generic rack apps and causing problems at deploy if no SECRET_KEY_BASE was set.

By adding a new class we can update the constraint so new Rails 6 apps can work as in the pre-releases.

@hone @schneems

### DIFF
--- a/lib/language_pack/rails6.rb
+++ b/lib/language_pack/rails6.rb
@@ -1,0 +1,16 @@
+require 'securerandom'
+require "language_pack"
+require "language_pack/rails5"
+
+class LanguagePack::Rails6 < LanguagePack::Rails5
+  # @return [Boolean] true if it's a Rails 6.x app
+  def self.use?
+    instrument "rails6.use" do
+      rails_version = bundler.gem_version('railties')
+      return false unless rails_version
+      is_rails = rails_version >= Gem::Version.new('6.x') &&
+        rails_version < Gem::Version.new('7.0.0')
+      return is_rails
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/language_pack/rails6.rb
```diff
@@ -0,0 +1,16 @@
+require 'securerandom'
+require "language_pack"
+require "language_pack/rails5"
+
+class LanguagePack::Rails6 < LanguagePack::Rails5
+  # @return [Boolean] true if it's a Rails 6.x app
+  def self.use?
+    instrument "rails5.use" do
```
Edouard-chin:
```suggestion
    instrument "rails6.use" do
```

lib/language_pack/rails6.rb
```diff
@@ -0,0 +1,16 @@
...
+require "language_pack/rails5"
+
+class LanguagePack::Rails6 < LanguagePack::Rails5
+  # @return [Boolean] true if it's a Rails 6.x app
+  def self.use?
+    instrument "rails6.use" do
+      rails_version = bundler.gem_version('railties')
+      return false unless rails_version
+      is_rails = rails_version >= Gem::Version.new('6.x') &&
+        rails_version <  Gem::Version.new('7.0.0')
```
Edouard-chin:
```suggestion
        rails_version < Gem::Version.new('7.0.0')
```

schneems:
I deployed this. Should be rolled out to our fleet soon.

rafaelfranca:
Thank you!

Edouard-chin:
Thanks for the quick merge !

PR #914, opened 2019-09-06
clean old bundler binstubs on load

hone:
It looks like there's a bug with `bundle clean`, where it doesn't clean up binstubs that aren't used. This patch removes the `vendor/bundle/bin` directory on cache load. Bundler will regenerate the binstubs even if no new gems need to be installed.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -1015,6 +1015,9 @@ params = CGI.parse(uri.query || "")
       rubygems_version_cache  = "rubygems_version"
       stack_cache             = "stack"
 
+      # bundle clean does not remove binstubs
+      FileUtils.rm_rf("vendor/bundler/bin")
+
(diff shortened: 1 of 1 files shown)


schneems:
Internal support reference https://heroku.support/764500

The broad issue was the customer used to have the `heroku` gem in their gemfile which generated a binstub. They then later added the heroku CLI buildpack which should give them access to a `heroku` command, however the binstub from the heroku gem still took priority, even once the gem was removed from the Gemfile and re-deployed.

I'm going to try to reproduce this behavior with a dockerfile so I can report to bundler and we can get it fixed upstream as well.

schneems:
Waiting on CI to have a successful run, also re-visiting https://github.com/heroku/heroku-buildpack-ruby/pull/890 to help with making progress on the flappy tests.

schneems:
Merging in, will let it sit on master for a bit before deploying.

Also, I'm realizing we didn't do a changelog. I'll make another PR to add one, but I would LOVE a bot that checked for the presence of a changelog entry.

PR #935, opened 2019-11-26
Vendor libpq5.12.1 into Slugs

schneems:
Libpq is the client library used to communicate to postgresql. There was a bug in libpq 5.11 that essentially said if there is a problem with a configuration value, for example if you tried to set `pool` (which should be an integer) to a string, then the value would be silently ignored. In libpq 5.12 this behavior was "fixed" so that if you try to set a value to an incorrect type it will raise an error:

```
PG::UnableToSend: no connection to the server
```

Or 

```
CONNECTION_BAD
```

Here is the commit that changed the behavior https://github.com/postgres/postgres/commit/e7a2217978d9cbb2149bfcb4ef1e45716cfcbefb

The largest failure mode is when someone accidentally forgets to put a space after their ERB tag in `database.yml` but before their comment. For example:

```
connect_timeout: <%= ENV['DB_CONNECT_TIMEOUT'] || 5 %># seconds
```

Would be evaluated to a string of "5# seconds". This will work in libpq5.11 but fail in libpq5.12. This can be fixed by adding a space between `%>` and `# seconds`.

This change was rolled out on our stack image, but then rolled back to customers getting failures in the middle of the night when they had seemingly changed nothing. 

By vendoring this library in the slug, we have more control around when the update will happen. Rather than breaking changes when the stack image rolls out which might happen in the middle of the night for some developers, this change will only be applied as customers are deploying their apps. If they have a release phase then the error will likely be caught there and prevented from going into production.

This behavior introduces a helpful warning message along with URLs for getting more information around this issue.

We do not anticipate that a majority of applications will be affected by this change, however those that are affected experience large problems. The intent of rolling this change out in a buildpack is to minimize the exposure of apps that will trigger this failing condition.

Eventually libpq5.12.1 will be the default version on the stack. This change works to prepare developers for that change.

## TODO

- [x] Changelog
- [ ] Devcenter article explaining the problem

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -164,7 +164,7 @@ WARNING
     EOF
 
     Dir.chdir("vendor") do
-      run!("curl --remote-name-all http://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-12/#{pkg}")
+      run!("curl --remote-name-all https://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-12/#{pkg}")
       run!("dpkg -x #{pkg} .")
 
       load_libpq_12_unless_env_var = <<~EOF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -147,7 +147,8 @@ WARNING
       This version includes a bug fix that can cause an exception
       on boot for applications with incorrectly configured connection
       values. For more information see:
-        <url>
+
+        https://devcenter.heroku.com/articles/libpq-5-12-1-breaking-connection-behavior
 
       If your application breaks you can rollback to your last build.
       You can also temporarially opt out of this behavior by setting:
@@ -158,7 +159,8 @@ WARNING
 
       In the future libpq 5.12 will be the default on the platform and
       you will not be able to opt-out of the library. For more information see:
-        <url>
+
+        https://devcenter.heroku.com/articles/libpq-5-12-1-breaking-connection-behavior
     EOF
 
     Dir.chdir("vendor") do
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -99,6 +99,7 @@ WARNING
       setup_export
       setup_profiled
       allow_git do
+        patch_libpq
         install_bundler_in_app
         build_bundler("development:test")
         post_bundler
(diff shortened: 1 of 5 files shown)


lib/language_pack/ruby.rb
```diff
@@ -117,6 +118,66 @@ def compile
...
+      ```
+
+      In the future libpq 5.12 will be the default on the platform and
+      you will not be able to opt-out of the library. For more information see:
+
+        https://devcenter.heroku.com/articles/libpq-5-12-1-breaking-connection-behavior
+    EOF
+
+    Dir.chdir("vendor") do
+      run!("curl --remote-name-all http://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-12/#{pkg}")
```
edmorley: As discussed on Slack it would be good to download this from S3 rather than a third party domain that may be down during the customer's build. Ideally the S3 upload should be performed by a script in this repo that can be reviewed and so be compliant. In that script I would also like for us to download from `https://apt.postgresql.org` rather than the non-HTTPS URL :-)

lib/language_pack/ruby.rb
```diff
@@ -117,6 +118,66 @@ def compile
...
+    out = run!("ls /usr/lib/x86_64-linux-gnu/ | grep libpq")
+    out.each_line do |line|
+      version = line.sub(/libpq\.so\./, '')
+      next unless version.match?(/\Ad/) # avoid libpq.so.libpq.a
+      return if Gem::Version.new(version) >= Gem::Version("5.12")
+    end
+
+    case ENV['STACK']
+    when 'heroku-16'
+      pkg = "libpq5_12.1-1.pgdg16.04+1_amd64.deb"
```
edmorley: Are old package point release versions ever deleted from the APT repo? Glancing at https://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-12/ and similar from older major releases it seems the answers is "no", but it's perhaps worth checking in more detail, since this has bitten us in the past with Canonical's APT repos.

lib/language_pack/ruby.rb
```diff
@@ -99,6 +99,7 @@ def compile
       setup_export
       setup_profiled
       allow_git do
+        patch_libpq
```
dzuelke: `patch` isn't the right word IMO, make it `vendor_libpq_v12` instead maybe?

lib/language_pack/ruby.rb
```diff
@@ -117,6 +118,66 @@ def compile
     raise e
   end
 
+  def patch_libpq
+    # Check for existing libraries
+    out = run!("ls /usr/lib/x86_64-linux-gnu/ | grep libpq")
```
dzuelke: This can be simplified to `run!("ls /usr/lib/x86_64-linux-gnu/libpq.so.5.11")` and catching the error, since we only want to apply the logic if 5.11 is there.

lib/language_pack/ruby.rb
dzuelke: On Ubuntu's PPAs, they are, yup, which is what's bitten us in the past.

lib/language_pack/ruby.rb
dzuelke: Yes, we have to mirror to our S3 bucket rather than downloading straight off a third party server.

lib/language_pack/ruby.rb
```diff
@@ -117,6 +118,66 @@ def compile
...
+
+      In the future libpq 5.12 will be the default on the platform and
+      you will not be able to opt-out of the library. For more information see:
+
+        https://devcenter.heroku.com/articles/libpq-5-12-1-breaking-connection-behavior
+    EOF
+
+    Dir.chdir("vendor") do
+      run!("curl --remote-name-all http://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-12/#{pkg}")
+      run!("dpkg -x #{pkg} .")
```
dzuelke: IMO the `usr` prefix is unnecessary in the path structure (and the 64 bit subdir too), but that's just cosmetic.

genslein:
from the pq ruby gem, I'd expect to see this as the current behavior with the updated messaging from 12.1 libpq https://github.com/ged/ruby-pg/blob/master/sample/async_api.rb#L28

schneems:
The existing code doesn't appear to work on heroku-16 and I don't know why

```
  1) Ruby apps vendoring libpq works on heroku-16
     Failure/Error: expect(out).to match(%Q{invalid integer value "15s"})
       expected "Succeeded\n" to match "invalid integer value \"15s\""
       Diff:
       @@ -1,2 +1,2 @@
       -invalid integer value "15s"
       +Succeeded
     # ./vendor/bundle/ruby/2.5.0/gems/rspec-support-3.8.2/lib/rspec/support.rb:97:in `block in <module:Support>'
     # ./vendor/bundle/ruby/2.5.0/gems/rspec-support-3.8.2/lib/rspec/support.rb:106:in `notify_failure'
```

PR #936, opened 2019-12-09
Vendor libpq5.12.1 into Slugs - Heroku 18 only

schneems:
Pick up from the work in https://github.com/heroku/heroku-buildpack-ruby/pull/935 but removing Heroku-16 support until we can get that working in Docker locally 

https://github.com/schneems/libpq_heroku_16_reproduction/tree/schneems/manually-download-install

(Note it's a branch)


edmorley:
Filed https://gus.my.salesforce.com/a07B00000079prxIAA for removing the vendoring now that the stack image has the latest libpq :-)

PR #946, opened 2020-01-22
Default WEB_CONCURRENCY and RAILS_MAX_THREADS for new apps

schneems:
These values are set based on recommendations from https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#recommended-default-puma-process-and-thread-configuration

These new settings only take effect if the user is deploying a new application or is using the `SENSIBLE_DEFAULTS` environment variable (legacy).

The reason we cannot make this the default for all applications is some people use the lack of an environment variable to set their application values. For example their `config/puma.rb` might look like this:

```
workers ENV.fetch("WEB_CONCURRENCY") { 2 }
```

In this scenario if they've purposefully unset the `WEB_CONCURRENCY` env-var then they would expect it to use a value of `2`. This expectation is not as strong for new applications and will minimize breaking behavior.

In addition to setting this default value it also utilizes the semi-standard `.profile.d/WEB_CONCURRENCY.sh` file so that the last buildpack in the list will be the one that "wins" at setting the value for WEB_CONCURRENCY.

Here is a copy of the current recommendations and the rationale behind the selected settings:

## Recommended default Puma process and thread configuration

Here are our default recommended values for Puma processes and threads:

<table>
<thead>
  <tr style="width: 100%">
    <th style="text-align: left; white-space: nowrap;">Dyno Type</th>
    <th style="text-align: left;">Recommended web process count (WEB_CONCURRENCY)</th>
    <th style="text-align: left;">Recommended web thread count (RAILS_MAX_THREADS)</th>
    <th style="text-align: left;">Dedicated cores</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td style="text-align: left; white-space: nowrap;">free</td>
    <td style="text-align: left; ">1 (2 if enough memory)</td>
    <td style="text-align: left; ">5</td>
    <td style="text-align: left; ">n/a</td>
</tr>
  <tr>
    <td style="text-align: left; white-space: nowrap;">hobby</td>
    <td style="text-align: left; ">1 (2 if enough memory)</td>
    <td style="text-align: left; ">5</td>
    <td style="text-align: left; ">n/a</td>
  </tr>
  <tr>
    <td style="text-align: left; white-space: nowrap;">standard-1x</td>
    <td style="text-align: left; ">1 (2 if enough memory)</td>
    <td style="text-align: left; ">5</td>
    <td style="text-align: left; ">n/a</td>
  </tr>
  <tr>
    <td style="text-align: left; white-space: nowrap;">standard-2x</td>
    <td style="text-align: left; ">2</td>
    <td style="text-align: left; ">5</td>
    <td style="text-align: left; ">n/a</td>
  </tr>
  <tr>
    <td style="text-align: left; white-space: nowrap;">performance-m</td>
    <td style="text-align: left; ">2</td>
    <td style="text-align: left; ">5</td>
    <td style="text-align: left; ">1</td>
  </tr>
  <tr>
    <td style="text-align: left; white-space: nowrap;">performance-l</td>
    <td style="text-align: left; ">4</td>
    <td style="text-align: left; ">5</td>
    <td style="text-align: left; ">4</td>
  </tr>
</tbody>
</table>

>warning
>Regardless of what we recommend here, you must tune the values to your application. If your application encounters [R14 - Memory Quota Exceeded errors](https://devcenter.heroku.com/articles/ruby-memory-use), you should consider reducing your process count.

>note
>The following material is not comprehensive and is subject to change. For a full understanding of tuning your application's performance, we recommend [The Complete Guide to Rails Performance](https://www.railsspeed.com/).

## Process count value

Increasing process count increases RAM utilization, which can be a limiting factor. Another factor for setting this value is the number of physical cores on the system. Due to the [GVL](https://en.wikipedia.org/wiki/Global_interpreter_lock), the Ruby interpreter (MRI) can only run one thread executing Ruby code at a time. Due to this limitation, to fully make use of multiple cores, your application should have a process count that matches the number of physical cores on the system.

If you go above the physical core count, then the processes will contend for limited resources. Due to this contention, they will spend extra time context switching that could have been spent executing code.

You can find the number of physical cores by running `nproc` on a system via `heroku run bash`. For example:

```term
$ heroku run bash --size=performance-l
$ nproc
8
```

The value returned by `nproc` includes "hyperthreads" in addition to physical cores. All physical cores used on Heroku have a hyperthread so to gain the "true" number of physical cores divide by two. For example, with performance-l dynos, there are four physical cores and four hyperthreads. This dyno can only physically execute instructions from four processes at a time.

>note
>The value for `nproc` from free, hobby, standard-1x, and standard-2x dynos are correct, but these cores are shared between multiple applications running in containers. While `nproc` for these dynos will all return `8`, it is best to assume only one process can execute at a time.

While the number of physical cores dictates the maximum number of processes that can execute at a given time, there are cases you want to tune process count above physical core count. Multiple processes can provide redundancy in case one process crashes. When a Puma worker process crashes, it will be restarted, but this process is not instantaneous. While the master process is replacing a worker process, having redundancy can mean that the second process can still process requests. For this reason, we typically recommend a minimum of `2` processes, if possible.

The other reason to have multiple processes that exceed physical core count is if your application is not thread-safe and cannot run multiple threads. If you are only running one process in this scenario, then your core will be idle while your application makes IO calls, such as network requests or database queries. In this scenario, having an extra process would allow it to work on another request while waiting on IO.

The final consideration when setting process type is memory use. Scaling out through processes typically uses more memory than using more threads. For more information on this, see: [what is a thread](https://www.schneems.com/2017/10/23/wtf-is-a-thread/). If your application is using so much memory that it starts to swap to disk, this will dramatically reduce the performance of your application. We highly recommend tuning your process count so that your application does not encounter a [R14 - Memory Quota Exceeded error](https://devcenter.heroku.com/articles/ruby-memory-use).

## Thread count value

Once you've found an optimal value for your process count, you can further tune the system's thread count. Threads in a Ruby (MRI) process allow your app to work on multiple requests at a time when there is IO involved (database calls, network calls, etc.). It is "cheaper" for an operating system to context switch between threads, and they also generally consume less memory overall than processes since they share memory. Adding more threads to a system will [increase the overall memory use of the Ruby app over time](https://schneems.com/2019/11/07/why-does-my-apps-memory-usage-grow-asymptotically-over-time/), but it's not usually the main concern for tuning the thread number.

We recommend the same number of threads for each process type. We picked five in part because it is the default value for the Active Record connection pool, but also in part because research has shown the value to be "good enough":

> What you're seeing there, with 5-10 threads being optimal for an I/O-heavy workload, is pretty typical of CRuby. [Appfolio: Threads, Processes, and Fibers](https://engineering.appfolio.com/appfolio-engineering/2019/9/4/benchmark-results-threads-processes-and-fibers)

**Dyno load to tune thread counts (on Performance dynos)** - Heroku does provide a [Dyno Load metric](https://devcenter.heroku.com/articles/metrics#dyno-load
) to understand CPU utilization. Ideally, when tuning overall thread count (number of processes multiplied by the number of threads), you want to consume all CPU time available to your application. The load value in this metric represents the number of tasks/threads that are waiting or currently executing at a given time. If this number is high, it indicates there are too many threads spawned and that your application may be suffering from contention. If the number very infrequently rises above your physical core count, then your application may benefit from additional threads.

For more information see [What is an acceptable amount of Dyno load?](https://help.heroku.com/88G3XLA6/what-is-an-acceptable-amount-of-dyno-load)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## v207 (unreleased)
+## Master (unreleased)
+
+* Set default values for WEB_CONCURRENCY for new applications based on dyno type and size (https://github.com/heroku/heroku-buildpack-ruby/pull/946)
+
+## v207 (12/16/2019)
 
 * Vendor in libpq 5.12.1 for Heroku-18 (https://github.com/heroku/heroku-buildpack-ruby/pull/936)
 * Remove possibilities of false exceptions being raised by removing `BUNDLED WITH` from the `Gemfile.lock` (https://github.com/heroku/heroku-buildpack-ruby/pull/928)
--- a/changelogs/v208/WEB_CONCURRENCY.md
+++ b/changelogs/v208/WEB_CONCURRENCY.md
@@ -1,0 +1,34 @@
+## New Ruby applications now have a set default value WEB_CONCURRENCY based on dyno type and size
+
+The values for WEB_CONCURRENCY and RAILS_MAX_THREADS are now set for new applications based on the dyno size and type. This will allow a developer upgrading to a performance-l dyno to take advantage of additional physical CPU cores without having to modify environment variables.
+
+Existing applications that want to opt-into having these values set to defaults can do so by setting this configuration value:
+
+```
+$ heroku config:set SENSIBLE_DEFAULTS=1
+```
+
+For more information see the [Ruby Support documentation](https://devcenter.heroku.com/articles/ruby-support#default-web_concurrency)
+
+
+----
+
+## Default WEB_CONCURRENCY
+
+For new applications, the values for WEB_CONCURRENCY and RAILS_MAX_THREADS are set by default according to our [recommended Puma configuration](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#recommended-default-puma-process-and-thread-configuration).
+
+You can see if your application has this set by running `heroku run bash` and then this command:
+
+```
+$ cat .profile.d/WEB_CONCURRENCY.sh | grep RAILS_MAX_THREADS
+export RAILS_MAX_THREADS=${RAILS_MAX_THREADS:-5}
+```
+
+If you see a blank output then this feature is not enabled. To enable this feature for an existing application you can set:
+
+```
+$ heroku config:set SENSIBLE_DEFAULTS=1
+```
+
+If you still do not see the expected output, it may be due to use of multiple buildpacks and the order they are being invoked. Multiple buildpacks such as the `heroku/nodejs` buildpack also set a default value for `WEB_CONCURRENCY`. When multiple buildpacks set this value, the last one to execute will "win" and those settings will take precedence. To avoid this complication place the `heroku/ruby` buildpack after other buildpacks or manually manage your environment variable values.
+
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -321,28 +321,49 @@ EOF
   end
 
   def set_default_web_concurrency
-    <<-EOF
-case $(ulimit -u) in
-256)
-  export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-512}
-  export WEB_CONCURRENCY=${WEB_CONCURRENCY:-2}
-  ;;
-512)
-  export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-1024}
-  export WEB_CONCURRENCY=${WEB_CONCURRENCY:-4}
-  ;;
-16384)
-  export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-2560}
-  export WEB_CONCURRENCY=${WEB_CONCURRENCY:-8}
-  ;;
-32768)
-  export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-6144}
-  export WEB_CONCURRENCY=${WEB_CONCURRENCY:-16}
-  ;;
-*)
-  ;;
-esac
-EOF
+    @metadata.touch("default_web_concurrency") if new_app?
+
+    if @metadata.exists?("default_web_concurrency") || env("SENSIBLE_DEFAULTS")
+      topic("Setting default WEB_CONCURRENCY")
+
+      default_concurrency = <<~EOF
+        export RAILS_MAX_THREADS=${RAILS_MAX_THREADS:-5}
+
+        case $(ulimit -u) in
+        256)
+          # 1x dyno
+          export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-512}
+          export HEROKU_CORES=${HEROKU_CORES:-1}
+          export WEB_CONCURRENCY=${WEB_CONCURRENCY:-1}
+          ;;
+        512)
+          # 2x dyno
+          export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-1024}
+          export HEROKU_CORES=${HEROKU_CORES:-1}
+          export WEB_CONCURRENCY=${WEB_CONCURRENCY:-2}
+          ;;
+        16384)
+          # perf-m dyno
+          export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-2560}
+          export HEROKU_CORES=${HEROKU_CORES:-1}
+          export WEB_CONCURRENCY=${WEB_CONCURRENCY:-2}
+          ;;
+        32768)
+          # perf-l dyno
+          export HEROKU_RAM_LIMIT_MB=${HEROKU_RAM_LIMIT_MB:-14336}
+          export HEROKU_CORES=${HEROKU_CORES:-4}
+          export WEB_CONCURRENCY=${WEB_CONCURRENCY:-4}
+          ;;
+        *)
+          ;;
+        esac
+      EOF
+      FileUtils.mkdir_p("#{build_path}/.profile.d")
+
+      File.open("#{build_path}/.profile.d/WEB_CONCURRENCY.sh", "w") do |file|
+        file.puts default_concurrency
+      end
+    end
   end
 
   # default JRUBY_OPTS
@@ -423,7 +444,7 @@ SHELL
       set_env_override "PATH",      profiled_path.join(":")
 
       set_env_default "MALLOC_ARENA_MAX", "2"     if default_malloc_arena_max?
-      add_to_profiled set_default_web_concurrency if env("SENSIBLE_DEFAULTS")
+      set_default_web_concurrency
 
       if ruby_version.jruby?
         add_to_profiled set_jvm_max_heap
--- a/spec/hatchet/rubies_spec.rb
+++ b/spec/hatchet/rubies_spec.rb
@@ -43,7 +43,18 @@ describe "Ruby Versions on cedar-14" do
     end
   end
 
-  it "should deploy jdk 8 on cedar-14 by default" do
+  it "should deploy jruby 1.7.16.1 (jdk 7) properly on cedar-14 with sys props file" do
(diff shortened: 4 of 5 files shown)


nateberkopec:
I think Perf-L could be more aggressive. Like 8.

Taytay:
After running on Perf-Ls for a while, I was also surprised to see the recommendation match the exact number of physical cores, rather than at _least_ the number of virtual cores. (Then again, we are very I/O bound due heavy DB utilization in our Rails app, so we can get away with more concurrency.)

nateberkopec:
So re: threadcount, I don't think it's optimal to set it to anything greater than 5. My experience + Noah's benchmarking has shown that for a web process, setting it to more than 5 results in greater memory usage without a worthwhile gain in throughput because the GVL is too saturated. 

How are you going to deal with the fact that Sidekiq will use RAILS_MAX_THREADS for concurrency setting, if it's set? Sidekiq's concurrency default of 10 is fine, IMO.

schneems:
My thought process looks a little like this: If this was an entirely CPU bound task with no IO and we've got four physical cores than the optimal number of processes is 4 and threads per process is 1. Any more and we're paying for context switching that we don't need.

That's not the world we live in though. We have IO in our apps. So to account for the IO (when the GVL gets unblocked due to an IO call, which allows threads to switch) we add threads. It's a bit of a magic number, 5 seems good. I've seen the post you're talking about.

So If you're saying that more than 5 threads isn't really all that helpful (presumably because IO is already being maxed out) then I don't understand why 8 processes with 5 threads is any different than 4 processes with 10 threads. If using 40 threads is unnecessary with 4 processes, wouldn't it be unnecessary with 8 processes? 

> How are you going to deal with the fact that Sidekiq will use RAILS_MAX_THREADS for concurrency setting, if it's set? Sidekiq's concurrency default of 10 is fine, IMO.

That's a good call, thanks for the heads up. I hadn't thought about it. I made the issue and apparently didn't see when it was merged in :/ https://github.com/mperham/sidekiq/issues/2985

My first thought is that I could add `ENV["DYNO"].split(".").first == "web"` detection and special case the web. 

As for picking numbers: I have a ton of experience and feedback right-sizing puma on Heroku but for some reason people don't really open up tickets about sidekiq config. Maybe it's because the metrics tab defaults to the `web` or people are just less perf sensitive in workers?

It sounds like you are suggesting falling back to the sidekiq default, which would be fine with me.  The original intent of this PR is to set the values for Puma. On the other hand, it's a tad confusing if it's one value in one context and a different value in other contexts. 

Actually now that I think about it, it would be REALLY confusing if someone was in a `heroku run bash` session looking at their `env` and not understanding where the RAILS_MAX_THREADS was coming from or where it was set. 

I could special case for when `ENV["DYNO"]` starts with `worker` but I know from experience that lots of people have multiple dyno worker process types, so one of them would have a different value than the others, which would also be extremely confusing.

I don't know how many people realize setting their `RAILS_MAX_THREADS` affects their sidekiq concurrency. Probably not a lot. 

I'm honestly not sure the best way to move forward. I think these are the options:

## Defaults for Sidekiq RAILS_MAX_THREADS on Heroku options

### 1.a) Do nothing

Let sidekiq pick up whatever we set for RAILS_MAX_THREADS for puma

Consequences:
- Sidekiq will be running at half of their default capacity (since our default is 5 and their default is 10)
- All RAILS_MAX_THREADS values across all dynos will be consistent


### 1.b) Do nothing but at least match RAILS_MAX_THREADS=10 for perf-l dynos

If we kept 4 processes and but bumped to 10 threads for perf-l dynos then we'll be matching the sidekiq defaults for the case where we've got four physical cores. 

Consequences:

- Sidekiq on perf-l dynos will be running at 10 which matches their default
- All RAILS_MAX_THREADS values across all dynos will be consistent
- Sidekiq on non perf-l dynos will not be running with their default. In those situations though, there's only one physical core. 
- Actually does running sidekiq on multiple cores even help if you're not using extra processes?


### 2.a) Special case DYNO starts with "web"

For the dynos like `web.1` set the RAILS_MAX_THREADS to something different. We have multiple options here, the easiest would be to only set a value when the dyno type is "web"

Consequences:
- Would allow sidekiq to set a default
- The default of sidekiq (10) is higher than the default of Rails (5) so there would likely be connection timeout errors. I'm honestly surprised this isn't more of an issue.
- Confusing for people using `heroku run` since the RAILS_MAX_THREADS wouldn't show up since we are special casing based on DYNO env var. 


### Some other option?

Any ideas? cc/ @mperham

nateberkopec:
I think you're assuming that GVL saturation == saturation of 1 CPU hyperthread. Do we know that's true? I don't think it is. IO, memory or disk waiting can cause a Ruby process to wait while it has the GVL. Not all IO operations necessarily release the GVL and memory or disk waiting certainly won't.

schneems:
> 1 CPU hyperthread

FWIW I tend to not care about hyperthreads and just assume it's hardware optimization around pre-fetching CPU data. The physical CPU is the thing I mostly care about.

> Not all IO operations necessarily release the GVL and memory or disk waiting certainly won't.

Interesting. I see what you're saying. You're referring to loading code from disk for example, rather than explicitly reading a file right? Or fetching data from ram. I wish we had some better ways to test this stuff. You're right we're not measuring this stuff, and I have no way of definitively saying how much it impacts an app. 

From my HCI class I saw first hand saturating a CPU with too many threads has a significant performance penalty, but TBH I don't know what that looks like in Ruby. Not sure how we would go about designing experiments and what we should be measuring. This might be a conversation for somewhere else.

## Next steps

I think the way I want to move forward is to remove the RAILS_MAX_THREADS setting so it uses whatever defaults are on the system and stick with my original recommendations for processes. They're conservative, but these are the defaults. People can always re-tune manually.

I've seen a few tickets come in recently where people are on perf-l dynos and are using 8-9 processes with 5 threads. They're seeing spikes up to load max of 26 and load averages of 6-7+. 

I also want to make a way to make these defaults "stickey" that way we can change them in the future without worrying about breaking people's existing apps. While I care a lot about performance, I care even more about stability.

nateberkopec:
> They're seeing spikes up to load max of 26 and load averages of 6-7+.

Nice. Well if you have data that disagrees, then you should go with that. Obviously we should be keeping load averages probably at 50-75% of the logical core count.

ArturT:
Hi @nateberkopec 

> Obviously we should be keeping load averages probably at 50-75% of the logical core count.

For PL dyno does it mean 50-75% of 8 vCPU so we should aim for 4-6 load averages?


On PL dyno you can run 8 puma workers and 5 threads (8*5 = 40) as recommended in this topic. What about not thread-safe rails apps? Would it make sense also for them to have 40 processes (40 puma processes and each with 1 puma thread) assuming it fits 14GB RAM size?

There is some CPU overhead of context switching between 40 processes (higher than when using threads with fewer puma workers I guess). Does it make sense to have 40 puma workers or we should aim for less because they are not as fast as threads?


>  They're seeing spikes up to load max of 26 and load averages of 6-7+.

Is it fine that we can see spikes like 26? Is it acceptable because we aim for averages to be within 50-75% of the logical core count and we should not worry that much about spikes time?

Is it better to utilize CPU as much as possible for most of the time at expense of performance degradation during traffic spike time? 

I guess having more puma workers makes sense when you have slow requests (doing some IO) because then less % of traffic would be directed to the slow puma worker. 
If you would have a low pool of puma workers then a higher % of traffic would be directed to the busy puma worker and more requests would be delayed by the time of the slow request.

Do you have any recommendations about max dyno load or how much should we worry about it?

Thanks

nateberkopec:
> For PL dyno does it mean 50-75% of 8 vCPU so we should aim for 4-6 load averages?

Yes

> What about not thread-safe rails apps? Would it make sense also for them to have 40 processes (40 puma processes and each with 1 puma thread) assuming it fits 14GB RAM size?

Generally, yes, as many processes as you can fit before running out of memory. However, you should also remove processes if you exceed a load average of ~6 regularly.

nateberkopec:
> There is some CPU overhead of context switching between 40 processes (higher than when using threads with fewer puma workers I guess).

I don't believe context switching is more expensive if 2 threads come from different processes. In any case, context switching costs nowadays are extremely, extremely low in the context of a ~200 millisecond web request.

> Is it fine that we can see spikes like 26?

Worry if it affects your response times in a way that you cannot accept.

PR #950, opened 2020-02-03
Fix RubyVersion#vendored_bundler?

jacobthemyth:
Ruby 2.6 started vendoring bundler, so this returns the correct value from `RubyVersion#vendored_bundler?`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,76 @@
+version: 2
+
+references:
+  default_docker_ruby_executor: &default_docker_ruby_executor
+    image: circleci/ruby:2.5.7-stretch
+    environment:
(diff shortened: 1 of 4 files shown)


schneems:
Is there an associated bug or issue that you're trying to work around with this fix?

The only place in the code we're using that is here https://github.com/heroku/heroku-buildpack-ruby/blob/cdae0636a79404d557097723ee99e9bfce15ae66/lib/language_pack/ruby.rb#L647.

I don't think the shim is ever needed. If theres' a case where the current buildpack fails, then let me know how I can reproduce the failure.

jacobthemyth:
@schneems Sorry, this was intended to be opened against our fork, but since this was listed as a remote, it was unintentionally opened against this repo, closing.

PR #953, opened 2020-02-26
Ruby version installation errors and warnings

schneems:
Currently when a customer tries to upgrade their stack and are using an old version of Ruby they get a very unhelpful error:

```
       Command: 'set -o pipefail; curl -L --fail --retry 5 --retry-delay 1 --connect-timeout 3 --max-time 30 https://s3-external-1.amazonaws.com/heroku-buildpack-ruby/heroku-18/ruby-1.9.3.tgz -s -o - | tar zxf - ' failed on attempt 1 of 3.
       Command: 'set -o pipefail; curl -L --fail --retry 5 --retry-delay 1 --connect-timeout 3 --max-time 30 https://s3-external-1.amazonaws.com/heroku-buildpack-ruby/heroku-18/ruby-1.9.3.tgz -s -o - | tar zxf - ' failed on attempt 2 of 3.
 !
 !     An error occurred while installing ruby-1.9.3
 !
 !     Heroku recommends you use the latest supported Ruby version listed here:
 !     https://devcenter.heroku.com/articles/ruby-support#supported-runtimes
 !
 !     For more information on syntax for declaring a Ruby version see:
 !     https://devcenter.heroku.com/articles/ruby-versions
 !
 !
 !     Debug InformationCommand: 'set -o pipefail; curl -L --fail --retry 5 --retry-delay 1 --connect-timeout 3 --max-time 30 https://s3-external-1.amazonaws.com/heroku-buildpack-ruby/heroku-18/ruby-1.9.3.tgz -s -o - | tar zxf - ' failed unexpectedly:
 !
 !     gzip: stdin: unexpected end of file
 !     tar: Child returned status 1
 !     tar: Error is not recoverable: exiting now
 !
 !     Push rejected, failed to compile Ruby app.
 !     Push failed
 ```

 This is awful and gives no context. The goal of this PR is to do a few things:

 - If a Ruby version does not exist on any stack, explicitly state the problem
 - If a Ruby version exists, but not on your current stack, list the stacks it is present on
 - If a Ruby version exists, but not for the next stack (i.e. heroku-16 if you're deploying to cedar-14) then the customer should get a warning

### CODE REVIEW

File: lib/language_pack/ruby.rb [hunk 1]
```diff
@@ -512,10 +534,15 @@ def install_ruby
     instrument 'ruby.install_ruby' do
       return false unless ruby_version
       installer = LanguagePack::Installers::RubyInstaller.installer(ruby_version).new(@stack)
+      file = "#{ruby_version.version_for_download}.tgz"
 
       if ruby_version.build?
         installer.fetch_unpack(ruby_version, build_ruby_path, true)
+        file.sub!("ruby", "ruby-build")
       end
+      @ruby_download_check = LanguagePack::Helpers::DownloadPresence.new(file)
```
edmorley: I'm not quite sure what the "build ruby" is (vs the normal ruby), but it seems like if `installer.fetch_unpack()` above were to fail (because the version doesn't exist), then `fetch_untar` (called by `fetch_unpack`) would raise `FetchError` and thus this line would never get called prior to the usage of `@ruby_download_check` in the `rescue`?

schneems: Great catch. I fixed that. I also added a hatchet integration test for the scenario where there's a bad ruby version.

schneems:
Thanks for the review

PR #968, opened 2020-03-20
[changelog skip] export STACK so it can be used in scripts

hone:
To support https://github.com/heroku/pack-images/pull/78


schneems:
Need to add `[ci skip]` to the title and re-push code to re-trigger the changelog test.

PR #888, opened 2019-05-30
Cloud Native Buildpack Support

schneems:

PR #973, opened 2020-03-31
Fix absolute path bug in .profile.d/ruby.sh $PATH

schneems:
In #888 a relative path was accidentally introduced into the `.profile.d/ruby.sh` script:

```
export PATH="bin:$HOME/vendor/bundle/bin:$HOME/vendor/bundle/ruby/2.6.0/bin:$PATH"
```

That first path `bin` is relative and should be `$HOME/bin`.

Without an absolute path, changing into a different directory means that the system version of ruby will be used

```
~ $ which ruby
bin/ruby
~ $ mkdir foo; cd foo; which ruby
/usr/bin/ruby
```

> Note different results from 

This is a regression  was not caught by tests previously because the version of ruby that the test `cd_ruby` uses is 2.5.1 and that happens to match system ruby. To fix for catching future regressions the test is now more explicit.

Unfortunately there's a bug in hatchet preventing that specific test https://github.com/heroku/hatchet/issues/71 so instead I'm at least asserting that it's an absolute path.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -429,13 +429,14 @@ SHELL
   # sets up the profile.d script for this buildpack
   def setup_profiled(ruby_layer_path = "$HOME", gem_layer_path = "$HOME")
     instrument 'setup_profiled' do
-      profiled_path = binstubs_relative_paths(gem_layer_path)
+      profiled_path = ["$HOME/bin"]
+      profiled_path << binstubs_relative_paths(gem_layer_path)
       profiled_path << "vendor/#{@yarn_installer.binary_path}" if has_yarn_binary?
       profiled_path << "$PATH"
 
       set_env_default  "LANG",     "en_US.UTF-8"
       set_env_override "GEM_PATH", "#{gem_layer_path}/#{slug_vendor_base}:$GEM_PATH"
-      set_env_override "PATH",      profiled_path.join(":")
+      set_env_override "PATH",      profiled_path.uniq.join(":")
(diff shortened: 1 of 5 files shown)


schneems:
Blocked by https://github.com/heroku/heroku-buildpack-activestorage-preview/issues/15

lib/language_pack/ruby.rb
```diff
@@ -211,7 +211,7 @@ def default_path(gem_layer_path = ".")
     # need to remove bin/ folder since it links
     # to the wrong --prefix ruby binstubs
     # breaking require. This only applies to Ruby 1.9.2 and 1.8.7.
-    safe_binstubs = binstubs_relative_paths(gem_layer_path) - ["bin"]
+    safe_binstubs = binstubs_relative_paths(gem_layer_path) - ["bin", "#{gem_layer_path}/bin"]
```
hone, line 214: we don't need `bin` here anymore right?

lib/language_pack/ruby.rb
```diff
@@ -429,13 +429,14 @@ def setup_export(layer = nil)
   # sets up the profile.d script for this buildpack
   def setup_profiled(ruby_layer_path = "$HOME", gem_layer_path = "$HOME")
     instrument 'setup_profiled' do
-      profiled_path = binstubs_relative_paths(gem_layer_path)
+      profiled_path = ["$HOME/bin"]
```
hone, line 432: is this redundant if `$HOME` is passed in?

PR #974, opened 2020-03-31
Ruby 2.6.6 is the new default

schneems:


schneems:
Depends on https://github.com/heroku/heroku-buildpack-ruby/pull/973

schneems:
>  is this process documented anywhere?

Not really. If you change one place then tests start failing so at least that's a guiding light.

PR #978, opened 2020-04-08
Fix CI Test runs not using bundler cache

schneems:
When testing the same code in the same pipeline the cache should be used so gems do not have to be re-installed. This same test passes using an older version of the buildpack:

```
$ HATCHET_BUILDPACK_BRANCH="v211" rspec spec/hatchet/ci_spec.rb:39
```


schneems:
Internal tickets ids:
- 848134
- 846451
- 847045

PR #982, opened 2020-04-18
Gracefully handle unrecognised stacks

edmorley:
Previously if the current stack was not in the download presence check's list of known stacks, the compile would fail with:

```
lib/language_pack/helpers/download_presence.rb:32:in `next_stack': undefined method `+' for nil:NilClass (NoMethodError)
	from /app/tmp/buildpacks/.../lib/language_pack/ruby.rb:469:in `warn_stack_upgrade'
	from /app/tmp/buildpacks/.../lib/language_pack/ruby.rb:462:in `warn_outdated_ruby'
	from /app/tmp/buildpacks/.../lib/language_pack/ruby.rb:115:in `block in compile'
```

Refs [W-7457565](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007QfR8IAK/view).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master (unreleased)
 
+* Gracefully handle unrecognised stacks ([#982](https://github.com/heroku/heroku-buildpack-ruby/pull/982))
+
 ## v215 (4/9/2020)
 
 * Fix bundler cache not being used in CI builds (https://github.com/heroku/heroku-buildpack-ruby/pull/978)
--- a/lib/language_pack/helpers/download_presence.rb
+++ b/lib/language_pack/helpers/download_presence.rb
@@ -28,13 +28,20 @@ class LanguagePack::Helpers::DownloadPresence
     end
   end
 
-  def next_stack(current_stack: )
-    next_index = @stacks.index(current_stack) + 1
+  def supported_stack?(current_stack: )
+    @stacks.include?(current_stack)
+  end
 
+  def next_stack(current_stack: )
+    return unless supported_stack?(current_stack: current_stack)
+
+    next_index = @stacks.index(current_stack) + 1
     @stacks[next_index]
   end
 
   def exists_on_next_stack?(current_stack: )
+    return false unless supported_stack?(current_stack: current_stack)
+
     next_index = @stacks.index(current_stack) + 1
     @threads[next_index]
   end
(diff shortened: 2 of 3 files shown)


edmorley:
@schneems I wonder as well if `DownloadPresence` should be refactored to take the current_stack at initialisation, with the methods like `next_stack` and `exists_on_next_stack?` then not needing any parameters?

schneems:
This looks good. I have a semantic change, I would like to get rid of the guard check method though and wrap the logic to be all-in-one like this: 

```ruby
  def next_stack(current_stack: )
    if (current_index = @stacks.index(current_stack))
      return @threads[current_index + 1]
    else
      return nil
    end
  end
  alias :exists_on_next_stack?, :next_stack
```

Then we can get this merged in.

> I wonder as well if DownloadPresence should be refactored to take the current_stack at initialization, with the methods like next_stack and exists_on_next_stack? then not needing any parameters? (Though this is perhaps best in another PR)

I don't entirely remember why, I'm guessing that I did this as an optimization to speed up tests since initializing the object and then querying it causes network requests.

edmorley:
Thank you for the review. I like the idea of reducing the duplication, though in the new version I think the meaning of `next_stack` has now changed? Before it was "the name of the stack that's one newer than the current stack", whereas now it's "the name of the stack that's one newer than the current stack iff there is a Ruby binary for that stack".

That difference is currently not a problem in practice, because of this return here:
https://github.com/heroku/heroku-buildpack-ruby/blob/06427671598fd1eba53b11789ac663d0508d5464/lib/language_pack/ruby.rb#L470

...but if `next_stack` ever gets used elsewhere that might not be the case?

schneems:
> the name of the stack that's one newer than the current stack iff there is a Ruby binary for that stack".

I intended it to return `nil` when it's at the end. In Ruby you can index past the end of an array and it will return nil:

```ruby
array = [:a, :b, :c]
puts array[2] # => c
puts array[3] # => nil
```

I was relying on this behavior already (to return nil when there is no "next" stack).

edmorley:
That's not what I meant as the difference. Before `next_stack` took the result from `stacks`, whereas now it uses `@threads`. The two aren't equivalent - since `stacks` is "all of the known stacks regardless of what Ruby versions they have", whereas `@threads` is "all of the known stacks which have that Ruby version".

This difference might not matter in practice, but I wanted to call it out, given that the name of `next_stack` to me means it's a general helper asking if there is a stack newer than the current one, and if so, what is it's name.

schneems:
Ahh, gotcha. I completely glossed over that variable name difference. My code isn't equivalent at all.

PR #986, opened 2020-04-23
Update Node and Yarn default versions

schneems:


changelogs/v216/node_yarn.md
```diff
@@ -0,0 +1,8 @@
+## Ruby apps now default to Node version 12.16.1 and Yarn version 1.22.0
+
+Applications using the `heroku/ruby` buildpack that do not have a version of Node installed by another buildpack (such as the `heroku/nodejs` buildpack) will now receive:
+
+- Node version 12.16.1
+- Yarn version 1.22.0
```
edmorley, line 6: How do you feel about also listing the previous versions, so that customers know what version to downgrade to, should they encounter issues? (Otherwise I guess they would need to find this PR, or inspect old build logs to see what versions were being mentioned in build output?) :-)

schneems: I like it. I'll add that. Great feedback, this type of comment is exactly why I started putting changlog messages in the PR.

edmorley: Yeah great idea to include them in the repo :-)

This conversation also reminded me of the great changelog entries that eslint uses, where it has an explicit "here's how to get back the old behaviour" code snippet for breaking changes. eg:
https://eslint.org/docs/user-guide/migrating-to-6.0.0#eslint-recommended-changes

lib/language_pack/helpers/nodebin.rb
```diff
@@ -2,15 +2,15 @@
...
+    version = "12.16.1"
     {
       "number" => version,
       "url"    => "https://s3.amazonaws.com/heroku-nodebin/node/release/linux-x64/node-v#{version}-linux-x64.tar.gz"
     }
   end
 
   def self.hardcoded_yarn
-    version = "1.16.0"
+    version = "1.22.0"
```
danielleadams, line 13: can you use 1.22.x here? If not, the latest is currently 1.22.4

lib/language_pack/helpers/nodebin.rb
```diff
@@ -2,15 +2,15 @@
 
 class LanguagePack::Helpers::Nodebin
   def self.hardcoded_node_lts
-    version = "10.15.3"
+    version = "12.16.1"
```
danielleadams, line 5: same question I have here was with yarn in regards to doing an latest patch

danielleadams: latest for Node is 12.16.2

PR #992, opened 2020-05-08
[changelog skip] Fix check changelog action

schneems:


schneems:
> Does this fix the thing where the changelog check still fails if I edit the PR title afterwards to add changelog skip? If so, this sounds great :-)

Yes, that's exactly what this does. I'll send over a PR to all the language teams.

PR #997, opened 2020-05-18
[changelog skip] Amend run-on sentence

lolaodelola:
This is a very minor change [requested of a customer](https://heroku.support/865586). I've amended the run on sentence to be two sentences instead so it reads a little easier.

The changelog check fails but I'm unsure if this change requires a new changelog item.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -519,7 +519,7 @@ SHELL
 
       #{@outdated_version_check.suggested_ruby_minor_version}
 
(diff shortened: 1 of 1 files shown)


schneems:
In order to get changeling to allow us to merge this, CI has to run. But CI only runs from branches from within the buildpack (so that other people can't maliciously steal our secret env vars). I moved your code into an internal branch rather than from your `lolaodelola/heroku-buildpack-ruby` fork here https://github.com/heroku/heroku-buildpack-ruby/pull/999

PR #1011, opened 2020-06-03
[close #1001][changelog skip] Put Yarn first on the path

### LINKED ISSUE #1001: Regression on master for deploying ember-cli-rails app
Hi @schneems I pinged you on twitter because of this. I think I have identified the offending commit and unfortunately it is 8413eec. I have yet to fully understand the issue but I am willing to help with debugging if possible. I guess a useful thing would be to have a minimal repro app that shows the same behaviour?

Here are some data points:

- Deploying using the currently released version of the buildpack (by just specifying `heroku/ruby`) works
- Deploying using master (using `https://github.com/heroku/heroku-buildpack-ruby`) fails (more on the failure later)
- Deploying using `https://github.com/heroku/heroku-buildpack-ruby#65c39f7e0d41bb9a44a9b74c0aafffc66250e487`works
- Deploying using `https://github.com/heroku/heroku-buildpack-ruby#8413eecb42ce8a99b47194ab9d237093175b915f` fails

The failure happens when ember-cli-rails tries to compile the ember app. It runs yarn install in the folder where the ember app is stored (`frontend` by default) where a package.json and a yarn.lock should be present. When a "broken" buildpack is at work, it looks as if yarn is instead either run in the root folder, where (historically) an empty package.json and yarn.lock is present.

I don't have enough inside knowledge about how this buildpack works to figure out why the corresponding commit would break that behaviour. The code in ember-cli-rails that installs the dependencies is [here](https://github.com/thoughtbot/ember-cli-rails/blob/master/lib/ember_cli/shell.rb#L35).

If it helps, I'll try to set up a minimal testing app. Unfortunately the app that breaks is closed source.

schneems:
The binstub that Rails generates has a bug in it that prevents calling `yarn install` from any directory other than the app root. We were accidentally handling this when we install yarn (without the nodejs buildpack) since we prepend the yarn path https://github.com/heroku/heroku-buildpack-ruby/blob/af368418f2aef9f455afd7f0b77b3137aa8d38a5/lib/language_pack/ruby.rb#L696. However if the app is using the `heroku/nodejs` then the `bin/yarn` binstub will appear first in the buildpack after the changes in https://github.com/heroku/heroku-buildpack-ruby/commit/8413eecb42ce8a99b47194ab9d237093175b915f.

To fix this, we are manually putting yarn first on the path at runtime and build time.

While this is a fix to a regression on master, it has never been released. This fix preserves existing behavior with the deployed buildpack. (In which the `bin/yarn` binstub is not being used).

### DIFF
--- a/hatchet.json
+++ b/hatchet.json
@@ -86,11 +86,8 @@
   "heroku": [
     "heroku/ruby-getting-started"
   ],
-  "multibuildpack": [
-    "sharpstone/node_multi"
-  ],
   "node": [
-    "sharpstone/webpacker_no_execjs"
+    "sharpstone/minimal_webpacker"
   ],
   "ci": [
     "sharpstone/rails5_ruby_schema_format",
--- a/hatchet.lock
+++ b/hatchet.lock
@@ -33,10 +33,8 @@
   - 9ddca7b694875499165eb56adffd3e29b38405c5
 - - "./repos/jruby/jruby_naether"
   - 4a11f8af5a3cca21f3659394e5bbac3cca9b6a2c
-- - "./repos/multibuildpack/node_multi"
-  - c8f822a0cdd3b7ed92f19bd1112bd9e301bc4a5e
-- - "./repos/node/webpacker_no_execjs"
-  - 26792540c924699c69e4c79019b65bc9192586ea
+- - "./repos/node/minimal_webpacker"
+  - bca8b8169c553a0f62f0263b621cf3a5813e4023
 - - "./repos/rack/default_ruby"
   - da748a59340be8b950e7bbbfb32077eb67d70c3c
 - - "./repos/rack/mri_187_nokogiri"
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -333,10 +333,10 @@ EOF
     instrument 'ruby.setup_language_pack_environment' do
       if ruby_version.jruby?
         ENV["PATH"] += ":bin"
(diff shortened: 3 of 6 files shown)


schneems:
the `yarn:install` task hard codes the path to yarn because of course it does 

```
$ cat -n /Users/rschneeman/.gem/ruby/2.7.1/gems/railties-6.0.3.1/lib/rails/tasks/yarn.rake
     1	# frozen_string_literal: true
     2
     3	namespace :yarn do
     4	  desc "Install all JavaScript dependencies as specified via Yarn"
     5	  task :install do
     6	    # Install only production deps when for not usual envs.
     7	    valid_node_envs = %w[test development production]
     8	    node_env = ENV.fetch("NODE_ENV") do
     9	      valid_node_envs.include?(Rails.env) ? Rails.env : "production"
    10	    end
    11	    system({ "NODE_ENV" => node_env }, "#{Rails.root}/bin/yarn install --no-progress --frozen-lockfile")
    12	  end
    13	end
    14
    15	# Run Yarn prior to Sprockets assets precompilation, so dependencies are available for use.
    16	if Rake::Task.task_defined?("assets:precompile")
    17	  Rake::Task["assets:precompile"].enhance [ "yarn:install" ]
    18	end
```

I need a new test.

schneems:
Got the new test that doesn't rely on webpacker or rails. Looks like it works well enough. In the process I found that we're duplicating some tests and logic. I was able to delete some tests but with each test doing more for a net of fewer tests in the buildpack.

Issue #1005, opened 2020-05-29
Error running rake or release tasks `ruby: bad interpreter: No such file or directory`


joakimk:
It works on https://github.com/heroku/heroku-buildpack-ruby.git#v215. The app is soon to be retired, so locking down the deploy stack is fine for us, but this could be an issue for others.

On master we get an error like:
```
remote: -----> Detecting rake tasks        
 
remote:         
 
remote:  !        
 
remote:  !     Could not detect rake tasks        
 
remote:  !     ensure you can run `$ bundle exec rake -P` against your app        
 
remote:  !     and using the production group of your Gemfile.        
 
remote:  !     bash: /tmp/build_FOO/bin/rake: /app/vendor/ruby-2.3.8/bin/ruby: bad interpreter: No such file or directory        
 
remote:  !        
```

schneems:
Can you please open a support ticket? Go to help.heroku.com search for "buildpack bug" and open a ticket giving access to your app.

schneems:
Did you manage to open a ticket so I can debug this?

-- 
Richard Schneeman
https://www.schneems.com
he/him

mmrobins:
@schneems I'm seeing the same issue.  I opened a ticket with an app attached that reproduces the issue https://help.heroku.com/tickets/882725

schneems:
It looks like my change doesn't play well with certain versions of Ruby 2.6. I recommend people upgrade to 2.6.6 if they're getting this failure mode:

```
 !     /tmp/build_<dir>/vendor/bundle/ruby/2.6.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:319:in `check_for_activated_spec!': You have already activated bundler 1.17.3, but your Gemfile requires bundler 1.17.2. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```
@joakimk it looks like your issue is different since you're using an older version of Ruby. Are you able to open a ticket? I would like to investigate your failure mode.

schneems:
@joakimk I've merged a fix to my last known regression into master. I'm looking to add some warnings and get a deploy out hopefully in the next week or two. Without more context to reproduce your bug the `heroku/ruby` release might contain the same regression you're seeing. If you have any time this week, can you open a ticket with some more information? Let me know if you have any other questions.

wwahammy:
I just wanted to jump in here. I'm also having the exact same error as  @joakimk when trying to deploy to Heroku. using the default buildpack release.

[...]

PR #1012, opened 2020-06-03
Switch to using /usr/bin/env bash

mmrobins:
The main motiviation for switching the shebang is so that I can use
https://github.com/edmorley/heroku-buildpack-timestamps

But it's generally a good idea to have shebang lines that make scripts
portable, even if it is only going to run on heroku

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # The actual compilation code lives in `bin/support/ruby_compile`. This file instead
 # bootstraps the ruby needed and then executes `bin/support/ruby_compile`
 
--- a/bin/test
+++ b/bin/test
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # The actual `bin/test-compile` code lives in `bin/ruby_test-compile`. This file instead
 # bootstraps the ruby needed and then executes `bin/ruby_test-compile`
 
--- a/bin/test-compile
+++ b/bin/test-compile
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # The actual `bin/test-compile` code lives in `bin/ruby_test-compile`. This file instead
 # bootstraps the ruby needed and then executes `bin/ruby_test-compile`
 
--- a/support/s3/hmac
+++ b/support/s3/hmac
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Implement HMAC functionality on top of the OpenSSL digest functions.
 # licensed under the terms of the GNU GPL v2
 # Copyright 2007 Victor Lowther <victor.lowther@gmail.com>
--- a/support/s3/s3
+++ b/support/s3/s3
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # basic amazon s3 operations
 # Licensed under the terms of the GNU GPL v2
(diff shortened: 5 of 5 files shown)


schneems:
Thanks, i've queued up tests to run on https://github.com/heroku/heroku-buildpack-ruby/pull/1013. Have you tried using the timing buildpack with your branch? Is that the only change needed?

mmrobins:
Wow, thanks for the quick response, I mainly made the PR to get the CI running cause I was too lazy to do it on my machine, figured I'd have more time to play locally before this got 👀 

Looks like this mostly works with the timing buildpack, I'm hitting an issue running rake tasks still that seems related, but I haven't dug into why yet

```
-----> Ruby app detected
2020-06-03 20:01:10 -----> Installing bundler 1.17.3
2020-06-03 20:01:10 -----> Removing BUNDLED WITH version in the Gemfile.lock
2020-06-03 20:01:11 -----> Compiling Ruby/Rails
2020-06-03 20:01:12 -----> Using Ruby version: ruby-2.6.1
2020-06-03 20:01:12 -----> Installing dependencies using bundler 1.17.3
2020-06-03 20:05:26 -----> Installing yarn-v1.22.4
2020-06-03 20:05:27 -----> Detecting rake tasks
2020-06-03 20:05:27 
/app/tmp/buildpacks/c29ed44d94f67cb235a686360fc04cceb0a1a453fc318cb4faa7c3ea4e326ee73b7a33283de3cb6373390c1364380073177649a841c967f180335363ffb7bc8d/lib/language_pack/helpers/rake_runner.rb:106:in `load_rake_tasks!': Could not detect rake tasks (LanguagePack::Helpers::RakeRunner::CannotLoadRakefileError)
ensure you can run `$ bundle exec rake -P` against your app
and using the production group of your Gemfile.
/tmp/build_31a8cba504f2bf9365bd3cf4ef83036f/vendor/bundle/ruby/2.6.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:319:in `check_for_activated_spec!': You have already activated bundler 1.17.3, but your Gemfile requires bundler 1.17.2. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```

mmrobins:
Looks like that issue running rake might just be because I re-ordered the buildpacks to put Ruby first so I could test this change, and I need the nodejs buildpack before ruby.  I'm re-running with buildpacks in the normal order to verify, but this is for an app that takes 35+ minutes to build currently, so I'll report back when it's finally done 🐌

mmrobins:
Actually it appears that problem with `Detecting rake tasks` is something I hit even in the master branch for `heroku-buildpack-ruby`, whether I use https://github.com/edmorley/heroku-buildpack-timestamps or not, and no matter what order I put the nodejs buildpack in.  I don't have that problem with `heroku/ruby`, which I assume is pretty much the same under the hood as `https://github.com/heroku/heroku-buildpack-ruby#v215` would be, and if I fork off the v215 release https://github.com/mmrobins/heroku-buildpack-ruby/tree/bash_bin_v215 I don't hit that issue and get all the way through the build

So yeah, something that changed between v215 and master is giving me issues, but the change to `/usr/bin/env bash` seems to work just fine

mmrobins:
Looks like the same thing as https://github.com/heroku/heroku-buildpack-ruby/issues/1005

mmrobins:
@schneems updated my ruby app from 2.6.1 -> 2.6.6 and everything works fine with this PR for me.  Not sure if I should do anything with this PR or if https://github.com/heroku/heroku-buildpack-ruby/pull/1013 is how to proceed

PR #1013, opened 2020-06-03
[changelog skip] Switch to using /usr/bin/env bash

schneems:
The main motiviation for switching the shebang is so that I can use
https://github.com/edmorley/heroku-buildpack-timestamps

But it's generally a good idea to have shebang lines that make scripts
portable, even if it is only going to run on heroku

Running tests for https://github.com/heroku/heroku-buildpack-ruby/pull/1012

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # The actual compilation code lives in `bin/support/ruby_compile`. This file instead
 # bootstraps the ruby needed and then executes `bin/support/ruby_compile`
 
--- a/bin/test
+++ b/bin/test
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # The actual `bin/test-compile` code lives in `bin/ruby_test-compile`. This file instead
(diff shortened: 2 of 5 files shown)


edmorley:
@schneems I'm not sure if this is something you do already, but for PRs that are cherrypicked to run Hachett, we can use the GitHub co-authored feature to ensure the resultant commit to master still credits the author:
https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-github

edmorley:
(Although saying that, perhaps it's only truly necessary for squash and merge, where the original commit is lost)

schneems:
> PRs that are cherrypicked to run Hachett

I don't cherry pick at all. You'll notice the commit gives me no credit. I run this:

```
git checkout -b mmrobins-bash_bin master
git pull https://github.com/mmrobins/heroku-buildpack-ruby.git bash_bin
git push
```

It is just their commit but on a branch from inside of the org so tests run. I do wish CI and friends had a "manually run this with protected env vars" option so we wouldn't have to do this dance. If only we knew a company that ran a CI service 🤔

PR #623, opened 2017-09-15
Revert "Merge pull request #586 from heroku/check-binstubs"

schneems:
This reverts commit c2ba079766dde7920b05738fc3e108d9b28116c3, reversing
changes made to c42a5781a2f6ca2efba7dad4fe7d49d8a800e336.

PR #1014, opened 2020-06-09
[close #990] Warn on bad shebang line

### LINKED ISSUE #990: Bad binstub shebang lines
There is a class of errors caused by when a binstub has a bad shebang line. For instance:

```
#!/usr/bin/env ruby2.5
```

Instead of 

```
#!/usr/bin/env ruby
```

This was previously attempted to be mitigated in:

- Introduced: https://github.com/heroku/heroku-buildpack-ruby/pull/586
- Rolled back: https://github.com/heroku/heroku-buildpack-ruby/pull/623

This came up recently in:
https://heroku.support/858951

However overall support volume for this class of errors is relatively low.

We could revisit adding a warning. We could also possibly alias the most common ruby versions such as `ruby2.5 => ruby`  etc. but this might have undesired side effects. It's unknown if anyone is attempting to "dual boot" multiple versions of Rubies. I've not seen anything like that yet, but it doesn't mean it doesn't exist.

schneems:
This article explains binstubs and "shebang" lines: https://devcenter.heroku.com/articles/bad-ruby-binstub-shebang. A relatively common problem when generating a binstub is for it to contain a bad shebang line. Here's an example:

```
#!/usr/bin/env ruby2.5
```

If you've got a binstub with that shebang line in your project and are running on Heroku 18 then your app will be forced to use `ruby2.5` binary no matter what version of Ruby you've specified in the `Gemfile`. This causes very odd behavior and weird, difficult to debug bundler errors. 

This PR does not fix the problem, but will at least warn anyone with a version number in their ruby shebang line.

This behavior was originally:

-    Introduced: #586
-    Rolled back: #623
 

It was rolled back because it would falsely claim that `#!/usr/bin/env bash` was incorrect also it only protected against a limited set of binstubs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * A bug in 2.6.0, 2.6.1, 2.6.3 require a Ruby upgrade, a warning has been added (https://github.com/heroku/heroku-buildpack-ruby/pull/1015)
 * The spring library is now disabled by setting the enviornment variable DISABLE_SPRING=1 (https://github.com/heroku/heroku-buildpack-ruby/pull/1017)
+* Warn when a bad "shebang" line in a binstub is detected (https://github.com/heroku/heroku-buildpack-ruby/pull/1014)
 * Default node version now 12.16.2, yarn is 1.22.4 (https://github.com/heroku/heroku-buildpack-ruby/pull/986)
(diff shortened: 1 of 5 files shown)


schneems:
I had to revert this in https://github.com/heroku/heroku-buildpack-ruby/pull/1020

PR #1015, opened 2020-06-09
Warn on 2.6 old versions due to bundler binstub bug

schneems:
When the PATH was refactored to put the binstubs directory `bin/` earlier on the path it introduced a regression with 2.6.0, 2.6.1, and 2.6.2 apps. This is due to an integration with bundler and is documented in our "known issues" https://devcenter.heroku.com/articles/bundler-version#known-upgrade-issues. Since it's confusing that this behavior previously worked but now no-longer works, I am adding a warning to ease the migration process. 

Related internal support ticket https://heroku.support/882725


schneems:
Another ticket:

- https://heroku.support/899351

PR #1018, opened 2020-06-11
[close #934] Skip rake task if it does not exist

### LINKED ISSUE #934: Task `assets:clean` is executed without checking whether it exists
## Context

Using Rails 6 without Sprocket, with Webpacker only.

## What happen

This exception happen after assets compilation : 

>        Done in 5.16s.
>        Asset precompilation completed (64.17s)
>        Cleaning assets
>        Running: rake assets:clean
>        I, [2019-11-25T17:07:47.298404 #605]  INFO -- sentry: ** [Raven] Sending event 5db44a5151c44a8a8699a7ccaa2edf71 to Sentry
>        rake aborted!
>        Don't know how to build task 'assets:clean' (See the list of available tasks with `rake --tasks`)
>        /tmp/build_400672b08b175a998e4d7f78b88ee830/vendor/bundle/ruby/2.6.0/gems/rake-13.0.1/exe/rake:27:in `<top (required)>'
>        (See full trace by running task with --trace)
> 

## Why

The task `assets:precompile` is defined when using webpacker or sprocket, but the `assets:clean` is not defined when using only webpacker. 

The rails buidpack check the `assets:precompile` exist before calling it : 

https://github.com/heroku/heroku-buildpack-ruby/blob/1a30c5a0a4c189201a979d6d97bfb38a1aca2572/lib/language_pack/rails4.rb#L84-L85

So I think a simple fix could be to check wether the `assets:clean` exist too before calling it : 

https://github.com/heroku/heroku-buildpack-ruby/blob/1a30c5a0a4c189201a979d6d97bfb38a1aca2572/lib/language_pack/rails4.rb#L98-L99

And /or check also for `webpacker:clean` task and invoke that one when required.

What do you think?

Other stuff I found : 

- Webpacker "enhance" the `assets:clean` task, but only if it exist because sprocket is also used (https://github.com/rails/webpacker/pull/1744/files)
- Related PR https://github.com/heroku/heroku-buildpack-ruby/pull/892

schneems:
If the `assets:clean` task does not exist, don't call it.


lib/language_pack/rails4.rb
```diff
@@ -95,12 +95,15 @@ def run_assets_precompile_rake_task
           log "assets_precompile", :status => "success"
           puts "Asset precompilation completed (#{"%.2f" % precompile.time}s)"
 
-          puts "Cleaning assets"
-          rake.task("assets:clean").invoke(env: rake_env)
+          clean_task = rake.task("assets:clean")
+          if clean_task.not_defined?
```
hone, line 99: if the clean task is not defined we run it?

schneems: Right, also the test was wrong. Fixed both.

schneems:
The fact that my original code would have broken the buildpack but passed tests was an indicator that we should test that behavior. As a result, I added a test that would have failed with my original code https://github.com/heroku/heroku-buildpack-ruby/pull/1018/commits/a67261f4289c0be7e07a99b9d0f0d55f038e83b9. Your review still stands, i'm going to wait on merging this until next week after i'm back to work.

PR #1022, opened 2020-06-17
use correct lock file (not default one)

MathieuDerelle:
`bundler_version_escape_valve!` was moved in this commit https://github.com/heroku/heroku-buildpack-ruby/commit/4b3bfc6fe1659e9e7b0a00f29d3f0341a73ff029
and still uses the default lockfile

### DIFF
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -208,9 +208,9 @@ class LanguagePack::Helpers::BundlerWrapper
   end
 
   def bundler_version_escape_valve!
-    topic("Removing BUNDLED WITH version in the Gemfile.lock")
-    contents = File.read("Gemfile.lock")
-    File.open("Gemfile.lock", "w") do |f|
+    topic("Removing BUNDLED WITH version in the #{@gemfile_lock_path}")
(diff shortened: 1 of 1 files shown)


schneems:
Thanks for the PR. Can you help give me some more context especially this:

> still uses the default lockfile

What behavior are you seeing and what behavior do you expect? This still works as far as I know. If you've got a use case where this fails, please let me know and i'll be happy to take a look. Also if you have a Heroku support ticket number then I can see your application code if that helps to report the bug.

MathieuDerelle:
Yes I had a problem (not being able to use `BUNDLE_GEMFILE` on heroku) but this PR does not intend to fix this.

That problem made me go through this repo's code ; to find out it was intended.

I just stumbled upon that string for gemfile lock path and a variable containing it.

schneems:
> Yes I had a problem (not being able to use BUNDLE_GEMFILE on heroku) but this PR does not intend to fix this.

I've tried to support this in the past, but it's non-trivial. For now we only support the Gemfile.lock in the apps root path. If you need it somewhere else you can use the https://elements.heroku.com/buildpacks/lstoll/heroku-buildpack-monorepo or similar.

If you still want this PR to go in if you can give me a reproduction of the bug you're seeing then we can move forwards.

PR #1026, opened 2020-06-23
[changelog skip] Failing test for #1005

schneems:
The files in `bin/` such as `bin/rake` are come from the compiled binaries. For some versions of Ruby they start with a binstub that looks like this:

```
$ cd ~/Downloads/ruby-2.3.8
$ ls bin
erb	gem	irb	rake	rdoc	ri	ruby
$ cat bin/rake | head -n 1
#!/app/vendor/ruby-2.3.8/bin/ruby
```

This does not work at build due to the path `/app` not being where the app is.

This test fails because `bin/rake` has a bad shebang line when `rake assets:precompile` is run it will give an error:

```
       +remote: -----> Detecting rake tasks
       +remote:        Could not detect rake tasks
       +remote:        ensure you can run `$ bundle exec rake -P` against your app
       +remote:        and using the production group of your Gemfile.
       +remote:        bash: /tmp/build_d8662c03cfa0321567ab12e0a1e6f4f3/bin/rake: /app/vendor/ruby-2.4.9/bin/ruby: bad interpreter: No such file or directory
```

### DIFF
--- a/hatchet.lock
+++ b/hatchet.lock
@@ -36,7 +36,7 @@
 - - "./repos/node/minimal_webpacker"
   - bca8b8169c553a0f62f0263b621cf3a5813e4023
 - - "./repos/rack/default_ruby"
-  - da748a59340be8b950e7bbbfb32077eb67d70c3c
+  - master
 - - "./repos/rack/mri_187_nokogiri"
   - 3fcce9c85bf560dba285cc385ae9845729195826
 - - "./repos/rack/mri_192"
--- a/spec/hatchet/ruby_spec.rb
+++ b/spec/hatchet/ruby_spec.rb
@@ -1,6 +1,25 @@
 require_relative '../spec_helper'
 
 describe "Ruby apps" do
+  it "has an executable bin/rake with a good binstub" do
+    # https://github.com/heroku/heroku-buildpack-ruby/issues/1005
+    Hatchet::Runner.new("default_ruby", stack: "heroku-18").tap do |app|
+      app.before_deploy do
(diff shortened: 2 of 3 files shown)


schneems:
To comment here: I fixed the problem in #1005 by re-compiling binaries. More information is available on those comments

dzuelke:
I would recommend moving binaries to `.heroku/ruby/`, because `vendor/` is used by other languages, such as PHP. You then have a self-contained directory with binaries that is not shared with user code etc.

Then, you can simply do this in `bin/compile`:

```
[[ $build_dir == '/app' ]] || ln -s $build_dir/.heroku/ruby /app/.heroku/ruby
```

And the problem will be gone forever. This is what the PHP and Python buildpacks do for the same purpose.

schneems:
This test is supersceded by https://github.com/heroku/heroku-buildpack-ruby/pull/1031 as we're no longer placing the `bin/rake` binary in the application root.


Also to address the other comment:

> I would recommend moving binaries to .heroku/ruby/, because vendor/ is used by other languages, such as PHP. You then have a self-contained directory with binaries that is not shared with user code etc.

The issue is that codon (the app that runs the builds) is a Ruby application and lives in `/app`. It uses the Ruby buildpack to deploy so anything we do in the Ruby buildpack will be mirrored in Codon. This means any changes to `/app` will affect codon for the remainder of that run and doing something like changing out it's Ruby version will have adverse effects.

To run Ruby builds in `/app` we either need to change how codon is built and have it live somewhere other than `/app` and change all default builds to build in `/app` or just wait for CNB stuff to be the default which seems good.

PR #1033, opened 2020-07-08
CNB: make the gems layer accessible to subsequent buildpacks

kamaln7:
The `gems` layer writes a couple of environment variables to `#{layer.path}/env.build` ([ref 1](https://github.com/heroku/heroku-buildpack-ruby/blob/125191fdf57bf80f0cebf1f2d6009f77be6ae99a/lib/language_pack/ruby.rb#L164), [ref 2](https://github.com/heroku/heroku-buildpack-ruby/blob/be6cd5fe8327cc2686f11d07e73e2b489ba0347d/lib/language_pack/base.rb#L249)) but because [the `build` option](https://github.com/buildpacks/spec/blob/main/buildpack.md#build-layers) is not set on the layer, these environment variables are not made accessible to subsequent buildpacks.

This sets `build: true` on the `gems` layer to correct that.


schneems:
Thanks for the PR I'm running tests (plus added a test for this behavior) in https://github.com/heroku/heroku-buildpack-ruby/pull/1037

schneems:
thanks for the PR I merged my test of your code in!

kamaln7:
Thank you, @schneems!

PR #1037, opened 2020-07-15
CNB: make the gems layer accessible to subsequent buildpacks #1033 

schneems:
Followup from #1033  with tests.


schneems:
Thanks!

PR #1039, opened 2020-07-22
[close #943] Use Bundler env vars for config

### LINKED ISSUE #943: Bundler Deprecation warnings
I am getting deprecation warnings when deploying this buildpack:

```
       Running: bundle install --without development:test --path vendor/bundle --binstubs vendor/bundle/bin -j4 --deployment
       [DEPRECATED] The `--deployment` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set deployment 'true'`, and stop using this flag
       [DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path 'vendor/bundle'`, and stop using this flag
       [DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'development:test'`, and stop using this flag
       [DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs`
```

Using Ruby 2.7.0, BUNDLED WITH 2.1.2, Rails 6.0.2

It's not really a blocker, as the build passes, I just thought to bring it up here

schneems:
Starting with Bundler 2.x deprecation warnings started appearing noting that flags will no longer persist config when running `bundle install`:

```
       Running: bundle install --without development:test --path vendor/bundle --binstubs vendor/bundle/bin -j4 --deployment
       [DEPRECATED] The `--deployment` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set deployment 'true'`, and stop using this flag
       [DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path 'vendor/bundle'`, and stop using this flag
       [DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'development:test'`, and stop using this flag
       [DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs`
```

This is generally a confusing part of running bundler locally where someone on a team might try running `--bundle_without` once and then be confused at future behavior.

To support persisting values, we can use the bundler env vars that start with `BUNDLE_`. These will be set at:

- Build time
  - exported for other buildpacks
- Runtime

In addition, they're being shown in the `bundle install` output so if a developer gets a failure on Heroku they can copy and paste the command to test locally.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -445,7 +445,7 @@ EOF
       set_export_default "BUNDLE_PATH", ENV["BUNDLE_PATH"], layer
       set_export_default "BUNDLE_WITHOUT", ENV["BUNDLE_WITHOUT"], layer
       set_export_default "BUNDLE_BIN", ENV["BUNDLE_BIN"], layer
-      set_export_default "BUNDLE_DEPLOYMENT", ENV["BUNDLE_DEPLOYMENT"], layer if ENV["BUNDLE_DEPLOYMENT"]
+      set_export_default "BUNDLE_DEPLOYMENT", ENV["BUNDLE_DEPLOYMENT"], layer if ENV["BUNDLE_DEPLOYMENT"] # Unset on windows since we delete the Gemfile.lock
     end
   end
 
@@ -485,7 +485,7 @@ EOF
       set_env_default "BUNDLE_PATH", ENV["BUNDLE_PATH"]
       set_env_default "BUNDLE_WITHOUT", ENV["BUNDLE_WITHOUT"]
       set_env_default "BUNDLE_BIN", ENV["BUNDLE_BIN"]
-      set_env_default "BUNDLE_DEPLOYMENT", ENV["BUNDLE_DEPLOYMENT"] if ENV["BUNDLE_DEPLOYMENT"]
+      set_env_default "BUNDLE_DEPLOYMENT", ENV["BUNDLE_DEPLOYMENT"] if ENV["BUNDLE_DEPLOYMENT"] # Unset on windows since we delete the Gemfile.lock
     end
   end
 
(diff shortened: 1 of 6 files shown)


edmorley:
(Removed languages-review label since on-call but this seems best reviewed by @hone who is already set as reviewer :-))

lib/language_pack/ruby.rb
```diff
@@ -426,6 +441,11 @@ def setup_export(layer = nil)
         set_export_default "JAVA_OPTS",  default_java_opts
         set_export_default "JRUBY_OPTS", default_jruby_opts
       end
+
+      set_export_default "BUNDLE_PATH", ENV["BUNDLE_PATH"], layer
+      set_export_default "BUNDLE_WITHOUT", ENV["BUNDLE_WITHOUT"], layer
+      set_export_default "BUNDLE_BIN", ENV["BUNDLE_BIN"], layer
+      set_export_default "BUNDLE_DEPLOYMENT", ENV["BUNDLE_DEPLOYMENT"], layer if ENV["BUNDLE_DEPLOYMENT"]
```
hone, line 448: doesn't `setup_language_pack_environment` always set this to be `1`?

lib/language_pack/ruby.rb
```diff
@@ -393,6 +403,11 @@ def setup_language_pack_environment(ruby_layer_path:, gem_layer_path:)
       paths << ENV["PATH"]
 
       ENV["PATH"] = paths.join(":")
+
+      ENV["BUNDLE_WITHOUT"] = env("BUNDLE_WITHOUT") || bundle_default_without
+      ENV["BUNDLE_PATH"] = bundle_path
+      ENV["BUNDLE_BIN"] = bundler_binstubs_path
+      ENV["BUNDLE_DEPLOYMENT"] = "1"
```
hone, line 410: If we display all the env vars when running the command, do we want to make `BUNDLE_DEPLOYMENT` configurable?

lib/language_pack/ruby.rb
schneems: If there's a windows Gemfile.lock then we delete it and remove the deployment mode https://github.com/heroku/heroku-buildpack-ruby/pull/1039/files#diff-3fc66e13e389ff4d15c7ca3ddb86464eR886

lib/language_pack/ruby.rb
schneems: I don't think there's a reason to deploy without `--deployment` mode unless they're using a windows Gemfile.lock.

lib/language_pack/ruby.rb
hone: ah yeah, i wonder if it's worth just adding a comment.

schneems: I added a comment. Thinking about it a bit more, it's unclear this behavior is needed for the export and profile.d files i.e. once we've already run `bundle install` then there's not much benefit in keeping the deployment mode off. The PR as is preserves existing behavior, so let's not muck it up for now. Just keep in mind we could likely make that change if we wanted.

PR #1048, opened 2020-08-20
[changelog skip] Update hatchet

schneems:


edmorley (approved):
I'd recommend enabling Dependabot for these - saves needing to manually open a PR and ask someone else for review. For an example config file, see:
https://github.com/heroku/heroku-buildpack-python/blob/main/.github/dependabot.yml

PR #1051, opened 2020-08-28
[changelog skip] Dependabot

schneems:
Re: Comment from @edmorley https://github.com/heroku/heroku-buildpack-ruby/pull/1048

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "c: dependencies"
```
edmorley, line 8: The current label name in this repo is `dependencies` (without the `c: ` prefix):
https://github.com/heroku/heroku-buildpack-ruby/labels

Not sure if having the prefix here was intentional or not? :-)

edmorley (approved):

schneems:
Weird test failure, the setup never ran

image

Issue #1046, opened 2020-08-19
Slug size huge after last update


henrik:
Haven't looked into this deeply but thought I'd flag it ASAP in case others run into it.

Yesterday we got a slug size of ~230 MB but today it was ~900 MB, well over Heroku's 500 MB limit.

So far in our research, it seems the culprit may be https://github.com/heroku/heroku-buildpack-ruby/commit/0449897f249bacc66f57c62e891dd847a40daaba.

We locked the buildpack to the preceding commit, https://github.com/heroku/heroku-buildpack-ruby/commit/35463d0715ffd59bce292675f8a965725e6ba9f0 (by specifying it as `https://github.com/heroku/heroku-buildpack-ruby.git#35463d0715ffd59bce292675f8a965725e6ba9f0`), and then we managed to deploy again.

This is all with the same code on our side: a commit succeeded yesterday, then we made an `--allow-empty` commit on top of that which failed to deploy, and another `--allow-empty` which succeeded after locking the buildpack. (We did it that way to try to confirm the issue.)

This is a Rails 5.0 app on Ruby 2.6.6p146 bundled with Bundler 2.1.2.

edmorley:
@henrik Hi :-) Thank you for filing this issue and the excellent detail in it!

The Ruby buildpack maintainer will be online later and will dig in more, but for now for anyone else affected by this a workaround is to switch to the stable published version of the Ruby buildpack (the `heroku/ruby` shorthand alias) rather than using the edge/development version from `https://github.com/heroku/heroku-buildpack-ruby`. (See https://devcenter.heroku.com/articles/buildpacks and https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-buildpacks)

Another advantage of using the published buildpack release from the buildpack registry, is that the official language buildpacks are pre-cached on the build workers, so saves the git clone of the repository at the start of every build.

@schneems There are the following tickets related to this:
https://heroku.support/910531
https://heroku.support/910504
https://heroku.support/910502
https://heroku.support/910438
https://heroku.support/910389

henrik:
Also seeing it with Rails 4.2 on Ruby 2.6.6p146 bundled with Bundler 1.17.3 – it's the same Rails app. We had the poor timing of attempting a Rails upgrade that ran into this issue :)

Thank you, @edmorley! Good call with the stable one. Not sure why we've been using edge – will look into.

The first of those support issues (910531) is from our team :)

henrik:
[...]

PR #1052, opened 2020-08-28
Fix gem installation location bug with Bundler env vars

schneems:
When the move from bundler flags to bundler env vars was first merged in an issue was reported #1046 , this lead to an investigation and bug report to bundler https://github.com/rubygems/rubygems/issues/3890 which lead to some older issues/prs:
 
- rubygems/bundler#3552
- rubygems/bundler#6628

The issue is that global configuration and local configuration results in different behavior when installing and using some features in bundler, notedly the ability to specify the install path. Due to this change, when the switch to bundler environment variables happened, it caused the size of some applications' slugs to increase dramatically, this was because the gems were essentially being installed twice.

The issue appears to not be in Bundler 2.1.4 so we're bumping the version for 2.x series. In the 1.x series, an environment variable `BUNDLE_GLOBAL_PATH_APPENDS_RUBY_SCOPE=1` can be used to force the desired behavior.

### DIFF
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -100,12 +100,16 @@ WARNING
       warn_bad_binstubs
       install_ruby(slug_vendor_ruby, build_ruby_path)
       install_jvm
-      setup_language_pack_environment(ruby_layer_path: File.expand_path("."), gem_layer_path: File.expand_path("."))
-      setup_profiled(ruby_layer_path: "$HOME", gem_layer_path: "$HOME") # $HOME is set to /app at run time
+      setup_language_pack_environment(
+        ruby_layer_path: File.expand_path("."),
+        gem_layer_path: File.expand_path("."),
+        bundle_path: "vendor/bundle",
+        bundle_default_without: "development:test"
+      )
       allow_git do
         install_bundler_in_app(slug_vendor_base)
         load_bundler_cache
-        build_bundler(bundle_path: "vendor/bundle", default_bundle_without: "development:test")
+        build_bundler
         post_bundler
         create_database_yml
         install_binaries
@@ -114,6 +118,7 @@ WARNING
       config_detect
       best_practice_warnings
       warn_outdated_ruby
+      setup_profiled(ruby_layer_path: "$HOME", gem_layer_path: "$HOME") # $HOME is set to /app at run time
       setup_export
       cleanup
       super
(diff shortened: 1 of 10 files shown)


lib/language_pack/ruby.rb
```diff
@@ -426,6 +442,12 @@ def setup_export(layer = nil)
         set_export_default "JAVA_OPTS",  default_java_opts
         set_export_default "JRUBY_OPTS", default_jruby_opts
       end
+
+      set_export_default "BUNDLE_PATH", ENV["BUNDLE_PATH"], layer
+      set_export_default "BUNDLE_WITHOUT", ENV["BUNDLE_WITHOUT"], layer
+      set_export_default "BUNDLE_BIN", ENV["BUNDLE_BIN"], layer
+      set_export_default "BUDLE_GLOBAL_PATH_APPENDS_RUBY_SCOPE", ENV["BUDLE_GLOBAL_PATH_APPENDS_RUBY_SCOPE"], layer if bundler.needs_ruby_global_append_path?
```
edmorley, line 449:
```suggestion
      set_export_default "BUNDLE_GLOBAL_PATH_APPENDS_RUBY_SCOPE", ENV["BUNDLE_GLOBAL_PATH_APPENDS_RUBY_SCOPE"], layer if bundler.needs_ruby_global_append_path?
```

edmorley (commented):
> an issue was reported rubygems/rubygems#3890

Should this link be to heroku/heroku-buildpack-ruby/issues/1046 instead? (The Rubygems issue is linked twice)

schneems:
Updated with the right link. Also I added devcenter changelog entries.

edmorley (approved):
Sorry thought I'd already reviewed this! :-)

lib/language_pack/ruby.rb
```diff
@@ -839,44 +867,48 @@ def write_bundler_shim(path)
...
+            Bundler will do a full resolve so native gems are handled properly.
+            This may result in unexpected gem versions being used in your app.
+            In rare occasions Bundler may not be able to resolve your dependencies at all.
+
+            https://devcenter.heroku.com/articles/bundler-windows-gemfile
+          WARNING
         end
 
+        bundle_command = String.new("")
+        bundle_command << "BUNDLE_WITHOUT=#{ENV["BUNDLE_WITHOUT"]} "
```
edmorley, line 902: @schneems I think these env vars need to be quoted so that spaces in them (as seen in https://heroku.support/924687) don't result in bash running a command other than bundle

PR #1060, opened 2020-09-07
Merge upstream v218

Soulou:
Tested on:

* sample-ruby-sinatra.is-easy-on-scalingo.com
* sample-ruby-rails.is-easy-on-scalingo.com

Both with success


schneems:
Sorry, but i'm not going to rename the "heroku" buildpack to "scalingo" buildpack 😉 I think this PR was made by mistake.

Soulou:
@schneems Indeed, that was a mistake from a fork we're maintaining, my apologies we'll ensure it won't happen again

PR #1061, opened 2020-09-11
[changelog skip] ✂️ Hatchet#setup!

schneems:
Originally the Hatchet#setup! method only created an app and set config. Now it's responsible for also mutating files if there's a before_deploy action and for creating a git directory if using a non-git path.

Originally calling `setup!` was necessary before calling `deploy` if you wanted to do something like set config on an app before deploy. It was necessary because the Hatchet::Runner.new interface did not have a `config:` interface. That's no longer the case, so it's not needed.

Using this outside of the repo is dangerous. We should probably make it a private API. But...first lets remove it from our existing buildpacks.

While nothing in the ruby buildpack currently accidentally modifies the existing disk, it's purely by accident. If we were to copy and paste a test and modify it, it's possible that could change. Let's remove all use of `Hatchet#setup!` to be safe.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,7 @@ GEM
       excon
       moneta
       multi_json (>= 1.9.2)
-    heroku_hatchet (7.1.0)
+    heroku_hatchet (7.1.3)
       excon (~> 0)
       platform-api (~> 3)
       rrrretry (~> 1)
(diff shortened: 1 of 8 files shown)


hone (approved):

hone:
@schneems should we be updating the hatchet `README` to reflect this?

schneems:
It turns out that the java tests use `setup!` a bunch. I'm looking into how to make it a "safe" api and I think I have a possible solution.

PR #1071, opened 2020-09-28
Fix dependabot?

schneems:
Dependabot caps the message prefix to 15 characters:

This PR allows for [log skip] to be a valid replacement for [changelog skip]. It replaces the previous [ci skip] since we are required to have CI run on every PR anyway. 

Then we tell dependabot to use that interface.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,5 +7,3 @@ updates:
       interval: "weekly"
     labels:
       - "dependencies"
-    commit-message:
-      prefix: "[changelog skip]"
--- a/.github/workflows/check_changelog.yml
+++ b/.github/workflows/check_changelog.yml
@@ -5,6 +5,7 @@ on:
   types: [opened, reopened, edited, synchronize]
 jobs:
   build:
+    if: !contains(github.event.issue.labels.*.name, "dependencies")
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1


edmorley (commented):
What do you think about instead exempting the Dependabot PRs via the PR author?

The user info is available in the GitHub PR event payload:
https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request

Which can be used like:
https://github.com/heroku/heroku-buildpack-python/blob/main/.github/workflows/check_changelog.yml#L10-L13

The advantages of this are:
* conditionals in the `if` key cause the check to show as "skipped" (grey) rather than a green check, meaning clearer what's happened + action completes sooner (since doesn't need to spin up a container)
* saves adding another variant of commit message prefix (inconsistency) and clutter in the commit log (not a big fan of the prefixes as you know :-))

schneems:
In the short term this problem is blocking me from merging https://github.com/heroku/heroku-buildpack-ruby/pull/1044. Ideally, we could standardize on our dependabot.yml files in the future. For now, I'm mostly looking to merge that other PR.

I'm okay on doing an exception based on author, but I'm can see a world in which dependabot becomes dependabot2 or something and it breaks and we have to change it again. In general I would prefer for my tooling to use standard interfaces than to have to make exceptions for each tool. Pragmatically though I'm not too worried about it. I think either way is fine.

edmorley:
> but I'm can see a world in which dependabot becomes dependabot2 or something and it breaks and we have to change it again

What about using the `dependencies` label? It's specified in this repo's dependabot config, so can't change unintentionally.

schneems:
Updated. It can also be accidentally/invisibly skipped if someone is using that label without realizing it has side effects though.

edmorley (approved):

PR #1083, opened 2020-10-07
Allow support for BUNDLE_WITHOUT with spaces

schneems:
In v218 we switched from configuring bundler via flags to env vars. It turns out that bundler supports defining an `--without` value with a space in it, but it does not support BUNDLE_WITHOUT with a space in it:

```
$ bundle install --without foo bar
$ bundle config | grep without -a1

without
Set for your local app (/private/tmp/40a4b42d7e884ebf62a1f85a3c5abd38/.bundle/config): [:foo, :bar]
```

But:

```
$ BUNDLE_WITHOUT="foo bar" bundle config | grep without -a1

without
Set via BUNDLE_WITHOUT: [:"foo bar"]
```

> Note in this example it is showing `:"foo bar"` as one group instead of two values.

To fix this, we can reproduce the bundler logic of replacing a space with a colon https://github.com/rubygems/rubygems/blob/ce27b98272d5c37ebf45b2b1b66894699ae47d58/bundler/lib/bundler/cli/install.rb#L151. 

Related ticket: https://heroku.support/924687

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Main (unreleased)
 
+* BUNDLE_WITHOUT now accommodates values with single spaces (https://github.com/heroku/heroku-buildpack-ruby/pull/1083)
+
 ## v219 (8/6/2020)
 
 * Fix double installation of bundler on CI runs when no test script is specified (https://github.com/heroku/heroku-buildpack-ruby/pull/1073)
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -406,6 +406,11 @@ EOF
       ENV["PATH"] = paths.join(":")
 
       ENV["BUNDLE_WITHOUT"] = env("BUNDLE_WITHOUT") || bundle_default_without
+      if ENV["BUNDLE_WITHOUT"].include?(' ')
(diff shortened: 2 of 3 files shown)


danielleadams (approved):

edmorley (commented):
Hi! Would it be worth also wrapping all of the env vars in quotes, to avoid malformed vars breaking the bash command, here?
https://github.com/heroku/heroku-buildpack-ruby/pull/1052#discussion_r500846817

schneems:
@edmorley seems good. It's this now:

```
       remote:        Running: BUNDLE_WITHOUT='foo:bar:baz' BUNDLE_PATH=vendor/bundle BUNDLE_BIN=vendor/bundle/bin BUNDLE_DEPLOYMENT=1 BUNDLE_GLOBAL_PATH_APPENDS_RUBY_SCOPE=1 bundle install -j4
```

edmorley:
@schneems Yeah I realise the bug is fixed; I mean more that normally when writing a bash script, it's best practice to put all variable usage inside double quotes (eg it's a shellcheck lint error to not do so). Whilst the bash command for the bundle install is an inline concatenation in Ruby, this still seems like something worth doing.

PR #1103, opened 2020-12-18
Add webpacker caching support

danielwestendorf:
Cache the default webpacker folders. This consists of the compiled output and a last-compilation-digest-ENV file, resulting in re-use of the existing pack files when the manifest. This cut `rails assets:precompile` down from 240.79s -> 42.85s on one of my apps.

Webpacker obviously wasn't a thing with rails4, but I chose to implement here quickly as a proof of concept.

### DIFF
--- a/lib/language_pack/rails4.rb
+++ b/lib/language_pack/rails4.rb
@@ -56,8 +56,16 @@ WARNING
     "public/assets"
   end
 
+  def public_webpacker_folder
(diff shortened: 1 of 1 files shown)


danielwestendorf:
Closing because of https://github.com/heroku/heroku-buildpack-ruby/pull/892#issuecomment-517811562

PR #1108, opened 2021-01-14
Handle BUNDLED WITH deletion in windows

schneems:
Windows Gemfile.lock uses `\r\n` instead of `\n` for newlines which can cause issues when matching regex. We are using regex to strip out BUNDLED WITH and were not previously handling windows Gemfile.lock by mistake.

There is other code that reads in the major version of BUNDLED WITH and it uses a read mode of `rt` which solves the problem.

The solution is to change this other location to also use a read mode of `rt`:

```
irb(main):009:0> !!File.read("./spec/fixtures/windows_lockfile/Gemfile.lock").match(/^BUNDLED WITH$(\r?\n)   (?<major>\d+)\.\d+\.\d+/m)
=> false
irb(main):010:0> !!File.read("./spec/fixtures/windows_lockfile/Gemfile.lock", mode: 'rt').match(/^BUNDLED WITH$(\r?\n)   (?<major>\d+)\.\d+
\.\d+/m)
=> true
```

The test now also assert the desired behavior of removing BUNDLED WITH occurs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Main (unreleased)
 
+* Fix Gemfile.lock read bug from preventing propper removal of BUNDLED WITH declaration ()
+
 ## v222 (11/02/2020)
 
 * CNB support for Heroku-20 (https://github.com/heroku/heroku-buildpack-ruby/pull/1096)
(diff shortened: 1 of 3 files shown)


edmorley (approved):
Good spot! :-)

schneems:
Related ticket https://heroku.support/950520

schneems:
Default stack bump made some tests fail, I'll fix that in another PR and rebase when done.

raldred:
Awesome thanks for getting this sorted @schneems

schneems:
Happy to help. I'll let this sit on `main` for a few days to test for stability. I'll likely aim to deploy next week Wednesday or Thursday, but that could slip based on other workloads.

PR #1109, opened 2021-01-14
Remove cedar-14 specs and get all other "default stack" tests to work with Heroku-20 (the new default)

schneems:


edmorley (approved):
Pre-emptively approving so you can merge once CI passes :-)

PR #1110, opened 2021-01-17
Caching Webpacker assets between builds

MikeRogers0:
I noticed between deploys the `public/packs` directory isn't cached then restored, which caused assets to be rebuilt by webpacker. This is kind of annoying if the JavaScript assets aren't changing between commits.

I also saw https://twitter.com/andrewculver/status/1350857304913154052 which made me think people are using multiple buildpacks to achieve this, so I think this would be reasonable to add to the default heroku/ruby buildpack.

I added a step to `LanguagePack::Rails6` which restores & caches `public/packs`, `tmp/cache/webpacker` & `node_modules`.

# Screenshots

<details><summary>Before ~45s</summary>

image

</details>

<details><summary>After (Without caching of `/tmp/cache/webpacker`) ~15s</summary>

image

This reduced the webpack compile down by ~30 seconds.

</details>

<details><summary>After (While caching `/tmp/cache/webpacker`)  ~3s</summary>

image

This avoid the precompliation completely, so it takes only ~3s
</details>

# Implementation Details

`public/packs` - This is where Webpacker will output its compiled assets to. After it's compiled it does run `rake assets:clean`, so I'm not to worried about expired assets being stored.
`node_modules` - Webpacker does call `yarn` to satisfy the NPM dependencies, so this is a fairly common folder for Rails app to have.


lib/language_pack/rails6.rb
```diff
@@ -14,6 +14,32 @@ def self.use?
...
+  def node_modules_folder
+    "node_modules"
+  end
+
+  def public_packs_folder
+    "public/packs"
+  end
+
+  def webpacker_cache_folder
+    "tmp/cache/webpacker"
```
MikeRogers0, line 26: This path is the [default value](https://github.com/rails/webpacker/blob/f9c29dd08268de6bddf4de50422258864938f75c/test/mounted_app/test/dummy/config/webpacker.yml#L7) for where Webpacker caches assets. As the other paths are hardcoded, I think this is acceptable.

ParamagicDev:
```rb
Webpacker.config.cache_path
```

^ the above should give you the cache path.

https://github.com/rails/webpacker/blob/6efb3ee5367dbced7b76b7daec35684db4d9570a/lib/webpacker/configuration.rb#L50

MikeRogers0: @ParamagicDev This is a bit tricker then I first thought 😬 

To check if webpacker is installed & get that info, I came up with:

```
bundle exec rails runner 'puts Webpacker.config.cache_path if defined?(Webpacker)'  
```

Though, I'm not a fan as if webpacker was to change that method we'd have to update this code. I think it's more resilient to just hardcode the path & let people who stick to defaults feel the speed.

schneems:
Hey. Thanks for the PR! As a heads up you should know that I'm working on a total re-write of the buildpack. You can see the progress here:

- https://github.com/heroku/heroku-buildpack-ruby-experimental

There's quite a bit of history of trying to get caching for webpack on the Ruby/Rails buildpack https://github.com/heroku/heroku-buildpack-ruby/pull/892#issuecomment-721368448. 

> I also saw https://twitter.com/andrewculver/status/1350857304913154052 which made me think people are using multiple buildpacks to achieve this

The benefit of multiple buildpacks is that you aren't blocked by me for getting some behavior that you want. Also using them enables development and use metrics on a feature. I do this with PRs to repos like Rails all the time. Instead of sending a feature over, I gem-ify it. Sometimes that's the end and sometimes it gets fully upstreamed to Rails.

andrewculver:
Hey Richard! For my part, I really, really want to see this functionality incorporated into the official build pack from Heroku, whether that's the current version or the new hotness.

One big disadvantage of third-party buildpacks is that they're subject to [supply chain attacks](https://en.wikipedia.org/wiki/Supply_chain_attack), and at the same time they don't have the same scale of adoption and supporting infrastructure that dependency ecosystems like Ruby gems and npm have for blasting out notifications to people when vulnerabilities are discovered or compromises happen.

The workflow for using a third-party buildpack seems to be:

1. Find a reputable source for the buildpack you're looking for, for some value of "reputable". 
2. Audit all the code to make sure you're not about to get owned.
3. [Fork the code to mitigate the risk of a supply chain attack.](https://twitter.com/dwestendorf/status/1350973312722227206)

If everyone is acting responsibly, basically that work get repeated on every single team that wants the desired functionality. As far as I can tell, the only way around that is for Heroku to own the desired functionality. I can only speak for myself here, but I think it would be true for many folks: If Heroku owns it, I no longer feel the need to go through all those steps, because I actually trust Heroku more than myself to not get owned. Put another way: If Heroku gets owned, I'm already toasted. This is, among many other things, what I hope I'm paying for!

I think this functionality is desirable for 100% of new Rails applications (which all have Webpacker enabled by default). I'm not sure which developers would want their "asset precompilation" step to take [238 seconds](https://twitter.com/andrewculver/status/1349911608672268291?s=20) when it could take [27 seconds](https://twitter.com/andrewculver/status/1350857304913154052?s=20) (my results recently).

> I do this with PRs to repos like Rails all the time. Instead of sending a feature over, I gem-ify it. Sometimes that's the end and sometimes it gets fully upstreamed to Rails.

I think the existing third-party buildpacks, [which go back to 2014](https://github.com/zakjan/heroku-buildpack-cacheload), represented that initial attempt to "Gem-ify" it. I think what we're specifically hoping for now is to see that solution standardized and maintained upstream for the benefit of every Heroku customer using Rails. 🙏🤞

danielwestendorf:
#1103 is another PR attempting to accomplish this. There is a good amount of customer demand to solve this pain point, and relying on 3rd party plugins isn’t viable (for reasons @andrewculver mentioned above).

Please, please, please consider an alternative response on this one, this could drive generations of stock rails 5.1+ apps away from heroku. Let us continue to love the platform! 💙

MikeRogers0:
I'm hoping we can change your mind @schneems :)

There appears to be demand for this to be in `heroku/ruby` based on https://github.com/heroku/heroku-buildpack-ruby/pull/1103 & https://github.com/heroku/heroku-buildpack-ruby/pull/892, as Webpacker is currently installed by default for new Rails projects I think it would improve the experience of a lot of users. 

I don't know the numbers of people using custom buildpacks (or both `heroku/node` & `heroku/ruby`), but personally I stick to official defaults whenever possible (Which is just `heroku/ruby` on most my rails projects). I don't think I'm in the minority in doing that :)

cabgfx:
> […] but personally I stick to official defaults whenever possible (Which is just `heroku/ruby` on most my rails projects). I don't think I'm in the minority in doing that :)

Same, @MikeRogers0 — would love to see these optimizations added to the default buildpack herre.

granth:
This actually _increased_ our build time, from ~1600 s to ~1800 s. (Yes, we have too many assets.)

The "Asset precompilation completed" time did drop by 100–200 s.

I assume the change resulted in a large cache and the load and store steps are slower as a result.

MikeRogers0:
> I assume the change resulted in a large cache and the load and store steps are slower as a result.

@granth How many times did you run it? When I tried it first deploy was slower, but subsequent deploys did see a performance boost.

granth:
@MikeRogers0 At least five builds. I waited to be sure I saw a build that started after the first build and therefore got the new cache.

granth:
Update: I tried this branch again after discovering I had set `additional_paths` incorrectly in `webpacker.yml`.

No big speed boost, unfortunately, but the build times are at worst about the same as with the regular `heroku/ruby` buildpack 👍

granth:
Update: with more builds it looks like the build time is averaging a bit longer again. ~1700 s.

I downloaded the cache directory using the `build-metadata` endpoint as in the `heroku-repo` plugin: https://github.com/heroku/heroku-repo/blob/master/lib/repo.js

```
heroku-cache → du -sh tmp/cache/webpacker public/packs/js node_modules
 34M	tmp/cache/webpacker
374M	public/packs/js
1.3G	node_modules
```

💀

MikeRogers0:
@granth 😬 I'm guessing yarn isn't removing older versions of stuff from node_modules. I'll have a mini experiment this weekend to see if I can speed it up :)

granth:
@MikeRogers0 Sorry, I don't think I can share the `package.json`. I measured `node_modules` after an `rm -rf node_modules && yarn` and it's still 1.2 GB 😬 so I don't think cache cleaning is the issue there.

jonatasdaniel:
Hey @MikeRogers0 , do you have any update on this?
I tried your branch and no improvements :/

MikeRogers0:
@jonatasdaniel  How many deploys did you do? It caches the node_modules & webpack cache, so first deploy isn't much faster but subsequent deploys should be :)

jonatasdaniel:
> @jonatasdaniel How many deploys did you do? It caches the node_modules & webpack cache, so first deploy isn't much faster but subsequent deploys should be :)

3 deploys, all same time.
I just replaced default ruby buildpack with your repo.

ParamagicDev:
If anyone wants to speed up webpacker build times for your JavaScript assets, you can try using ESBuild. I have a gist of how to set it up here:

https://gist.github.com/ParamagicDev/54759bc8553a22d507fbc97e4eae14cc

https://esbuild.github.io

MikeRogers0:
@ParamagicDev That looks like a way better solution!

dankimio:
These changes reduced our asset precompilation time from 350s down to 15s.

MikeRogers0:
I'm going to close this PR. Based on what @schneems commented https://github.com/heroku/heroku-buildpack-ruby/pull/892#issuecomment-721368448 it's unlikely to to be merged.

justin808:
I agree that the responsibility of caching the output of rails/webpacker shouldn't be part of the ruby-buildpack's responsibility. I shared an alternative: 

https://github.com/heroku/heroku-buildpack-ruby/pull/892#issuecomment-851722678

PR #1115, opened 2021-01-27
Rails 2, 3, and 4 LTS tests

schneems:
Rails version support is limited by a few things:

- Rails library support by maintainers
- Ruby version support
- Behavior support in the Ruby buildpack

## Ruby version support

We only provide the latest non-EOL versions of Rubies on stacks. That means when cedar-14 went away, older versions of Ruby such as Ruby 1.9 are no longer available on the platform.

## Rails library support by maintainers

We do not patch Rails versions. This includes not patching for security, and not patching for Ruby version compatibility. A customer's ability to run any given Rails version on the platform is limited by the version of Ruby that it can run on. 

RailsLTS backports Ruby version compatibility to Rails 2, 3, and 4.

## Behavior support in the Ruby buildpack

We will continue providing support for the behavior provided by the buildpack to Rails 2,3, and 4 customers provided. However we hit an issue where we could not run tests for these behaviors because the versions on Rubygems do not work with versions of Ruby available on current stacks. 

To address this we're working with the RailsLTS team who have given us a license to use for our test suite. Now these versions of Rails can continue to be supported by Heroku maintainers along with tests for that behavior.


edmorley (approved):
🎉

schneems:
Apparently my check changelog action isn't firing :(

PR #1119, opened 2021-02-05
Use jvm buildpack to install JRuby dependencies

schneems:
Currently the jvm logic is duplicated in this buildpack and the jvm buildpack. By using the JVM buildpack we can reduce duplication and increase consistency across the Heroku platform.


edmorley:
@schneems Is this ready for review? The reviewers feature has requested review from the languages team due to codeowners, but it seems that the CI is being iterated on for the moment. Any objection to converting this PR to a draft, to make the "doesn't need review yet" state clearer? :-)

schneems:
This is now ready for review. Also It's got a rebase from the fixes in #1122

edmorley (approved):

PR #1122, opened 2021-02-08
[changelog skip] This fixes the tests and I don't know why

schneems:
Recently the CI tests on main started failing:

```
  1) RubyVersion correctly sets default legacy version

     Failure/Error:
       Hatchet::App.new("default_ruby").in_directory_fork do |dir|
         ruby_version   = LanguagePack::RubyVersion.new(@bundler.install.ruby_version, is_new: false)
         version_number = LanguagePack::RubyVersion::LEGACY_VERSION_NUMBER
         version        = LanguagePack::RubyVersion::LEGACY_VERSION
         expect(ruby_version.version_without_patchlevel).to eq(version)
         expect(ruby_version.engine_version).to eq(version_number)
         expect(ruby_version.to_gemfile).to eq("ruby '#{version_number}'")
         expect(ruby_version.engine).to eq(:ruby)
       end
     RuntimeError:
       -----> Installing bundler 1.17.3
       -----> Removing BUNDLED WITH version in the Gemfile.lock
       /app/lib/language_pack/helpers/bundler_wrapper.rb:151:in `block in ruby_version': There was an error parsing your Gemfile, we cannot continue (LanguagePack::Helpers::BundlerWrapper::GemfileParseError)
       /app/vendor/bundle/ruby/2.6.0/gems/bundler-2.1.4/lib/bundler/spec_set.rb:86:in `block in materialize': Could not find rake-13.0.1 in any of the sources (Bundler::GemNotFound)
```

I'm honestly not sure how this failure seemingly came out of nowhere. 

The fix is to wrap all calls to `@bundle.install` in a `Bundler.with_unbundled_env` block which essentially "resets" the bundler environment to how it was before the bundle exec mutated it.

I'm fine with doing that in these locations. It's a common strategy, for example: https://github.com/heroku/heroku-buildpack-ruby-experimental/blob/c210f289312e6185a60b953396f57f2ef0b60c6a/spec/unit/ruby_detect_version_spec.rb#L82

The thing I don't understand is why it's only needed now, or what changed to make this needed.

### DIFF
--- a/spec/helpers/bundler_wrapper_spec.rb
+++ b/spec/helpers/bundler_wrapper_spec.rb
@@ -42,34 +42,44 @@ describe "BundlerWrapper" do
 
   it "detects windows gemfiles" do
     Hatchet::App.new("rails4_windows_mri193").in_directory_fork do |dir|
-      expect(@bundler.install.windows_gemfile_lock?).to be_truthy
+      Bundler.with_unbundled_env do
+        expect(@bundler.install.windows_gemfile_lock?).to be_truthy
+      end
     end
   end
 
   describe "when executing bundler" do
-    before do
-      @bundler.install
-    end
-
     it "handles apps with ruby versions locked in Gemfile.lock" do
       Hatchet::App.new("problem_gemfile_version").in_directory_fork do |dir|
-        expect(@bundler.ruby_version).to eq("ruby-2.5.1-p0")
+        Bundler.with_unbundled_env do
+          @bundler.install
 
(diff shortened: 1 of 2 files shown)


schneems:
This is ready for review. I didn't want to fix the commit title because frankly I have no extra information to add, it's already linking back to this PR, and I've rebased another commit against this one.

edmorley:
`main` is currently green, so presuming this is no longer needed?

Closing to remove it from the languages review queue.

PR #1124, opened 2021-02-09
Add bootchecking for rails applications

### LINKED ISSUE #1120: Rails smoke test during release phase?
Hi, the buildpack currently does some fingerprinting to notice whether the ruby app being deployed is a rails app by attempting to run `rails runner`.

If the application _is_ a rails application but for some reason can't be loaded properly (maybe some configuration differences between CI and production) this test will fail but the release will still succeed.

We've adding a rake task to all of our applications as part of the `release` task in our Procfile that does runs `rails runner` and if there is a failure returns the non-zero exit code causing the release to fail and therefore does not actually release the applications.

This can catch some awkward issues that can slip through dev and CI and cause apps to deploy but not be able to actually process any requests!

```ruby
require 'open3'

task final_frontier: :environment do
  _, stderr, status = Open3.capture3("rails runner 'puts Rails.env'")

  raise "Rails app failed to boot:\n\n#{stderr}" unless status.success?
end
```

Is this something that could be useful to be included in this buildpack?
Maybe only added if there is an explicit ENV var that denotes that the applications _is_ expected to be running rails?

CGA1123:
Adds "RailsBootcheck" to the Rails buildpack.

RailsBootcheck attempts to load your rails application and will fail the
the build of your application if it is unable to load.

This can help prevent scenarios where you can accidentally ship a
version of your application that cannot boot properly. Which will lead
to errors as the application is released and traffic is routed to it.

You can opt-in to this feature via the `HEROKU_RAILS_BOOTCHECK_ENABLE`
variable, new rails applications will have this variable set by default.

You can opt-out of this feature via the `HEROKU_RAILS_BOOTCHECK_DISABLE`
variable.

Closes #1120

### DIFF
--- a/lib/language_pack/helpers/rails_bootcheck.rb
+++ b/lib/language_pack/helpers/rails_bootcheck.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+class LanguagePack::Helpers::RailsBootcheck
+  include LanguagePack::ShellHelpers
+
+  def initialize(timeout = 65)
+    @timeout = timeout
+  end
+
+  def call
+    return unless opted_in? && !opted_out?
+
+    topic("Bootchecking rails application")
+
+    process = ProcessSpawn.new(
+      "rails runner 'puts Rails.env'",
+      user_env: true,
+      timeout:  @timeout,
+      file:     "./.heroku/ruby/compile/rails_bootcheck.txt"
+    )
+
+    if process.timeout?
+      failure("timeout", process.output)
+    elsif !process.success?
+      failure("failure", process.output)
+    end
+  end
+
+  private
+
+  def failure(type, output)
+    message = String.new("Bootchecking rails application #{type}\n")
+    message << "set HEROKU_RAILS_BOOTCHECK_DISABLE=1 to disable this feature\n"
+
+    if !output.empty?
+      message << "\n"
+      message << output
+    end
+
+    error(message)
+  end
+
+  def opted_in?
+    env("HEROKU_RAILS_BOOTCHECK_ENABLE")
+  end
+
+  def opted_out?
+    env("HEROKU_RAILS_BOOTCHECK_DISABLE")
+  end
+end
--- a/lib/language_pack/rails2.rb
+++ b/lib/language_pack/rails2.rb
@@ -1,6 +1,7 @@
 require "fileutils"
 require "language_pack"
 require "language_pack/rack"
+require "language_pack/helpers/rails_bootcheck"
 
 # Rails 2 Language Pack. This is for any Rails 2.x apps.
 class LanguagePack::Rails2 < LanguagePack::Ruby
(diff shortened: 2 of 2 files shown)


CGA1123:
I wanted to open this PR to see whether this is going in the right direction so far.

It's a bit late already for me to look into getting `hatchet` running today, but I'll make some time tomorrow to do so and get some coverage on these changes.

Thanks 😄

schneems:
Hey, thanks for the PR! I'm kind of a one-person show right now and feature development on the buildpack tends to be very slow as stability is usually the main driver. There's also other issues (as you've noticed) like our test suite doesn't run for PRs due to sensitive creds and running the suite locally takes a LOOONG time (and can't even execute all the tests).

(Also I just merged in something pretty ambitious around JVM changes. I need to roll that out slowly and carefully).

I appreciate the PR, though. I'm not sure when I'll be able to get around to focusing on this and rolling it out. I think that I would re-use the RailsRunner logic and add a step at the end of `def compile` and `def build` to do the check and the logic. 

I like that the logic is wrapped up in it's own class that can be tested independently. I'm also in the process of a re-write over at https://github.com/heroku/heroku-buildpack-ruby-experimental and that's the dominant pattern I'm using. 

I do want to set the expectation that most larger features like this are not merged wholesale which is unfortunate but also the nature of buildpack development as anything that goes in needs to be supported (quite literally, via customer support tickets) for years to come. 

So if you're interested in tweaking this based on that feedback, then I would say go for it. But keep in mind that I might not be able to give you movement or even propper feedback and attention for some long cycles. When I'm ready to revisit this I might be able to pull your change in and work on-top of it, but likely we'll not be able to get to a place where I could just merge and ship the whole thing.

CGA1123:
That makes sense. I appreciate the requirements for the buildpacks are different than in most projects and that rolling out is slow and gradual! People want stable buildpacks, companies and developers rely on that (including me, so thank you for keeping the bar high 😄)!

I'm happy to tweak this changeset to make use of the `RailsRunner`, adding an additional check on `success?` which raises if this feature is enabled. And hopefully getting this to a place where it is as low friction for you as possible when/if an opportunity comes up to work on this.

edmorley (commented):
(Cancelling review to get this out of the languages team review queue, as it's already had a review from the Ruby SME: https://github.com/heroku/heroku-buildpack-ruby/pull/1124#issuecomment-777624170)

schneems:
Realistically I won't be merging this in, so I think it's better to close. Sorry for the delay and keeping you waiting for so long. I also think this could be extracted into it's own buildpack that works with any app. I'm thinking:

- Read the Procfile, extract `web` (this format isn't EXACTLY yaml, but it's close enough you can use a yaml parser on it for 99% of the cases)
- Run that command in the background with a specified PORT
- Repeatedly hit that port for X seconds until it receives a successful response or the background process crashes
- If all is well, exit cleanly, otherwise print the logs from the server and exit with an error.

Issue #1106, opened 2021-01-11
Gemfile.lock built with Bundler 2.2.x fails


schneems:
From https://github.com/rubygems/rubygems/issues/4269

### Describe the problem as clearly as you can

Here's an example repo: https://github.com/schneems/bundler_223_fails_deploy. It has one dependency `rake` and the Gemfile.lock was generated by running `bundle install` on my mac. When I try to use this Gemfile/Gemfile.lock combo it fails:

```
docker run -it --rm ruby:2.7 bash
git clone https://github.com/schneems/bundler_223_fails_deploy
cd bundler_223_fails_deploy
export BUNDLE_WITHOUT='development:test'
export BUNDLE_DEPLOYMENT=1
export BUNDLE_PATH=vendor/bundle
bundle install -j4
```

Then

```
 bundle exec rake -P
bundler: failed to load command: rake (/bundler_223_fails_deploy/vendor/bundle/ruby/2.7.0/bin/rake)
Gem::Exception: can't find executable rake for gem rake. rake is not currently included in the bundle, perhaps you meant to add it to your Gemfile?
  /usr/local/lib/ruby/2.7.0/bundler/rubygems_integration.rb:374:in `block in replace_bin_path'
  /usr/local/lib/ruby/2.7.0/bundler/rubygems_integration.rb:402:in `block in replace_bin_path'
  /bundler_223_fails_deploy/vendor/bundle/ruby/2.7.0/bin/rake:23:in `<top (required)>'
```

### What were you expecting to happen?

I expect that a Gemfile.lock produced on mac can also work on linux

### What actually happened?

The gem appears to install, but then it cannot be loaded.

### Extra context 

Either adding `ruby` to the `PLATFORMS` section or `x86_64-linux` fixes the issue, but this is not immediately clear.

### If not included with the output of your command, run `bundle env` and paste the output below

<details>
  <summary>bundle env (on my mac)</summary>
## Environment

```
Bundler       2.2.3
  Platforms   ruby, x86_64-darwin-19
Ruby          3.0.0p0 (2020-12-25 revision 95aff214687a5e12c3eb57d056665741e734c188) [x86_64-darwin19]
  Full Path   /Users/rschneeman/.rubies/ruby-3.0.0/bin/ruby
  Config Dir  /Users/rschneeman/.rubies/ruby-3.0.0/etc
RubyGems      3.2.3
  Gem Home    /Users/rschneeman/.gem/ruby/3.0.0
[...]

PR #1132, opened 2021-02-25
[close #1126] Update bundler to 2.2.11

### LINKED ISSUE #1126: Bundler < 2.2.10 has "Source Priority" vulnerability
> The developer was able to expose (in a non-malicious way) a vulnerability present in well-known dependency managers, where given a library name they will end up preferring installing it from a public source rather than from a private source.
>
> Unfortunately, Bundler had this vulnerability.
>
> The issue has been fixed in bundler 2.2.10
> https://bundler.io/blog/2021/02/15/a-more-secure-bundler-we-fixed-our-source-priorities.html

AFAICT, this buildpack is currently using bundler 2.1.4 (`BLESSED_BUNDLER_VERSIONS`).

I know that a lot of care and thought goes into bundler upgrades. So, I'm not suggesting a specific course of action, but it is a vulnerability that concerns me.

schneems:
Bundler < 2.2.10 has "Source Priority" vulnerability #1126. This PR updates bundler to the latest version (2.2.11).

### DIFF
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -37,7 +37,7 @@ class LanguagePack::Helpers::BundlerWrapper
 
   BLESSED_BUNDLER_VERSIONS = {}
   BLESSED_BUNDLER_VERSIONS["1"] = "1.17.3"
-  BLESSED_BUNDLER_VERSIONS["2"] = "2.1.4"
+  BLESSED_BUNDLER_VERSIONS["2"] = "2.2.11"
   BUNDLED_WITH_REGEX = /^BUNDLED WITH$(\r?\n)   (?<major>\d+)\.\d+\.\d+/m
 
   class GemfileParseError < BuildpackError
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Main (unreleased)
 
+* Bundler 2.x is now 2.2.11 (https://github.com/heroku/heroku-buildpack-ruby/pull/1132)
+
 ## v224 (2/25/2021)
 
 * Ruby buildpack now relies on the JVM buildpack to install java for Jruby apps (https://github.com/heroku/heroku-buildpack-ruby/pull/1119)
--- a/changelogs/unreleased/bundler_2_2_11.md
+++ b/changelogs/unreleased/bundler_2_2_11.md
@@ -1,0 +1,3 @@
+##  Ruby apps with Bundler 2.x now receive version 2.2.11
+
+The [Ruby Buildpack](https://devcenter.heroku.com/articles/ruby-support#libraries) now includes Bundler 2.2.11. Applications specifying Bundler 2.x in their `Gemfile.lock` will now receive this version of bundler.


edmorley (approved):

lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -37,7 +37,7 @@ class LanguagePack::Helpers::BundlerWrapper
 
   BLESSED_BUNDLER_VERSIONS = {}
   BLESSED_BUNDLER_VERSIONS["1"] = "1.17.3"
-  BLESSED_BUNDLER_VERSIONS["2"] = "2.1.4"
+  BLESSED_BUNDLER_VERSIONS["2"] = "2.2.11"
```
dentarg, line 40: @schneems are you sure you want everyone with Bundler 2.1.x in their lockfile to get Bundler 2.2.11? https://github.com/heroku/heroku-buildpack-ruby/issues/1126#issuecomment-786236309

schneems: Yes, we only have a single "blessed" version for 1.x and 2.x Bundler versions. It's not just 2.1 its 2.x that will get this version. 

Heroku needs to have a tight coupling on the bundler version to be able to provide support for it. Otherwise, we end up playing edge-case whack-a-mole and general behavior is undefined.

jaredbeck:
I encountered the [documented issue with 2.2.3](https://devcenter.heroku.com/articles/bundler-version#bundler-2-2-3). The documented fix worked for me, and the build output links to the docs, but I'm just mentioning it here also, in case it helps someone. 

Example failure output:

```
remote:        Your app was upgraded to bundler 2.2.11.
remote:        Previously you had a successful deploy with bundler 2.1.4.
..
remote:        Unable to find a spec satisfying sorbet-static (= 0.5.6300) in the set. Perhaps
remote:        the lockfile is corrupted? Found sorbet-static (0.5.6300-universal-darwin-14)
remote:        that did not match the current platform.
..
remote:  !     Failed to install gems via Bundler.
```

schneems:
@jaredbeck I've reported a similar issue but with different failure mode upstream to bundler. I'm tracking that work at https://github.com/heroku/heroku-buildpack-ruby/issues/1106 can you open up a new issue on bundler to report this failure mode? It might be related, or maybe not. Either way the developer experience around it from bundler could be improved. It's unclear from this message what the issue is or how to correct it unless you're keenly aware of the platform locking changes.

schneems:
> Sure. Our suggestion will be to improve the error message, perhaps referencing the bundler docs? 

I think me asking you to open the ticket is two part: awareness and documentation. Frequently bundler maintainers are not aware of the edgiest of edge cases of how bundler fails. Opening an issue assures they've at least had the opportunity to 👁️ it. Secondly if someone is searching that error message, this thread is related but tangential. I treat painful/unexpected experiences as UX bugs.

Docs could be part of mitigating this UX bug, but ultimately my ask would be some kind of behavior change to remediate this additional burden on X ruby devs across N of their apps. I've got ideas, but there's no consensus that the proposal would even be accepted. By ensuring all issues are in the tracker it allows us to have a better high-level view of the failure modes when designing usability fixes.

>  The docs are not bad

But the experience is still bad :(

It's frustrating to have this not work out of the box. For customers to do everything right, then get a failure on their first deploy. We're inserting an extra required step where there was previously none. 

- Bad: doing nothing, keeping this extra step in, and requiring all new apps to have a failed first deploy before they remember to re-lock their app.

I suggest one or more of the following:

- Good: Reducing newly-required steps from N to 1 (could be accomplished by allowing setting these values to be global instead of per-project).
- Better: ???
- Best: Reducing newly-required steps to 0 by doing the above and targeting x86 linux target by default.

Here's where my conversation ended with bundler https://github.com/rubygems/rubygems/issues/4269#issuecomment-763059395. 

I think upstreaming bundler to auto use "ruby" in the PLATFORMS seems like a low-hanging win (close https://github.com/rubygems/rubygems/issues/4280). 

Fixing the core issue will take deep changes as discussed in https://github.com/rubygems/rubygems/issues/4269#issuecomment-763059395. I think to get the behavior desired by adding platforms to the Gemfile.lock, we shouldn't be adding it for customers since it could possibly resolve different dependencies. So that leaves fixing the behavior upstream. I'm extremely swamped right now and while I would love to drop everything and work on trying to push my "best" strategy upstream, it's not something I'll be able to get to in the short term (also I have no forecast of when I would be able to). Not to mention they've not really indicated that such a solution would be accepted or merged.

PR #1158, opened 2021-05-24
Rename config.yml to cinfig.yml

anjim213:


edmorley:
This account attempted bitcoin mining in #1159. Reported to GitHub.

PR #1163, opened 2021-06-01
Remove support for Heroku-16

edmorley:
The Heroku-16 stack reached end-of-life on May 1st, 2021, and from June 1st, 2021, builds were disabled:
https://help.heroku.com/0S5P41DC/heroku-16-end-of-life-faq

This change:
* Removes Heroku-16 + Cedar-14 from the Ruby version presence check.
* Removes the now unused `libyaml` vendoring step, since it only ran on Heroku-16 and Cedar-14.
* Makes the buildpack release process no longer vendor the Heroku-16 Ruby archive.

Refs GUS-W-9329723.


schneems (approved):
Thanks!

Issue #1176, opened 2021-07-30
Detect `gems.rb` too.


ioquatix:
`gems.rb` should be a valid option here.

https://github.com/heroku/heroku-buildpack-ruby/blob/20b1919df498eff6010374b1ca1b91d9f7b56d2a/bin/detect#L14

ioquatix:
It looks like we might also need to update the logic here:

```
grep: /tmp/build_b93d19a1/Gemfile.lock: No such file or directory
-----> Compiling Ruby/NoLockfile
 !
 !     Gemfile.lock required. Please check it in.
 !
 !     Push rejected, failed to compile Ruby app.
 !     Push failed
```

schneems:
Someone put in a request to support a `BUNDLE_GEMFILE=<path>` env var. I think that would be the more general path forward.

ioquatix:
Maybe you can add that but you should then still auto detect `gems.rb` to match the behaviour of bundler.

PR #1177, opened 2021-07-30
Add support for `gems.rb`.

### LINKED ISSUE #534: "gems.rb" and "gems.locked" doesn't work

ioquatix:
Fixes #534 and #1176.

### DIFF
--- a/bin/detect
+++ b/bin/detect
@@ -14,6 +14,9 @@ fi
 if [ -f "$APP_DIR/Gemfile" ]; then
   echo "Ruby"
   exit 0
+elif [ -f "$APP_DIR/gems.rb" ]; then
+  echo "Ruby"
+  exit 0
 else
   echo "no"
   exit 100
--- a/lib/language_pack/no_lockfile.rb
+++ b/lib/language_pack/no_lockfile.rb
@@ -3,7 +3,8 @@ require "language_pack/ruby"
 
 class LanguagePack::NoLockfile < LanguagePack::Ruby
   def self.use?
-    !File.exists?("Gemfile.lock")
+    !File.exist?("Gemfile.lock") &&
+    !File.exist?("gems.locked")
   end
 
   def name
@@ -11,6 +12,6 @@ class LanguagePack::NoLockfile < LanguagePack::Ruby
   end
 
   def compile
-    error "Gemfile.lock required. Please check it in."
+    error "'Gemfile.lock' or 'gems.locked' required. Please check it in."
   end
 end


schneems:
Thanks for the PR. Sorry for the delay in the review. You may have seen that we're in the process of porting over code to the Cloud Native Buildpack spec here's a re-write in progress https://github.com/heroku/buildpacks-ruby. And that's actually been put on pause as we've decided to standardize on Rust as a buildpack language, so that progress needs to be ported over.

- **Opportunity**: I am -1 on requiring Gemfile to be moved to gems.rb. I'm +0 on the gems.rb interface (in addition to Gemfile). I'm not sure that I want to support this. I'm conducting a super unscientific and informal poll on twitter https://twitter.com/schneems/status/1432347740990697478. My sense is that the community is not behind gems.rb as an interface.
- **Implementation**: This won't work as-is as we manually parse the Gemfile.lock for internal feature detection. We would need to copy at least `gems.locked` to `Gemfile.lock` if we wanted to support this (which I'm not sure about, and leaning no).

A short-term workaround would be to create a "gems.rb" buildpack. It could move the files over to Gemfile and Gemfile.lock. Then `heroku/ruby` would run afterward and not require any changes.

The only area it might break is if for some reason the developer is expecting the `gems.rb` and `gems.locked` to be present on disk at those specific names.

ioquatix:
I agree that having two files is less than ideal.

I prefer `gems.rb` since the file extension matches the expectations of the language. That being said, `gems.rb` and `gems.locked` is inconsistent if you specify `BUNDLE_GEMFILE=foo.rb` it will produce `foo.rb.lock`:

image

It's super inconsistent.

Unfortunately, this was the situation created by bundler. I don't think public opinion matters that much, since this is how bundler works. Either you support it or you don't. In fact, I'm strongly of the opinion that you should be leaning on bundler to provide this meta-data e.g.

```
> bundler gemfile
gems.rb:gems.locked

> BUNDLE_GEMFILE=foo.rb bundle gemfile
foo.rb:foo.rb.lock
```

But as far as I know, this does not exist.

In any case, maybe I can just run with `BUNDLE_GEMFILE=gems.rb` and it would work?

edmorley:
The upstream migration to `gems.rb` was cancelled - if anything it seems projects should be removing support for it, not adding?

See:
https://github.com/ruby/setup-ruby/pull/84#issuecomment-686854903
https://github.com/ruby/setup-ruby/pull/84#issuecomment-687221101

edmorley (commented):
(Cancelling review to get this out of the languages team review queue)

ioquatix:
To clarify, `gems.rb` is a supported file name, and should be supported in Heroku, it's not deprecated or being removed.

edmorley:
@ioquatix Hi! This post would seem to suggest otherwise:
https://github.com/ruby/setup-ruby/pull/84#issuecomment-687221101

Is there some more recent official upstream documentation that confirms `gems.rb` is still officially supported?

edmorley:
Ah I missed the last sentence of that post, ignore me :-)

schneems:
Setting `BUNDLE_GEMFILE` does not currently work due to https://github.com/heroku/heroku-buildpack-ruby/issues/1348. It is supported in the CNB, but we've not backported to classic.

I will not be supporting `gems.rb` at this time. One way to get this feature would be a micro buildpack that renames `gems.rb` and the associated lockfile to what the `heroku/ruby` buildpack expects.

ioquatix:
I think the correct solution here is to defer to Bundler as the source of truth: https://github.com/Shopify/ruby-lsp/pull/1139 is an example of going from hard coded gemfile path => bundler.

However, I no longer have a vested interest in Heroku working out of the box with `gems.rb`.

PR #1179, opened 2021-08-02
Allow configuring `HATCHET_BUILDPACK_BASE`

deivid-rodriguez:
So that specs can also pass when run on forks.


schneems:
Thanks! Merged in #1182. Also fwiw. I'm starting work on a replacement buildpack over at github.com/heroku/buildpacks-ruby

PR #1180, opened 2021-08-02
[changelog skip] Allow `rspec --only-failures` to work by default

deivid-rodriguez:
It's useful when focusing on fixing a failing test run, since it only runs currently failing specs.


schneems:
Merged in https://github.com/heroku/heroku-buildpack-ruby/pull/1182

PR #1188, opened 2021-08-20
use WEB_CONCURRENCY.sh for WEB_CONCURRENCY default

dzuelke:
The Node.js, PHP and Python buildpacks write their WEB_CONCURRENCY defaults logic to this file name.

This ensures that the last buildpack to run defines the defaults an app will use.

If each buildpack were to use its own filename, then whatever file is evaluated first by the Shell on startup would set an initial value, and following files would see the variable already set and likely skip their calculation, or apply their own lower/upper limit enforcement, as it cannot know whether the variable was set by a user, or by some other buildpack's startup logic

(the Shell sources files using a glob pattern from .profile.d/ on startup, which on GNU Bash is in alnum order, see https://www.gnu.org/software/bash/manual/bash.html#Filename-Expansion)

This change uses the same filename as other buildpacks, ensuring that e.g. if heroku/nodejs is used as the first and heroku/ruby as the second buildpack, the defaults set by this buildpack will apply.

Right now, in this constellation, .profile.d/ruby.sh runs first, and sets a default value, then .profile.d/WEB_CONCURRENCY.sh runs later, and does nothing, as the defaults fit within Node.js's current logic - but if a user sets WEB_CONCURRENCY=0 manually, then Node.js' logic kicks in, and re-sets the value to 1, as Node.js doesn't allow a value of 0.

Ruby's Puma however allows "single mode", and for that, the value needs to be 0: https://github.com/puma/puma/issues/2393

Refs https://github.com/heroku/heroku-buildpack-nodejs/pull/926

GUS-W-9788823

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Main (unreleased)
 
-## v229 (?/??/2021)
-
 * Fix interoperability with other Heroku buildpacks' `$WEB_CONCURRENCY` handling (https://github.com/heroku/heroku-buildpack-ruby/pull/1188)
 
 ## v228 (6/24/2021)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Main (unreleased)
 
+## v229 (?/??/2021)
+
+* Fix interoperability with other Heroku buildpacks' `$WEB_CONCURRENCY` handling (https://github.com/heroku/heroku-buildpack-ruby/pull/1188)
+
 ## v228 (6/24/2021)
 
 * Bundler 2.x is now 2.2.21 (https://github.com/heroku/heroku-buildpack-ruby/pull/1170)
--- a/lib/language_pack/base.rb
+++ b/lib/language_pack/base.rb
@@ -206,11 +206,11 @@ private ##################################
   def setup_language_pack_environment
   end
 
-  def add_to_profiled(string)
+  def add_to_profiled(string, filename = "ruby.sh", mode = "a")
     profiled_path = @layer_dir ? "#{@layer_dir}/ruby/profile.d/" : "#{build_path}/.profile.d/"
 
     FileUtils.mkdir_p profiled_path
-    File.open("#{profiled_path}/ruby.sh", "a") do |file|
+    File.open("#{profiled_path}/#{filename}", mode) do |file|
       file.puts string
     end
   end
--- a/lib/language_pack/base.rb
+++ b/lib/language_pack/base.rb
@@ -206,7 +206,7 @@ private ##################################
   def setup_language_pack_environment
   end
 
-  def add_to_profiled(string, filename = "ruby.sh", mode = "a")
+  def add_to_profiled(string, filename: "ruby.sh", mode: "a")
     profiled_path = @layer_dir ? "#{@layer_dir}/ruby/profile.d/" : "#{build_path}/.profile.d/"
 
     FileUtils.mkdir_p profiled_path
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -412,7 +412,9 @@ EOF
       set_env_override "DISABLE_SPRING", "1"
 
       set_env_default "MALLOC_ARENA_MAX", "2"     if default_malloc_arena_max?
-      add_to_profiled (env("SENSIBLE_DEFAULTS") ? set_default_web_concurrency : ""), "WEB_CONCURRENCY.sh", "w" # always write that file, even if its empty (meaning no defaults apply), for interop with other buildpacks - and we overwrite that file rather than appending (which is the default)
+
+      web_concurrency = env("SENSIBLE_DEFAULTS") ? set_default_web_concurrency : ""
+      add_to_profiled(web_concurrency, filename: "WEB_CONCURRENCY.sh", mode: "w") # always write that file, even if its empty (meaning no defaults apply), for interop with other buildpacks - and we overwrite the file rather than appending (which is the default)
 
       # TODO handle JRUBY
       if ruby_version.jruby?
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -412,7 +412,7 @@ EOF
       set_env_override "DISABLE_SPRING", "1"
 
       set_env_default "MALLOC_ARENA_MAX", "2"     if default_malloc_arena_max?
-      add_to_profiled set_default_web_concurrency if env("SENSIBLE_DEFAULTS")
+      add_to_profiled (env("SENSIBLE_DEFAULTS") ? set_default_web_concurrency : ""), "WEB_CONCURRENCY.sh", "w" # always write that file, even if its empty (meaning no defaults apply), for interop with other buildpacks - and we overwrite that file rather than appending (which is the default)
 
       # TODO handle JRUBY
       if ruby_version.jruby?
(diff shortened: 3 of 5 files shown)


dzuelke:
(still needs tests, obviously)

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Main (unreleased)
 
+## v229 (?/??/2021)
```
schneems, line 3: For the changelog version and date I usually just have the entries under "unreleased" and then when I'm ready to release I'll add the version and put the date. It lets people know at a glance if what's on main is unreleased or not.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+* Fix interoperability with other Heroku buildpacks' `$WEB_CONCURRENCY` handling (https://github.com/heroku/heroku-buildpack-ruby/pull/1188)
+
```
schneems, line 6: Also I do think we need a devcenter changelog for this. I pre-write them and store them at this directory: https://github.com/heroku/heroku-buildpack-ruby/tree/main/changelogs. 

Here's an example changelog for this:

```
Ruby buildpack now clears previous default buildpack WEB_CONCURRENCY values

Some buildpacks wills set a default value of the WEB_CONCURRENCY environment variable. For applications using multiple buildpacks, the name of the `.profile.d` script dictates which value takes precedence rather than the order of the buildpack. This can cause problems with the WEB_CONCURRENCY value if a customer has both the `heroku/nodejs` buildpack and the `heroku/ruby` buildpack.

Now if the Ruby buildpack runs after the Node buildpack, it will remove the default value set by nodejs.

Details about the change in behavior can be found on the [pull request to the buildpack](https://github.com/heroku/heroku-buildpack-ruby/pull/1188).
```

lib/language_pack/base.rb
```diff
@@ -206,11 +206,11 @@ def log(*args)
   def setup_language_pack_environment
   end
 
-  def add_to_profiled(string)
+  def add_to_profiled(string, filename = "ruby.sh", mode = "a")
```
schneems, line 209: This is a clean refactor. I like this approach.

My strong preference to args is to use named args `filename: "ruby.sh"`. I was going to try to point at a "convention" list I'm keeping at https://github.com/heroku/buildpacks-ruby#coding-conventions but it looks like that didn't make the cut. I'll need to update that.

Can you switch over to named args?

lib/language_pack/ruby.rb
```diff
@@ -412,7 +412,7 @@ def setup_profiled(ruby_layer_path: , gem_layer_path: )
       set_env_override "DISABLE_SPRING", "1"
 
       set_env_default "MALLOC_ARENA_MAX", "2"     if default_malloc_arena_max?
-      add_to_profiled set_default_web_concurrency if env("SENSIBLE_DEFAULTS")
+      add_to_profiled (env("SENSIBLE_DEFAULTS") ? set_default_web_concurrency : ""), "WEB_CONCURRENCY.sh", "w" # always write that file, even if its empty (meaning no defaults apply), for interop with other buildpacks - and we overwrite that file rather than appending (which is the default)
```
schneems, line 415: With the additional arguments and the ternary it makes this quite the line. Since `env("SENSIBLE_DEFAULTS")` should return nil we could short-circuit it with `env("SENSIBLE_DEFAULTS") || ""` which is a little more idiomatic. I'm not sure how I would write that up in a convention though.

spec/hatchet/ruby_spec.rb
```diff
@@ -206,3 +206,24 @@
...
+      "heroku/nodejs",
+      :default
+    ]
+    before_deploy = -> { run!(%Q{echo "{}" > package.json}) }
+    Hatchet::Runner.new('default_ruby', stack: DEFAULT_STACK, buildpacks: buildpacks, before_deploy: before_deploy).deploy do |app|
+      expect(app.run("cat .profile.d/WEB_CONCURRENCY.sh").strip).to be_empty
+      expect(app.run("echo $WEB_CONCURRENCY").strip).to be_empty
+      expect(app.run("echo $WEB_CONCURRENCY", :heroku => {:env => "WEB_CONCURRENCY=0"}).strip).to eq("0")
+    end
+  end
```
schneems, line 222: Add a newline between `it` blocks please.

schneems (changes requested):
Thanks this is huge!

This is good to go as is, however, I requested some changes. Some of them I would like your feedback on, especially the devcenter changelog language.

Some of the requested changes are on style/convention. I think in general we should leave this to linters, but doing that work for this codebase is not ideal. Even still there's stuff linters don't catch. I'm starting to document some of my conventions (see the link). I would like feedback on that concept/approach.

lib/language_pack/ruby.rb
dzuelke: but we want `set_default_web_concurrency` if `SENSIBLE_DEFAULTS` is set, no?

I just realized this (new and old) check also doesn't cover `SENSIBLE_DEFAULTS=0`...

schneems: > but we want set_default_web_concurrency if SENSIBLE_DEFAULTS is set, no?

Ahh, right. Sorry got my wires crossed. I want to get rid of the ternary, it's already confused me in my own review 🤪. We've got some options then. 

We could put the logic to decide to return the script or an empty string inside of the poorly named `set_default_web_concurrency` method.

```ruby
def set_default_web_concurrency # probably worth renaming to something like "default_web_concurrency"
  return "" if !env("SENSIBLE_DEFAULTS")
  # ...

end
```

> I just realized this (new and old) check also doesn't cover SENSIBLE_DEFAULTS=0...

That's logically consistent with Ruby. Any value that's not empty is truthy. It's common practice to just check for env var presence (for boolean env vars) rather than the specific contents "disabled,false,0,wrong,off,bad,etc." for example https://github.com/codetriage/CodeTriage/blob/main/config/environments/production.rb#L27. It does mean we have to a bit more cautious with our interfaces like this one where a customer essentially can never opt out of a default value unless they do something like this in their Procfile:

```
web: unset WEB_CONCURRENCY && ...
```

This is one of the prime hesitations setting a default for ruby and why these actions were originally guarded behind the (also poorly named SENSIBLE_DEFAULTS env var).

lib/language_pack/base.rb
dzuelke: sure

lib/language_pack/ruby.rb
dzuelke: Yeah I know that about Ruby, it's just that most programs that read env vars like this will usually treat an unset variable the same as one set to "0"...

(that example, by the way, uses Rails' `present?`, which returns `false` for an empty or blank whitespace string, which is distinct from our current behavior, which evaluates to `true` in that case... `FOO= ruby -e 'puts !!ENV["FOO"]'`)

Anyway, it definitely doesn't belong into `set_default_web_concurrency`, that function should only be concerned with generating the shell code, and the calling code decides when to call it.

CHANGELOG.md
dzuelke: that's an explanation of the implementation, not of the behavior (which is what matters to customers); I'll write something up ;)

lib/language_pack/ruby.rb
dzuelke: alright check out the updated version... ternary is still in but moved to its own line

schneems: > it definitely doesn't belong into set_default_web_concurrency

This whole thing is very poorly architected in that it wasn't really architected at all. It's not really any less consistent to put the check in there than not. Moving to it's own line is good though. Let's ship it.

schneems (approved):

PR #1189, opened 2021-08-21
[changelog skip] drop sqlite3 test Ruby version requirement

dzuelke:
due to broken quoting, the test currently just echos a string to stdout rather than writing it to Gemfile as intended
    
however, the desired Ruby version isn't available on heroku-20, and the test clearly works without the requirement, so we can omit it altogether


Malax (approved):

schneems:
Great catch @dzuelke

PR #1190, opened 2021-08-21
move to new style changelog check

dzuelke:
[changelog skip] now uses either PR description or GitHub label, and the v2 checkout action

### DIFF
--- a/.github/workflows/check_changelog.yml
+++ b/.github/workflows/check_changelog.yml
@@ -1,16 +1,21 @@
 name: Check Changelog
 
 on:
- pull_request:
-  types: [opened, reopened, edited, synchronize]
+  pull_request:
+    types: [opened, reopened, edited, labeled, unlabeled, synchronize]
+
 jobs:
-  build:
-    if: |
-      !contains(github.event.pull_request.labels.*.name, 'dependencies') &&
-      !contains(github.event.pull_request.labels.*.name, 'automation')
+  check-changelog:
     runs-on: ubuntu-latest
+    if: |
+      !contains(github.event.pull_request.body, '[skip changelog]') &&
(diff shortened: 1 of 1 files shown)


edmorley (changes requested):
For most of the other repos, we've switched away from using the PR title for skipping the check (since it's pretty obtrusive to have it in the title, particularly when it ends up in the eventually merged commit message), and instead allowing a choice of either PR description or GitHub label. 

In addition, there are a number of other papercuts fixed (eg using native GitHub actions skipping feature for faster skipping and correct check status + listening to some additional event types) and compatibility added for `actions/checkout` v2+.

Would you mind switching this check to use the new style instead? eg:
https://github.com/heroku/heroku-buildpack-python/blob/main/.github/workflows/check_changelog.yml

.github/workflows/check_changelog.yml
```diff
@@ -1,16 +1,21 @@
...
-      !contains(github.event.pull_request.labels.*.name, 'dependencies') &&
-      !contains(github.event.pull_request.labels.*.name, 'automation')
+  check-changelog:
     runs-on: ubuntu-latest
+    if: |
+      !contains(github.event.pull_request.body, '[skip changelog]') &&
+      !contains(github.event.pull_request.body, '[changelog skip]') &&
+      !contains(github.event.pull_request.body, '[skip ci]') &&
+      !contains(github.event.pull_request.labels.*.name, 'skip changelog') &&
+      !contains(github.event.pull_request.labels.*.name, 'dependencies')
```
edmorley, line 15:
```suggestion
      !contains(github.event.pull_request.labels.*.name, 'dependencies') &&
      !contains(github.event.pull_request.labels.*.name, 'automation')
```

edmorley (approved):
Thank you! :-)

PR #1193, opened 2021-08-30
Increase default Ruby version to 2.7.4

schneems:
Our policy is to stay on the latest patch release of the middle supported Ruby version. Supported versions are 3.0, 2.7, and 2.6 so the latest of the 2.7 is 2.7.4.

GUS-W-9831879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## v229 (7/30/2021)
 
-* Default Ruby version is now 2.7.4 (https://github.com/heroku/heroku-buildpack-ruby/pull/974)
-
+* Default Ruby version is now 2.7.4 (https://github.com/heroku/heroku-buildpack-ruby/pull/1193)
 * Fix interoperability with other Heroku buildpacks' `$WEB_CONCURRENCY` handling (https://github.com/heroku/heroku-buildpack-ruby/pull/1188)
 
 ## v228 (6/24/2021)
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## v229 (7/30/2021)
 
+* Default Ruby version is now 2.7.4 (https://github.com/heroku/heroku-buildpack-ruby/pull/974)
+
```
edmorley, line 6:
```suggestion
* Default Ruby version is now 2.7.4 (https://github.com/heroku/heroku-buildpack-ruby/pull/1193)
```

edmorley (approved):
Woot! :-D

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## v229 (7/30/2021)
 
+* Default Ruby version is now 2.7.4 (https://github.com/heroku/heroku-buildpack-ruby/pull/1193)
```
edmorley, line 5: @schneems This changelog entry has been added under the existing `v229` release, rather than under the "unreleased" section - I've opened #1205 to fix it.

schneems: This happened on a rebase perhaps? If we put effort into productizing our changelog check it would be nice to optionally enforce:

- Entry must be in the "unreleased" section (with optional over-ride)
- Entry dates in headers must be added with the current date (with optional over-ride)
- Maybe more

Thanks for spotting this!

edmorley: I missed it in my review, so my bad!

PR #1205, opened 2021-09-29
Correct CHANGELOG entries for #1188 and #1193

edmorley:
* Moves the entry for #1193 to "unreleased", since it was not part of the v229 release.
* Fixes the date for the v229 release (it was published to the buildpack registry on `2021-08-30T15:57:32.024Z`).
* Turns the `Main` heading into an H2 so it's at the same heading level as the versioned headings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
-# Main (unreleased)
+# Changelog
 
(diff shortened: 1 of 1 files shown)


doriskwan (approved):

schneems:
Thanks!

PR #1212, opened 2021-10-13
[close #1202] Add support for Rails 7 & jsbundling

### LINKED ISSUE #1202: Yarn appears to only be available alongside Webpacker
I'm experimenting with a Rails 7.x alpha app on Heroku, which now provides an alternative to Webpacker call [`jsbundling-rails`](https://github.com/rails/jsbundling-rails). These bundlers effectively run externally of Rails and inject their builds into the regular old asset pipeline.

They still expect yarn to be available to install and build, but it appears as though `yarn` may not be available if Webpacker isn't detected.

```
       Running: rake assets:precompile
       sh: 1: yarn: not found
       sh: 1: yarn: not found
       Asset precompilation completed (1.52s)
       Cleaning assets
       Running: rake assets:clean
```

The easy solution is to add the heroku/nodejs buildpack and then it'll build fine. But if the ruby buildpack already includes support for yarn perhaps it should always be available, or better at detecting when it's required. Perhaps the presence of a `package.json` or `yarn.lock` should be a good enough hint to make it available?

schneems:
Rails 7 introduced the gem:

```
gem "jsbundling-rails", "~> 0.1.0"
```

Which no longer relies on `webpacker`. 

Issue #1202 was opened since developers using jsbundling would expect `yarn` to be available but won't have the `webpacker` gem which we use to gate detection for yarn installation.

With this change we now install nodejs when a `package.json` is detected and yarn when `yarn.lock` is detected.

### DIFF
--- a/changelogs/unreleased/jsbundling_packagejson.md
+++ b/changelogs/unreleased/jsbundling_packagejson.md
@@ -1,0 +1,10 @@
+## Nodejs will now be installed for Ruby applications that have a `package.json` file
+
+Ruby applications using the `heroku/ruby` buildpack now receive a default version of `node` installed if they have a `package.json` file in the root directory of their application.
+
+Prior to this change, only applications using the `webpacker` or `execjs` gem would trigger node installation logic. This change is intended to facilitate Rails 7 applications using `jsbundling-rails` without `webpacker`.
+
+>Note
+>Applications using the `heroku/nodejs` buildpack before the `heroku/ruby` buildpack will not see a change in behavior
+
+For more information, see [Heroku Ruby support documentation](https://devcenter.heroku.com/articles/ruby-support#installed-binaries).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Main (unreleased)
 
+* Applications with `package.json` now get `nodejs` installed (https://github.com/heroku/heroku-buildpack-ruby/pull/1212)
+* Applications with `yarn.lock` now get `yarn` installed (https://github.com/heroku/heroku-buildpack-ruby/pull/1212)
+
 ## v230 (10/5/2021)
 
 * Default Ruby version is now 2.7.4 (https://github.com/heroku/heroku-buildpack-ruby/pull/1193)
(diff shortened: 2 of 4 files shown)


edmorley:
If I'm reading the code correctly, both before and after this change, if an app uses `heroku/nodejs` and has that buildpack before the Ruby buildpack, then the Ruby buildpack will never install its own copy of Node.

However for apps that either:
1. Use `heroku/nodejs` and have it after `heroku/ruby`
2. Or, don't actually use the Node buildpack and just happen to have a `package.json` lying around (eg for another purpose, eg other tools)

...then the Ruby buildpack will start installing its own Node, when it didn't before - even if they aren't actually using Rails 7.

In the case of (1), this means there will be two Node.js versions installed. Will this cause any problems? PATH order? Slug size?

In the case of (2) only one Node.js version will be installed, but it may/may not be unused depending on whether they use this Rails 7 feature. For apps not using Rails 7 this seems like it will increase slug size unnecessarily?

You'll know whether the above is an issue or not, but wanted to double check these had been considered :-)

changelogs/unreleased/jsbundling_packagejson.md
```diff
@@ -0,0 +1,10 @@
+## Nodejs will now be installed for Ruby applications that have a `package.json` file
+
+Ruby applications using the `heroku/ruby` buildpack now receive a default version of `node` installed if they have a `package.json` file in the root directory of their application.
+
+Prior to this change, only applications using the `webpacker` or `execjs` gem would trigger node installation logic. This change is intended to facilitate Rails 7 applications using `jsbundling-rails` without `webpacker`.
+
+>Note
+>Applications using both the `heroku/nodejs` and the `heroku/ruby` buildpacks will not see a change in behavior
```
edmorley, line 8: ...(when the `heroku/nodejs` buildpack is prior to `heroku/ruby`)

dentarg:
> > Use heroku/nodejs and have it after heroku/ruby
> 
> That's extremely uncommon.

Why do you think that? 

Out of 9 apps we have using both `heroku/ruby` and `heroku/nodejs`, 8 have the nodejs buildpack after the ruby buildpack.

PR #1227, opened 2021-11-09
No-op Lxpc

schneems:


edmorley (commented):
Would a smaller/safer fix be to adjust this line?
https://github.com/heroku/heroku-buildpack-ruby/blob/fe4db44101a9482c09cadd6e5b5ebf112e278f79/lib/language_pack/instrument.rb#L28

The alternate codepath there is what is already used for the GitHub buildpack URLs.

schneems:
Great idea, moved it to https://github.com/heroku/heroku-buildpack-ruby/pull/1228

PR #1228, opened 2021-11-09
Make lxpc a no-op

schneems:


edmorley (approved):

thomas-holmes:
Thanks. At risk of pointing out the obvious (and I'm not really familiar with these codebases). We'll also need to make sure this is supported for our old version of ruby that Railgun uses.

PR #1229, opened 2021-11-11
Remove instrumentation

schneems:
Previously we used logplex to collect production performance statistics. It worked like this. We collect the stats, then stream them to a logplex app endpoint using l2met format. A logplex drain read from the data, serialized it into a Postgres database where we later wrote custom dashboards via SQL queries to find places to optimize.

That setup was ad-hoc. After we did the immediate perf work we decommissioned the drain and the Postgres database. We didn't remove the logplex integration or the instrumentation partially because we didn't know if we would need to do another round of performance analysis and partially because it was believed not removing it was "safer" (when this happened there were far fewer tests on the buildpack and developing on it was a far riskier endeavor).

We haven't used this in years and are in the process of re-writing the Ruby buildpack. Any future instrumentation and performance analysis will be done away from this codebase. We can safely remove the instrumentation, lpxc (used to send logs to the endpoint), dontenv (used to source the environment variable for logplex), and the logic around preserving and copying the `.env` file from prior builds.

#GUS-W-8216066


edmorley (approved):
🎉

Issue #1214, opened 2021-10-19
Why isn't this buildpack using LTS Node?


AlecRust:
I see in the Heroku [Node.js buildpack docs](https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version) it says:

> If a Node version isn’t specified in the engine, the `14.x` release will be used.

This is what I would expect, as `14.x` is the LTS at time of writing.

However this Ruby buildpack (which is the only one used in our app) uses Node 12.16.2. **I would have expected both the Ruby and the Node buildpacks to use LTS Node**.

It [looks like](https://devcenter.heroku.com/articles/ruby-support#installed-binaries) I could also load the Node buildpack in order to set a specific Node version, but I don't really want to set a Node version, I'm happy for LTS to be used.

Can this Ruby buildpack be updated to use the current LTS version to match the Node buildpack?

AlecRust:
Just a small bump on this, an ever growing list of npm packages are dropping support for 12.x. A user of this Ruby buildpack would expect to be able to install the latest version of common dependencies like `eslint`, `stylelint` etc.

AlecRust:
Here's a PR: https://github.com/heroku/heroku-buildpack-ruby/pull/1231

schneems:
This is deployed today

PR #1231, opened 2021-11-19
Update Node and Yarn default versions

AlecRust:
Updates Node to latest LTS and Yarn to latest version.

See https://github.com/heroku/heroku-buildpack-ruby/issues/1214 for details.


schneems:
Thanks for the PR. I'm pulling your change in via https://github.com/heroku/heroku-buildpack-ruby/pull/1237

PR #1242, opened 2021-12-15
Update Pack CLI to v0.23.0 in CI

edmorley:
Changes:
https://github.com/buildpacks/pack/releases

And switch to the official `buildpacks/pack` orb:
https://circleci.com/developer/orbs/orb/buildpacks/pack

GUS-W-10301655.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,13 @@
 version: 2.1
 orbs:
-  buildpacks: jkutner/buildpacks@0.1.0
+  pack: buildpacks/pack@0.2.4
   ruby: circleci/ruby@0.2.1
 jobs:
   build:
     machine: true
     steps:
-      - buildpacks/install-pack:
-          version: "0.11.0"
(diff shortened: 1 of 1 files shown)


edmorley:
@schneems I'm not sure why the Circle CI CNB tests are now failing. The CI run is hanging (and eventually timing out) after starting WEBrick. Unfortunately there is no pack CLI log output in the CI logs - it seems any pack command output is not being propagated.

It does seem the CNB tests here manually invoke pack:
https://github.com/heroku/heroku-buildpack-ruby/blob/9762663063ee3aab85bf85a594164bb3ad1485e2/spec/cnb/basic_local_pack_spec.rb#L15

...compared to those in the new repo that don't:
https://github.com/heroku/buildpacks-ruby/blob/main/spec/cnb/buildpack_spec.rb

I'm presuming this is a pack CLI arg difference or similar, but without the output it's hard to debug.

edmorley:
Pack is being updated as part of #1270.

PR #1249, opened 2021-12-16
Reintroduce PR limit

schneems:
When multiple PRs are opened concurrently multiple tests are fired off. These all compete for the same API limits. Hatchet has baked in rate limiter. When it runs out of API requests, it will slow down. If too many concurrent CI executions are kicked off at one time, all the builds slow one another down, and prevent any of them from completing.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,7 @@ version: 2
 updates:
   - package-ecosystem: "bundler"
     directory: "/"
+    open-pull-requests-limit: 1 # Limit concurrent CI runs from executing, do not remove
     schedule:
       interval: "weekly"
   - package-ecosystem: "github-actions"


edmorley (changes requested):
Can we try raising the limits instead? The raised limits for Python's CI user have been working fine (and by raised I mean 50x multiplier not 5x, plus a raised max apps count). I'll apply those same changes to Ruby now.

edmorley:
With raised limits these concurrent runs all passed:
Screenshot

schneems:
> Can we try raising the limits instead?

The last time I spoke to web services (years ago) they told me very firmly to not raise them higher than 5. It has been quite some time, but there was once an incident triggered due to buildpack tests in logplex as buildpack tests use services in ways that customers generally won't. It might be safer to raise the limit now, but I would want to ask them before doing it.

In general, I actually don't like these PRs to the Ruby buildpack. None of the dependencies we use in production and Ruby tends to have many patch releases in many components that just aren't that critical in dev. It ends up being a lot of toil. 

Even when we get X concurrent PRs we have to rebase X times so it doesn't really help that much due to the requirement that PRs be up to date.

I appreciate you staying on top of the PRs, but it's not a value add task to this repo (other than Hatchet and generally I do that manually as I'm typically dogfooding a fix). I'm happy to talk about this more.

edmorley (approved):

Issue #1250, opened 2021-12-16
formatting of database.yml changed in the last release!


benhutton:
Hey ya'll,

Check out https://github.com/heroku/heroku-buildpack-ruby/pull/1229/files#diff-d7f6f21bc5e80a7032506dae28ef835682a02ae9d2d9d162faccf793a826f1bcR944-R950. That change of indentation breaks the yaml in database.yml. Our Heroku CI runs are now failing.

![image](https://user-images.githubusercontent.com/61645/146439008-ce6afe39-f098-416f-951e-55e6b2babf3f.png)

schneems:
Thanks for opening an issue. 

I'm looking now. Do you have a support ticket as well? Can you give me the link? You're not on Rails 7 are you (there's an issue with Rails 7 and Heroku CI)? What is the error message are you getting?

Also fwiw you can now check-in database.yml as of Rails 4.1+ https://devcenter.heroku.com/changelog-items/709. Checking in a `config/database.yml` file will allow you to work around this issue in the short term.

latortuga:
I just spent the last 20 minutes identifying this issue as well. It prevents deployment for me because the app cannot boot - for folks googling out there, I get the following:

```
ActiveRecord::AdapterNotSpecified: staging database is not configured                                        ││                                   ││                                                                      │```

I am not on Rails 7 (actually Rails 2.3 LTS).

schneems:
This is deployed.

benhutton:
@schneems thanks for the quick fix! Sorry for my slow response.

I do not have a support ticket, nor am I on Rails 7. We're on a non-rails custom ruby stack. The error was very particular to our custom ruby code and how we were expecting a database.yml file to be.

PR #1251, opened 2021-12-16
Fix database.yml indentation

benhutton:
Fix for #1250


schneems:
Thanks! Running tests over in https://github.com/heroku/heroku-buildpack-ruby/pull/1252

Issue #1254, opened 2021-12-17
Could not determine schema/structure


fedegl:
I upgraded to Rails 7 and started getting these errors on the CI environment.

It seems to fail at loading the DB from dump and thus the tables don't exist when rails is initialized and crashed.

```-----> Clearing db:test:purge db:create db:structure:dump rake tasks
-----> Detecting rake tasks
       Could not determine schema/structure from `ActiveRecord::Base.schema_format`:
       [2021-12-17T20:01:11 (process) #668][ERROR] appsignal: Push API key not set after loading config
       [2021-12-17T20:01:11 (process) #668][ERROR] appsignal: Not starting, no valid config for this environment
             PaperTrail 12.1.0 is not compatible with ActiveRecord 7.0.0. We allow PT
             contributors to install incompatible versions of ActiveRecord, and this
             warning can be silenced with an environment variable, but this is a bad
             idea for normal use. Please install a compatible version of ActiveRecord
             instead (>= 5.2, < 7.0). Please see the discussion in paper_trail/compatibility.rb
             for details.
       /app/app/overrides/measured/units/volume_override.rb:3: warning: already initialized constant Measured::Volume
       /app/vendor/bundle/ruby/3.0.0/gems/measured-2.8.1/lib/measured/units/volume.rb:2: warning: previous definition of Volume was here
       /app/vendor/bundle/ruby/3.0.0/gems/activerecord-7.0.0/lib/active_record/connection_adapters/postgresql/database_statements.rb:19:in `exec': PG::UndefinedTable: ERROR:  relation "vendors" does not exist (ActiveRecord::StatementInvalid)
       LINE 9:  WHERE a.attrelid = '"vendors"'::regclass
                                   ^
       	from /app/vendor/bundle/ruby/3.0.0/gems/activerecord-7.0.0/lib/active_record/connection_adapters/postgresql/database_statements.rb:19:in `block (2 levels) in query'
       	from /app/vendor/bundle/ruby/3.0.0/gems/activesupport-7.0.0/lib/active_support/concurrency/share_lock.rb:187:in `yield_shares'
       	from /app/vendor/bundle/ruby/3.0.0/gems/activesupport-7.0.0/lib/active_support/dependencies/interlock.rb:41:in `permit_concurrent_loads'
       	from /app/vendor/bundle/ruby/3.0.0/gems/activerecord-7.0.0/lib/active_record/connection_adapters/postgresql/database_statements.rb:18:in `block in query'
[...]

PR #1257, opened 2022-01-03
Specialize Rails7 Database setup on Heroku CI

schneems:
## Part 1: Test Rails 7 on Heroku CI

Added a test for Rails 7 to exercise behavior in https://github.com/heroku/heroku-buildpack-ruby/issues/1254. Update the Rails 7 example app to use the released version instead of alpha. Also, I updated the app to be a Heroku CI compatible app (adds app.json):

- https://github.com/sharpstone/rails-jsbundling/commit/34bd25cef7c09758b3e7763a5d4be6dc94364606
- https://github.com/sharpstone/rails-jsbundling/commit/a98a4e41708b48a36a0adb941df4b7d28d25d5d5
- https://github.com/sharpstone/rails-jsbundling/commit/b4ec664cbe63dd9176884f746f64a46bed879c21
- https://github.com/sharpstone/rails-jsbundling/commit/1fb48b40bf506fbc3fb7c00a67e8f98dce82ac07

Add CI tests against Rails 7. Which was needed to test specialization

### Background

The Ruby buildpack sets up a Rails app's database for a customer. In Rails there are individual migrations which will add or remove a single table/column/etc. Then there is a "schema" that captures the final result of the whole process. Running individual migrations is fragile and prone to break it's also slow. To speed things up we first apply the saved schema, then we run the migrations.

Rails has two different formats for storing schema. There is a simplified "ruby" format that uses a DSL to describe the schema of the database. This is easy to read and understand but it doesn't support database specific calls (for example a GIN index in postgres cannot be applied to a Mysql database). So before we can apply a schema we have to determine which format of schema the developer wants us to apply.

Both schema formats can live in the app at the same time and they can be switched via a config flag in the app.

We handled that previously by calling these two tasks if they existed:

```
    schema_load    = rake.task("db:schema:load_if_ruby")
    structure_load = rake.task("db:structure:load_if_sql")
```

Otherwise we would fallback on calling:

```
rails runner 'puts ActiveRecord::Base.schema_format'
```

In Rails 6.1 the tasks used for database setup in our test pack were deprecated. They were removed in Rails 7 https://edgeguides.rubyonrails.org/7_0_release_notes.html. Existing Rails 7 apps still fallback to the `rails runner` method but it is slower and error prone see #1254.

### Fix

This PR moves to specialize the tasks for Rails7+ to use `db:schema:load` per deprecation instruction:

```
DEPRECATION WARNING: Using `bin/rails db:schema:load_if_ruby` is deprecated and will be removed in Rails 7.0. Configure the format using `config.active_record.schema_format = :ruby` to use `schema.rb` and run `bin/rails db:schema:load` instead. (called from <main> at /app/bin/rake:4)
```

That task will handle the logic for both determining the schema format and applying the schema.

GUS-W-10372684

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Main (unreleased)
 
+* Fix deprecated rake tasks for Rails 7 on Heroku CI (https://github.com/heroku/heroku-buildpack-ruby/pull/1257)
+
 ## v235 (2022/01/03)
 
 * Bundler 2.x is now 2.2.33 (https://github.com/heroku/heroku-buildpack-ruby/pull/1248)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Main (unreleased)
 
+## v236 (2022/01/04)
+
 * Fix deprecated rake tasks for Rails 7 on Heroku CI (https://github.com/heroku/heroku-buildpack-ruby/pull/1257)
 
 ## v235 (2022/01/03)
--- a/Rakefile
+++ b/Rakefile
@@ -95,6 +95,11 @@ namespace :buildpack do
 
       changelog_md.write(contents.gsub("## Main (unreleased)", new_section))
     end
+
+    version_rb = Pathname(__dir__).join("lib/language_pack/version.rb")
+    puts "Updating version.rb"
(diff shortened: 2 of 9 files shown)


schneems:
Related internal ticket number: 1064070

PR #1313, opened 2022-06-13
Adjust curl retry and connection timeout handling

edmorley:
In the shimmed CNBs used in `heroku/builder` we have been seeing quite a few transient errors related to buildpacks downloading from S3.

Adding appropriate retries and connection timeouts to all of our buildpack's curl usages should help with these, as well as make builds more reliable in general for users on Heroku, plus also anyone using a shimmed CNB locally with Pack CLI (where the network connection may be even less reliable).

The `--retry-connrefused` option has been used since otherwise curl doesn't retry cases where the connection was refused. Ideally we would use `--retry-all-errors` which takes that one step further, however that option was only added in curl 7.71, so is only supported by Heroku-22+.

For the connection timeouts, I've ensured they are configurable via `CURL_CONNECT_TIMEOUT`, like elsewhere in the buildpack.

Note: Whilst `curl_retry_on_18` does retries, this is only for exit code 18 (partial downloads, which can't be handled by curl's own retry feature), so retries via curl itself are needed too, and it's fine to have this duplication (see also https://github.com/heroku/heroku-buildpack-php/commit/6936331d25a87d14f6dc701c74703da8eeb40247).

For more on curl options, see:
https://curl.se/docs/manpage.html

GUS-W-11283397.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Main (unreleased)
 
+* Adjust curl retry and connection timeout handling (https://github.com/heroku/heroku-buildpack-ruby/pull/1312)
+
 ## v242 (2022/06/07)
 
 * Ensure `bin/release` exits zero if `tmp/heroku-buildpack-release-step.yml` does not exist (https://github.com/heroku/heroku-buildpack-ruby/pull/1309)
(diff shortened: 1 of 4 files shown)


edmorley:
Example Ruby failure we've seen:
https://app.circleci.com/pipelines/github/heroku/builder/1887/workflows/d19f629b-2692-4732-9a3b-5b6cd9d6811f/jobs/39866

PR #1315, opened 2022-07-21
Procfile

2rba:
Add Procfile support


dzuelke:
This is already the platform behavior, @2rba. If no `Procfile` exists, the processes from https://devcenter.heroku.com/articles/buildpack-api#bin-release are taken.

2rba:
@dzuelke thank you for the comment.
I am interested in Procfile support for cloud native buildpacks. Any advice on how it can be achieved?

PR #1326, opened 2022-10-31
Remove circle tests

schneems:
GUS-W-11832774


edmorley (commented):
What are your thoughts about also deleting this no-longer-run CNB spec?
https://github.com/heroku/heroku-buildpack-ruby/blob/main/spec/cnb/basic_local_pack_spec.rb

I know the buildpack still has basic CNB support, but leaving this CNB test around without running it means it'll get out of date/break anyway, plus potentially cause confusion for people looking at the repo working out what tests are actually used/run. Should you ever need to refer to the CNB tests they'll still exist in the Git log.

If that spec was deleted, the following can also be cleaned up:
- the `--exclude-pattern` here:
https://github.com/heroku/heroku-buildpack-ruby/blob/83da2a7651f97b5354900144a57d6c5c53862269/app.json#L23
- the cutlass usages in `spec/spec_helper.rb`
- the cutlass dependency can be dropped

edmorley:
The existing Heroku CI tests are also failing:
https://dashboard.heroku.com/pipelines/ac057663-170b-4bdd-99d0-87560eb3a570/tests/1728

schneems:
@edmorley updated, tests are now passing. They were failing due to the stack image default change.

spec/cnb/basic_local_pack_spec.rb
```diff
@@ -1,95 +0,0 @@
-require_relative '../spec_helper'
-
```
edmorley, line 2: With this spec removed, the following are unused and can also be removed:
- the cutlass usages in `spec/spec_helper.rb`
- the cutlass dependency can be dropped

(Mentioned in the last review: https://github.com/heroku/heroku-buildpack-ruby/pull/1326#pullrequestreview-1162175333)

edmorley (changes requested):

edmorley (approved):
Looks great - thank you!

PR #1331, opened 2022-11-08
Upgrade default node version to 18.12.1

schneems:


joshwlewis (commented):
Anyone affected by this default change that is also running `heroku-18` will probably see this error: https://kb.heroku.com/R7DTSTD0/fix-node-js-error-version-glibc_2-28-not-found. It seems fairly likely that folks will fall into this scenario. If we're not going to conditionally choose the default (which I think we should consider), should we at least mention the heroku-18 + Node.js 18 incompatibility in the changelog?

joshwlewis:
Closing in favor of #1342.

PR #1342, opened 2022-11-29
Bump default Node.js and Yarn versions

joshwlewis:
In #1331, the Node.js version change to 18.x is likely to cause issues for folks running on `heroku-18` (as mentioned in https://github.com/heroku/heroku-buildpack-ruby/pull/1331#pullrequestreview-1174488149). This is an alternative update, which should bring both the default `yarn` and `node` installations to the most recent patch without negatively impacting users on `heroku-18`. Node.js 16 is a maintenance LTS, so should be fine to use for a bit longer. Longer term, we may want to conditionally select Node.js 18 for folks on `heroku-20`+.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ## v246 (2022/11/29)
 
-* Default Node.js version now 16.18.11
-* Default Yarn version now 1.22.19
+* Default Node.js version now 16.18.11 (https://github.com/heroku/heroku-buildpack-ruby/pull/1342)
+* Default Yarn version now 1.22.19 (https://github.com/heroku/heroku-buildpack-ruby/pull/1342)
 
 ## v245 (2022/11/16)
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Main (unreleased)
 
+## v246 (2022/11/29)
+
+* Default Node.js version now 16.18.11 (https://github.com/heroku/heroku-buildpack-ruby/pull/1342)
```
edmorley, line 7:
```suggestion
* Default Node.js version now 16.18.1 (https://github.com/heroku/heroku-buildpack-ruby/pull/1342)
```

edmorley (approved):
Thank you!

colincasey (approved):

sgerrand:
Suggest that you update the Heroku developer documentation to reflect this change: https://devcenter.heroku.com/articles/ruby-support#installed-binaries

PR #1366, opened 2023-03-02
Add a scheduled Hatchet app cleaner workflow

edmorley:
As of Hatchet v8.x, Hatchet now:
- Removes the apps it creates after each test runs, rather than only enabling maintenance mode. This occurs even if an individual test fails.
- Cleans up older apps iff it runs into the app limit. (This is only done when the app limit is hit, in order to prevent hundreds of unnecessary API requests, when test concurrency is high.)

However, leftover apps can still occur in the following cases:
- The GitHub Actions run is cancelled, or aborts in a non-clean way.
- If a Hatchet bug (or a mis-use of Hatchet by the tests) cause Hatchet to exit non-cleanly.
- If there is a platform incident causing the Heroku API to not be available.

Therefore, this scheduled cleanup job has been added to catch any such leftover apps. It runs daily at 6am UTC, and can also be triggered manually via the GitHub dashboard.

The cleanup job only deletes apps older than 10 minutes, so that it doesn't break any in progress CI jobs.

GUS-W-12629391.

### DIFF
--- a/.github/workflows/hatchet_app_cleaner.yml
+++ b/.github/workflows/hatchet_app_cleaner.yml
@@ -1,0 +1,31 @@
+name: Hatchet app cleaner
+
+on:
+  schedule:
+    # Daily at 6am UTC.
+    - cron: "0 6 * * *"
+  # Allow the workflow to be manually triggered too.
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  hatchet-app-cleaner:
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


dzuelke:
This repo isn't on Hatchet 8.0 yet, @edmorley !

edmorley:
@dzuelke I'm aware, I wasn't going to merge this until #1365 is merged.

edmorley:
This repo is now using Hatchet 8.x as of #1367.

Malax (approved):

PR #1371, opened 2023-03-07
Relax Ruby version constraint in Gemfile

edmorley:
So that the Hatchet app cleaner can use latest Ruby 3.1.

Resolves:

```
Your Ruby version is 3.1.3, but your Gemfile specified 3.1.2
```

From:
https://github.com/heroku/heroku-buildpack-ruby/actions/runs/4355145552/jobs/7611367502

Follow-up to #1366 and #1368.

GUS-W-12629391.


schneems (approved):
Thanks!

PR #1405, opened 2023-11-18
feat: allows to skip node installation

pftg:
In order to use Bun with Rails, will be great to avoid redundant Node and Yarn installations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Main (unreleased)
 
+- Apps with the environment variable `HEROKU_SKIP_NODE_INSTALL=1` set will no longer install auto-install nodejs and yarn.
+
 ## v262 (2023/11/08)
 
 - Warn when relying on default Node.js or Yarn versions (https://github.com/heroku/heroku-buildpack-ruby/pull/1401)
--- a/lib/language_pack/ruby.rb
+++ b/lib/language_pack/ruby.rb
@@ -1013,7 +1013,7 @@ params = CGI.parse(uri.query || "")
   # @note execjs will blow up if no JS RUNTIME is detected and is loaded.
   # @return [Array] the node.js binary path if we need it or an empty Array
   def add_node_js_binary
-    return [] if node_js_preinstalled?
+    return [] if node_js_preinstalled? || env("HEROKU_SKIP_NODE_INSTALL")
 
     if Pathname(build_path).join("package.json").exist? ||
          bundler.has_gem?('execjs') ||
@@ -1045,7 +1045,7 @@ params = CGI.parse(uri.query || "")
   end
 
   def add_yarn_binary
-    return [] if yarn_preinstalled?
+    return [] if yarn_preinstalled? || env("HEROKU_SKIP_NODE_INSTALL")
 
     if Pathname(build_path).join("yarn.lock").exist? || bundler.has_gem?('webpacker')
 


pftg:
@edmorley @schneems hey guys, could you review this PR

nertzy:
I have a similar use case. We pre-build our assets as part of our CI pipeline and then push with `heroku builds:create`. This buildpack then still insists on installing a Node runtime that is not used as part of the Heroku build process.

schneems:
> In order to use Bun with Rails, will be great to avoid redundant Node and Yarn installations.

I'm unfamiliar with bun, looking now. I found https://medium.com/nyc-ruby-on-rails/what-is-bun-d343601fed77 which is paywalled. I found https://bun.sh/. It looks like it's an alternative Node runtime like Deno which is written in Rust https://github.com/denoland/deno. I'm not clear on how it gets installed into a Ruby app or why.

I found https://github.com/rails/rails/pull/49241 linked from Ruby weekly. 

You can create a new app like this:

```
$ mkdir -p /tmp/ecdb467647e9cb8b6dad2775d6eda59f; cd /tmp/ecdb467647e9cb8b6dad2775d6eda59f
⛄️ 3.1.4 🚀 /tmp/ecdb467647e9cb8b6dad2775d6eda59f
⛄️ 3.1.4 🚀 /tmp/ecdb467647e9cb8b6dad2775d6eda59f
$ rails new my_app  --javascript bun --css tailwind --database sqlite3
```

But then I'm unable to get it to build assets:

```
$ rake assets:precompile
yarn install v1.22.19
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.04s.
yarn run v1.22.19
error Command "build" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
rake aborted!
jsbundling-rails: Command build failed, ensure `yarn build` runs without errors
/Users/rschneeman/.gem/ruby/3.1.4/gems/jsbundling-rails-1.2.2/lib/tasks/jsbundling/build.rake:14:in `block (2 levels) in <main>'
/Users/rschneeman/.gem/ruby/3.1.4/gems/rake-13.1.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => assets:precompile => javascript:build
(See full trace by running task with --trace)
```

Here's the full rails output https://gist.github.com/schneems/33ae9ace6b120e363a107ef9a8014b54

I'm unclear on how bun is installed on Heroku or how it's invoked. If it's a node runtime then `which node` should return a zero exit code indicating it's already installed and doesn't need to be reinstalled. If you don't need yarn installed you can delete the `yarn.lock` file and we won't install it, or alternatively provide a `bin/yarn` executable which will return zero on `which yarn`.

It looks like via https://bun.sh/docs/quickstart it doesn't give you a node shim but gives you a new command `bun`. you could create your own `bin/node` that redirects arguments to `bun` and that would also short-circuit our installing node logic.

I found https://github.com/oven-sh/bun/discussions/559 but there's not a single recommended way to install it on Heroku.

It's worth noting that in the future we'll be moving away from installing node and yarn and leaving that up to the node buildpack. This is in our CNB, which cannot be used right now https://github.com/heroku/buildpacks-ruby.

schneems:
I don't think I'll be merging this as it's not a feature I'll be able to forward port to the CNB and there are already mechanisms for avoiding installing those binaries. I think that `which node` and `which yarn` along with `package.json` and `yarn.lock` files are strong interfaces that we should rely on instead.

If you need help getting bun to work on a Heroku app, please open a support ticket and I can take a look. I can't write a custom buildpack, but I know a lot about buildpacks and could help someone who is interested.

pftg:
Hey @schneems, thanks for the review.

I do not have any issue installing bun, I have a build pack for this https://github.com/jetthoughts/heroku-buildpack-bun

> It looks like via https://bun.sh/docs/quickstart it doesn't give you a node shim but gives you a new command bun. you could create your own bin/node that redirects arguments to bun and that would also short-circuit our installing node logic.

yep, one of the options

Overall, there was no problem installing the bun; the problem was to prevent the installation of nodes and yarn. I'll check your options with stubbing `node`. Based on the code `run("which node").strip`, it should work.

PR #1406, opened 2023-11-29
Allow custom vendored Ruby URL

### LINKED ISSUE #1164: BUILDPACK_VENDOR_URL doesn't allow for using vendored Ruby since v201
A long time ago we forked this Ruby buildpack and recently I've been looking at going back to using upstream. One of the changes I've noticed is that prior to v227 it was possible to specify `BUILDPACK_VENDOR_URL` and this is where Ruby binaries would be fetched from. Since the extraction of the Ruby binary downloading out into bash in specifically https://github.com/heroku/heroku-buildpack-ruby/pull/886 the `BUILDPACK_VENDOR_URL` env var is ignored and binaries are always fetched from `https://s3-external-1.amazonaws.com/heroku-buildpack-ruby`. 

This feature to allow getting binaries from different locations was initially added way back in https://github.com/heroku/heroku-buildpack-ruby/pull/238 and has proliferated into other Heroku buildpacks since then e.g. Python used it, but recently swapped to use `BUILDPACK_S3_BASE_URL` in https://github.com/heroku/heroku-buildpack-python/pull/1085). The lack of support for it now for Ruby binaries seems at odds with the fact VENDOR_URL is still being used elsewhere in the buildpack today (e.g. plugins)

Is this something that you'd be interested in adding support for with the new downloader system? Without that feature we're going to be forced from the off to keep our fork, regardless of anything else that might crop up 😅

mlarraz:
Lets us provide our own custom set of vendored Ruby runtimes.

Useful for things like building Ruby on different architectures than what Heroku supports, or with different compile flags.

Currently in use on our fork of this repo, but we'd like to merge it upstream.

Fixes #1164.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Main (unreleased)
 
-- Fix BUILDPACK_VENDOR_URL support (https://github.com/heroku/heroku-buildpack-ruby/pull/1409)
+- Fix BUILDPACK_VENDOR_URL support (https://github.com/heroku/heroku-buildpack-ruby/pull/1406)
 
 ## v262 (2023/11/08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Main (unreleased)
 
+- Fix BUILDPACK_VENDOR_URL support (https://github.com/heroku/heroku-buildpack-ruby/pull/1409)
+
 ## v262 (2023/11/08)
 
 - Warn when relying on default Node.js or Yarn versions (https://github.com/heroku/heroku-buildpack-ruby/pull/1401)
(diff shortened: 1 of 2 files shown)


schneems:
I think this would close https://github.com/heroku/heroku-buildpack-ruby/issues/1164. Thanks for working on it!

schneems (approved):

edmorley:
Rather then review/merge #1409, lets merge this PR to make the credit/attribution clearer. We still need the separate non-forked PR for integration tests to run (since they use credentials not available to PRs from forks), however, we can take advantage of the fact that GitHub's status checks API associates a pass/fail result with a commit SHA256 and not a specific branch/remote, to let the successful CI test result from that PR carry across to this PR.

We needed one more change, which was the changelog addition, so I've pushed that to this PR from the other branch (which is possible since this PR has the `Maintainers are allowed to edit this pull request` option enabled).

I'll manually merge this PR now, since auto-merge was enabled on #1409, so if I had reviewed that PR it would have been auto-merged anyway.

edmorley (approved):

PR #1410, opened 2023-12-06
Allow custom vendored Node URL

mlarraz:
Follow-up to https://github.com/heroku/heroku-buildpack-ruby/pull/1406

Lets us ship custom builds, targeting different platforms than the Heroku default.

This is slightly more complicated than the Ruby change, as the URL contains a lot of platform-specific information, as well as interpolates the Node version.

My solution is to allow for a variable in the URL that gets substituted with the version.

This allows for dynamic lookups just like normal, and avoids hardcoding to a single version.

### DIFF
--- a/lib/language_pack/helpers/node_installer.rb
+++ b/lib/language_pack/helpers/node_installer.rb
@@ -18,7 +18,11 @@ class LanguagePack::Helpers::NodeInstaller
     # clean up un-used files. Instead we untar in a temp directory which
     # helps us avoid accidentally deleting code out of the user's slug by mistake.
     Dir.mktmpdir do |dir|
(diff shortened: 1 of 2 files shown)


edmorley (commented):
(Switching code review request from the Languages team alias to the Ruby language owner)

schneems:
The best way to install a custom node binary would be to make a micro buildpack https://devcenter.heroku.com/articles/buildpack-api. If `which node` returns zero it indicates to the Ruby buildpack that we do not need to install a node version. In the future we are moving away from installing node with CNBs (not available on Heroku yet) and instead relying on `heroku/node` buildpack for that functionality https://github.com/heroku/buildpacks-ruby.

PR #1421, opened 2024-02-20
v266 and update the release scripts

schneems:
The changelog format was updated, this change updates the deploy rake scripts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+## [v266] - 2024/02/20
+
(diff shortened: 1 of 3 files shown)


edmorley (changes requested):
Could you try using the new "prepare release" workflow to perform the release? It can be used self-serve (doesn't need another languages team member to review), and already supports the new changelog format:
https://github.com/heroku/heroku-buildpack-ruby/actions/workflows/prepare-release.yml

We should remove the old method from `Rakefile` (I didn't know it was there, or I would have removed it as part of #1417)

schneems:
- https://github.com/heroku/heroku-buildpack-ruby/pull/1423
- https://github.com/heroku/heroku-buildpack-ruby/pull/1424

PR #1423, opened 2024-02-20
Fix prepare release script

schneems:
There's no OLD_VERSION it is EXISTING_VERSION https://github.com/heroku/languages-github-actions?tab=readme-ov-file#inputs-2.


edmorley (approved):
Good spot, sorry for my mistake!

PR #1424, opened 2024-02-20
Clean up release automation

schneems:
The scripts in .github/scripts were for CNB releases which have been moved over to https://github.com/heroku/buildpacks-ruby and have not been used for some time.

The rake task to prepare releases has been moved to a GitHub action. The deploy task is still used, it's updated to work with the new standard CHANGELOG format.


Rakefile
```diff
@@ -68,64 +68,12 @@ end
...
-
-    command = "heroku buildpacks:publish heroku/ruby #{deploy.next_version}"
-    puts "Releasing to heroku: `#{command}`"
-    exec(command)
-  end
-  desc "stage a tarball of the buildpack, this runs on github actions to deploy CNB"
-  task :tarball do
-    tarballer = Tarballer.new(name: "heroku-buildpack-ruby", directory: __dir__)
-    tarballer.call
+    puts("Use https://github.com/heroku/heroku-buildpack-ruby/actions/workflows/prepare-release.yml")
```
edmorley, line 76: Both the `release` and `tarball` tasks have been deleted here, but the `release` task is presumably still used, since it's what publishes to the buildpack registry? (The GitHub Actions workflow only prepares the release)

edmorley (changes requested):

PR #1426, opened 2024-02-21
Update bundler to 2.5.6

schneems:
Close #1408

Issue #1408, opened 2023-12-02
Support Bundler 2.4.20+ to enable `ruby file: '.ruby-version'` in Gemfile


eliotsykes:
It'd be handy for the Heroku Ruby buildpack to enable using the `ruby file: ...` option in `Gemfile` by supporting Bundler 2.4.20.

This was introduced in Bundler 2.4.19:
```rb
ruby file: '.ruby-version'
```

Bundler 2.4.20 improved on this to support more formats, including `.tool-versions`:
```rb
ruby file: '.tool-versions'
```

Changelog entry: https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#2420-september-27-2023

dentarg:
`ruby File.read(".ruby-version")` works well (if you didn't know)

eliotsykes:
Thanks @dentarg 👍. Yes, I've used and liked that technique for some time too. 

It works cleanly for `.ruby-version`, [not so well for `.tool-versions` which requires a bit more work](https://github.com/rubygems/rubygems/issues/6895#issuecomment-1684400451):

```rb
source "https://rubygems.org" ruby(/^ruby\s(.*)$/.match(File.read(".tool-versions"))[1].strip)
```

schneems:
Feature requests are the number 1 reason (after bug fixes) for us to rev the bundler version. So thanks for letting me know you want it!

As an FYI In the future https://github.com/heroku/buildpacks-ruby will eventually replace this buildpack (a long way off). It pulls the bundler version from the Gemfile.lock instead of using this "blessed" system, which I would like to move away from.

Timing-wise, I'm unsure when I'll be able to get this out. I'm mindful that a lot of companies have holidays coming up and a 1 in a million bug for bundler shows up pretty often for us (so I want to balance out stability, even if there's no known cause for concern). 

To recap: Yes, I'll upgrade for sure. No commitment on the timeline, though.

janklimo:
+1 for this. We recently switched to the new DSL in Gemfile but had to revert the change until the buildpack can support it. Thanks for looking into it @schneems!

dentarg:
@schneems you might want to get to this before the next Rails release? If that release will include https://github.com/rails/rails/pull/49360

dentarg:
Submitted ticket #1349048 to Heroku support about this

janklimo:
[...]

PR #1427, opened 2024-02-21
Update bundler to 2.4.22

schneems:
Prior https://github.com/heroku/heroku-buildpack-ruby/pull/1426
Closing https://github.com/heroku/heroku-buildpack-ruby/issues/1408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Update Bundler 2 apps to version 2.4.22 (https://github.com/heroku/heroku-buildpack-ruby/pull/1427)
 
 ## [v266] - 2024-02-20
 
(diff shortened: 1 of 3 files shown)


schneems:
Breaking change (that seems undocumented, so I'm not sure if it's intentional or not).

```
$ bundle _2.3.27_ platform --ruby
ruby 3.0.0
$ bundle _2.4.22_ platform --ruby
No ruby version specified
```

Also 2.4.x dropped support for Ruby 2.5: https://github.com/rubygems/rubygems/blob/3fbb8bb93cf54b14374825859541a522c43a8377/bundler/CHANGELOG.md#breaking-changes-1

I need to investigate if it's a policy change and simply untested or if it hard breaks (uses newer APIs etc.) I'm guessing it uses newer APIs.

I would like to not break my customers on Ruby 2.5. So I'll need to think on another strategy. Possibly multiple versions of "blessed" bundler versions like 2.2, 2.3, 2.4, 2.5 etc.

Again, in the long term I want to get out of this managed bundler version game, but I don't have that luxury in the short term.

PR #1428, opened 2024-02-22
Bundler versions 2.4.22 and 2.5.6 are now available for Ruby Applications

schneems:
Prior https://github.com/heroku/heroku-buildpack-ruby/pull/1427
Closing https://github.com/heroku/heroku-buildpack-ruby/issues/1408

The [Ruby Buildpack](https://devcenter.heroku.com/articles/ruby-support#libraries) now installs a version of bundler based on the major and minor version listed in the `Gemfile.lock` under the `BUNDLED WITH` key. Previously, it only used the major version. Now, this logic will be used:

- `BUNDLED WITH` 1.x will receive bundler `1.17.3`
- `BUNDLED WITH` 2.0.x to 2.3.x will receive bundler `2.3.25`
- `BUNDLED WITH` 2.4.x will receive bundler `2.4.22`
- `BUNDLED WITH` 2.5.x and above will receive bundler `2.5.6`

It is strongly recommended that you have both a `RUBY VERSION` and `BUNDLED WITH` version listed in your `Gemfile.lock`. If you do not have those values, you can generate them and commit them to git:

```
$ bundle update --ruby
$ git add Gemfile.lock
$ git commit -m "Update Gemfile.lock"
```

Applications without these values specified in the `Gemfile.lock` may break unexpectedly when the defaults change.

### DIFF
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -42,9 +42,9 @@ class LanguagePack::Helpers::BundlerWrapper
   BLESSED_BUNDLER_VERSIONS["2.4"] = "2.4.22"
   BLESSED_BUNDLER_VERSIONS["2.5"] = "2.5.6"
   BLESSED_BUNDLER_VERSIONS.default_proc = Proc.new do |hash, key|
-    if key.start_with?("1")
+    if Gem::Version.new(key).segments.first == 1
       hash["1"]
-    elsif key.start_with?("2")
+    elsif Gem::Version::new(key).segments.first == 2
       if Gem::Version.new(key) > Gem::Version.new("2.5")
         hash["2.5"]
       elsif Gem::Version.new(key) < Gem::Version.new("2.3")
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -37,8 +37,38 @@ class LanguagePack::Helpers::BundlerWrapper
 
   BLESSED_BUNDLER_VERSIONS = {}
   BLESSED_BUNDLER_VERSIONS["1"] = "1.17.3"
-  BLESSED_BUNDLER_VERSIONS["2"] = "2.3.25"
-  BUNDLED_WITH_REGEX = /^BUNDLED WITH$(\r?\n)   (?<major>\d+)\.\d+\.\d+/m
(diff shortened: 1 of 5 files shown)


lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -37,8 +37,38 @@ class LanguagePack::Helpers::BundlerWrapper
...
   BLESSED_BUNDLER_VERSIONS = {}
   BLESSED_BUNDLER_VERSIONS["1"] = "1.17.3"
-  BLESSED_BUNDLER_VERSIONS["2"] = "2.3.25"
-  BUNDLED_WITH_REGEX = /^BUNDLED WITH$(\r?\n)   (?<major>\d+)\.\d+\.\d+/m
+  # Heroku-20's oldest Ruby verison is 2.5.x which doesn't work with bundler 2.4
+  BLESSED_BUNDLER_VERSIONS["2.3"] = "2.3.25"
+  BLESSED_BUNDLER_VERSIONS["2.4"] = "2.4.22"
+  BLESSED_BUNDLER_VERSIONS["2.5"] = "2.5.6"
+  BLESSED_BUNDLER_VERSIONS.default_proc = Proc.new do |hash, key|
+    if key.start_with?("1")
```
joshwlewis, line 45: This would also catch `10` or `100`. I'm guessing you don't want that.

```suggestion
    if key.start_with?("1.")
```

Or I guess you could use the version parser to extract the major version to protect from other weird scenarios:

```suggestion
    if Gem::Version.new(key).segments.first == 1
```

lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -37,8 +37,38 @@ class LanguagePack::Helpers::BundlerWrapper
...
+      hash["1"]
+    elsif key.start_with?("2")
```
joshwlewis, line 47: Same comment as above. This would pass for `20.1`, `200.1`, etc.

```suggestion
    elsif key.start_with?("2.")
```
or
```suggestion
    elsif Gem::Version::new(key).segments.first == 2
```

joshwlewis (approved):

lib/language_pack/helpers/bundler_wrapper.rb
schneems: Great catch. Thanks

PR #1436, opened 2024-03-21
Use cmd variable

JuanVqz:
The `cmd` variable was not used, so, openning this PR to use it.

### DIFF
--- a/lib/language_pack/helpers/rake_runner.rb
+++ b/lib/language_pack/helpers/rake_runner.rb
@@ -49,9 +49,9 @@ class LanguagePack::Helpers::RakeRunner
         cmd = "rake #{task}"
 
         if quiet_option
-          self.output = run("rake #{task}", options)
+          self.output = run(cmd, options)
         else
-          self.output = pipe("rake #{task}", options)
+          self.output = pipe(cmd, options)
         end
       end
       self.time = time


schneems:
Thank's for the PR!

The value of that variable is that any if any changes are made in one branch they're reflected in the other. To say: It was done intentionally.

## PRs to this repo

In general it's hard to accept patches on this project because the test infrastructure requires secret tokens so any PRs I have to pull down manually and then merge into a branch within the project. It's a large burden which we do occasionally from time to time, but I won't be doing that in this case.

## PRs in general

In general unless a repo is asking for a refactoring, many open source projects don't take PRs that shuffle around code. In more busy projects it can have the effect of creating merge conflicts for outstanding work and it also affects the "git blame" output. Not that it can't be done, just that the costs need to be outweighed by the benefits. Most projects will prefer a PR that has a measurable impact, for example if it solves a bug (those are the easiest PRs to get merged in).

## Help wanted

If you're interested in buildpack work I would suggest taking a look at https://github.com/heroku/buildpacks-ruby it doesn't have the same secret burden overhead and is the future of buildpacks. It's in rust, but you don't have to have rust experience to open an issue or fix a doc typo etc. It could maybe be more worth your time.

For more about the CNB project check out https://github.com/heroku/buildpacks and I would love it if you ran through the tutorial https://github.com/heroku/buildpacks/blob/main/TUTORIAL.md and gave us feedback on the discussion https://github.com/heroku/buildpacks/discussions.

PR #1439, opened 2024-04-10
Heroku-24 and multi-arch (arm/amd) - Stacked 3/3

schneems:
Heroku-24 base image supports two architectures amd64 and arm64. We've built binaries for these two architectures https://github.com/heroku/docker-heroku-ruby-builder/pull/38. Effectively this means that the s3 bucket that holds the Ruby binaries has an additional folder. Previously files were at `<stack>/ruby-<version>.tgz` now they are at `<stack>/<arch>/ruby-<version>.tgz` but only for `heroku-24` and future stacks moving forward.

To support multiple architectures, the buildpack needs to detect the current architecture and whether or not the current stack supports multiple architectures.

Beyond downloading binaries, the buildpack is aware of the S3 structure to travers version numbers in order to warn customers when a newer version of a ruby version is available. We also warn customers if their current Ruby version is not available on the next stack. This behavior is implemented and tested in this PR, however we're not turning on warnings for `heroku-24` yet as customers cannot currently use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Heroku-24 stack initial support. Includes multi-architecture (arm64/amd64) logic that has not been tested on the platform (https://github.com/heroku/heroku-buildpack-ruby/pull/1439)
 - Remove unused Rubinius and Ruby 1.9.2 codepaths (https://github.com/heroku/heroku-buildpack-ruby/pull/1440)
 
 ## [v267] - 2024-02-28
--- a/lib/language_pack/base.rb
+++ b/lib/language_pack/base.rb
@@ -20,6 +20,8 @@ class LanguagePack::Base
   VENDOR_URL           = ENV['BUILDPACK_VENDOR_URL'] || "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com"
   DEFAULT_LEGACY_STACK = "cedar"
   ROOT_DIR             = File.expand_path("../../..", __FILE__)
+  MULTI_ARCH_STACKS    = ["heroku-24"]
+  KNOWN_ARCHITECTURES  = ["amd64", "arm64"]
 
   attr_reader :build_path, :cache, :stack
 
@@ -35,8 +37,23 @@ class LanguagePack::Base
     @id            = Digest::SHA1.hexdigest("#{Time.now.to_f}-#{rand(1000000)}")[0..10]
     @fetchers      = {:buildpack => LanguagePack::Fetcher.new(VENDOR_URL) }
     @layer_dir     = layer_dir
+    @arch = get_arch
 
     Dir.chdir build_path
+  end
+
+  def get_arch
+    command = "dpkg --print-architecture"
+    arch = run!(command, silent: true).strip
+
+    if !KNOWN_ARCHITECTURES.include?(arch)
+      raise <<~EOF
+        Architecture '#{arch}' returned from command `#{command}` is unknown.
+        Known architectures include: #{KNOWN_ARCHITECTURES.inspect}"
+      EOF
+    end
+
+    arch
   end
 
   def self.===(build_path)
--- a/lib/language_pack/fetcher.rb
+++ b/lib/language_pack/fetcher.rb
@@ -7,10 +7,11 @@ module LanguagePack
 
     include ShellHelpers
 
-    def initialize(host_url, stack: nil)
+    def initialize(host_url, stack: nil, arch: nil)
       @host_url = Pathname.new(host_url)
       # File.basename prevents accidental directory traversal
       @host_url += File.basename(stack) if stack
+      @host_url += File.basename(arch) if arch
     end
 
     def exists?(path, max_attempts = 1)
--- a/lib/language_pack/helpers/download_presence.rb
+++ b/lib/language_pack/helpers/download_presence.rb
@@ -18,13 +18,17 @@
 class LanguagePack::Helpers::DownloadPresence
   STACKS = ['heroku-20', 'heroku-22']
 
-  def initialize(file_name:, stacks: STACKS)
+  def initialize(file_name:, arch: , multi_arch_stacks:, stacks: STACKS )
     @file_name = file_name
     @stacks = stacks
     @fetchers = []
     @threads = []
     @stacks.each do |stack|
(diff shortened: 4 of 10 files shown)


lib/language_pack/installers/rbx_installer.rb
```diff
@@ -1,20 +0,0 @@
-require "language_pack/installers/ruby_installer"
-require "language_pack/shell_helpers"
-
-class LanguagePack::Installers::RbxInstaller
```
joshwlewis, line 4: Does dropping this file mean the buildpack is dropping rubinius support with this release? If so, it should be in the changelog. If not, maybe this removal belongs in another PR?

lib/language_pack/ruby_version.rb
```diff
@@ -87,18 +83,6 @@ def jruby?
...
 
-    # determine if we're using rbx
-    # @return [Boolean] true if we are and false if we aren't
-    def rbx?
-      engine == :rbx
-    end
-
-    # determines if a build ruby is required
-    # @return [Boolean] true if a build ruby is required
-    def build?
```
joshwlewis, line 98: Should this also be in the changelog? "Dropped support for building 1.8.7 and 1.9.2" or something? This seems like something that could be shipped independently from heroku-24/multi-arch?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Heroku-24 stack image preview support along with multi-architecture (arm64/amd64) support (https://github.com/heroku/heroku-buildpack-ruby/pull/1439)
```
joshwlewis, line 5: I'd suggest dropping this "multi-architecture support" / "arm64 support" bit. We don't have any real way to run or integration test classic buildpacks on arm64 hardware. We also don't currently have a way to run the resulting slugs on arm64 either.

joshwlewis (changes requested):
I'd suggest reframing this PR as heroku-24 support (without the `arm64` stuff). There's really no way to activate the `arm64` codepath (aside from unit tests), so we probably shouldn't be advertising support for it (I think it's totally fine to have the `arm64` codepath, though).

lib/language_pack/ruby_version.rb
schneems: The oldest version of ruby on heroku-20 is 2.5 and heroku-22 is 3.1 (https://devcenter.heroku.com/articles/ruby-support#unsupported-ruby-versions). Ruby 1.9.2 was not available on heroku-18 either:

```
⛄️ 3.1.4 🚀 /Users/rschneeman/Documents/projects/rpm-Project-Code-gatech-kbai (main)
$ curl -I https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-18/ruby-2.5.7.tgz
HTTP/1.1 200 OK
x-amz-id-2: Hl89PTav6LfvU90vP97YsBc5DEBptax9s4PZU+3ubIrIUBY54lD0f2FXMOCgh9XQEQRU+5ia1x4=
x-amz-request-id: KVDZM54WRBN5ASGT
Date: Mon, 15 Apr 2024 15:55:27 GMT
Last-Modified: Tue, 01 Oct 2019 14:46:30 GMT
ETag: "93ff97625abfc6ccd1a071bb0fda8e66"
x-amz-version-id: null
Accept-Ranges: bytes
Content-Type:
Server: AmazonS3
Content-Length: 11995820

⛄️ 3.1.4 🚀 /Users/rschneeman/Documents/projects/rpm-Project-Code-gatech-kbai (main)
$ curl -I https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-18/ruby-1.9.2.tgz
HTTP/1.1 404 Not Found
x-amz-request-id: 1NKVW4VD3T94T71R
x-amz-id-2: tTow7CFW9BcWmAbkEWvrqPz5zJdsy7ry8Jm39v+EURa4RN/UpslBhhGf1iHzOrpJ4n1BJxiGGoI=
Content-Type: application/xml
Date: Mon, 15 Apr 2024 15:55:31 GMT
Server: AmazonS3
```

This is removing REALLY old cruft, so I don't think it needs a changelog. The rubinius stuff is in a similar situation. I have a habit of not touching even dead code branches in the buildpack, but keeping those unused paths around means that it's more resistant to change. I've come to understand from working with Ed how much of a liability not-cleaning up such paths can be. As this code was making my job harder, I decided that now was a good time to remove it.

In short: It's being removed right now because it was difficult to disentangle the prior download and build logic which touched these 2 other branches:

- Build rubies
- Rubinius

I'll try extracting it out to it's own PR, but this PR will be blocked on that one getting merged first.

CHANGELOG.md
schneems: I am trying on the phrase "preview support," though clearly the **preview** didn't weight that much on your interpretation of the message.

I want to memorialize when that logic went live so that X years from now we could find this moment in time based on the changelog. I'm worried if I don't add it now, I'll never add it in the future. 

Also, people use the classic buildpack off of Heroku. Mostly dokku is the largest consumer.

So I would like the word "arm" to show up in the changelog. Maybe there's a better wording. Here's a stab:

> - Initial Heroku-24 base image support. Includes initial functionality to detect and use multi-architecture (arm64/amd64) binaries, though this functionality is not tested or verified (https://github.com/heroku/heroku-buildpack-ruby/pull/1439)

I think this is fairly explicit about what this PR is and isn't.

schneems:
@joshwlewis I split up this PR into three PRs:

- https://github.com/heroku/heroku-buildpack-ruby/pull/1441
- https://github.com/heroku/heroku-buildpack-ruby/pull/1440
- This one

lib/language_pack/base.rb
```diff
@@ -35,10 +36,24 @@ def initialize(build_path, cache_path = nil, layer_dir=nil)
     @id            = Digest::SHA1.hexdigest("#{Time.now.to_f}-#{rand(1000000)}")[0..10]
     @fetchers      = {:buildpack => LanguagePack::Fetcher.new(VENDOR_URL) }
     @layer_dir     = layer_dir
+    @arch = get_arch(stack: @stack)
 
     Dir.chdir build_path
   end
 
+  def get_arch(stack:)
+    output = run!("arch", silent: true).strip
```
edmorley, line 45: If you use `dpkg --print-architecture` then you won't need the mapping, since `dpkg --print-architecture` prints `amd64` and `arm64`.

schneems:
Updated to use `dpkg --print-architecture` and rebased against main.

joshwlewis (approved):

PR #1440, opened 2024-04-15
Remove RBX and "build" support - Stacked 2/3

schneems:
Rubinius is a Ruby implementation that was available to use on the platform a long time ago but has not been supported for some time:

Ruby 1.9.2 requires a "build" branch of logic that can be removed as well. It is not on any existing stacks and was not present on Heroku-18:

```
$ curl -I https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-18/ruby-2.5.7.tgz
HTTP/1.1 200 OK
x-amz-id-2: Hl89PTav6LfvU90vP97YsBc5DEBptax9s4PZU+3ubIrIUBY54lD0f2FXMOCgh9XQEQRU+5ia1x4=
x-amz-request-id: KVDZM54WRBN5ASGT
Date: Mon, 15 Apr 2024 15:55:27 GMT
Last-Modified: Tue, 01 Oct 2019 14:46:30 GMT
ETag: "93ff97625abfc6ccd1a071bb0fda8e66"
x-amz-version-id: null
Accept-Ranges: bytes
Content-Type:
Server: AmazonS3
Content-Length: 11995820
$ curl -I https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-18/ruby-1.9.2.tgz
HTTP/1.1 404 Not Found
x-amz-request-id: 1NKVW4VD3T94T71R
x-amz-id-2: tTow7CFW9BcWmAbkEWvrqPz5zJdsy7ry8Jm39v+EURa4RN/UpslBhhGf1iHzOrpJ4n1BJxiGGoI=
Content-Type: application/xml
Date: Mon, 15 Apr 2024 15:55:31 GMT
Server: AmazonS3
```

Other unused functionality that's being removed:

- CDN support for downloading Ruby binaries. This provided no meaningful download improvement when deploying as the binaries are on S3 in the same data center. This was disabled a long time ago.
- A "default cache" that was warmed with common gems for the most common Ruby version. This was effective at reducing "first build" time but the complexity exploded if we were to try and support N versions of Rails and M versions of Ruby for this feature. Additionally the community has moved to pre-built binaries for some expensive components such as nokogiri.

PR #1441, opened 2024-04-15
Fix tests after bundler change - Stacked 1/3

schneems:
The Ruby buildpack uses the Ruby buildpack on CI. When the bundler versions changed in https://github.com/heroku/heroku-buildpack-ruby/commit/56ff138eea331620d3579370693dd39140a642cd it also changed the behavior of these tests (which use the local bundler version). That means that after the buildpack was deployed, it retroactively meant that tests on main were failing.

The issue with these tests is that they're using a feature of bundler that's been removed, specifically this change https://devcenter.heroku.com/changelog-items/2809. In these tests the Ruby version is specified in the Gemfile but not the Gemfile.lock:

- Gemfile https://github.com/sharpstone/mri_193/blob/master/Gemfile
- Gemfile.lock https://github.com/sharpstone/mri_193/blob/master/Gemfile.lock 

With the newer version of bundler, these tests now fail. To fix this tests will need to be updated or removed (based on their current relevance).


spec/helpers/ruby_version_spec.rb
```diff
@@ -60,20 +60,6 @@
...
-        version_number = "1.9.3"
-        version        = "ruby-#{version_number}"
-        expect(ruby_version.version_without_patchlevel).to eq(version)
-        expect(ruby_version.engine_version).to eq(version_number)
-        expect(ruby_version.to_gemfile).to eq("ruby '#{version_number}'")
-        expect(ruby_version.engine).to eq(:ruby)
-      end
-    end
-  end
-
```
schneems, line 76: 1.9.3 is no longer supported on the platform, we can remove this test

spec/helpers/ruby_version_spec.rb
```diff
@@ -103,11 +89,39 @@
...
+
+          RUBY VERSION
+            ruby 3.2.3p157
+
+          BUNDLED WITH
+            2.4.19
+        EOF
+
         ruby_version   = LanguagePack::RubyVersion.new(@bundler.install.ruby_version, is_new: true)
-        version_number = "2.0.0"
```
schneems, line 110: Update test to use a modern Ruby version and move away from the less flexible github based fixture. Write the Gemfile and Gemfile.lock instead so the developer can see what's being used in the test at a glance.

spec/helpers/ruby_version_spec.rb
```diff
@@ -117,32 +131,42 @@
...
+        ruby_version   = LanguagePack::RubyVersion.new(
+          @bundler.install.ruby_version,
+          is_new: true
+        )
+        version_number = "2.6.8"
+        engine_version = "9.3.6.0"
+        engine = :jruby
+        expect(ruby_version.version_without_patchlevel).to eq("ruby-#{version_number}-#{engine}-#{engine_version}")
         expect(ruby_version.engine_version).to eq(engine_version)
-        expect(ruby_version.to_gemfile).to eq(to_gemfile)
```
schneems, line 132: Same as above, update to not rely on a github template. Update to use a more recent Ruby version

spec/helpers/ruby_version_spec.rb
```diff
@@ -117,32 +131,42 @@
...
-  it "surfaces error message from bundler"  do
-    bundle_error_msg = "Zomg der was a problem in da gemfile"
-    error_klass      = LanguagePack::Helpers::BundlerWrapper::GemfileParseError
-    Hatchet::App.new("bad_gemfile_on_platform").in_directory_fork do |dir|
-      Bundler.with_unbundled_env do
-        @bundler       = LanguagePack::Helpers::BundlerWrapper.new().install
-        expect { LanguagePack::RubyVersion.new(@bundler.ruby_version) }.to raise_error(error_klass, /#{Regexp.escape(bundle_error_msg)}/)
-      end
-    end
-  end
```
schneems, line 147: This test is supposed to exercise this code https://github.com/heroku/heroku-buildpack-ruby/blob/7866d1efbdc0e4d5f66caff2cdd7f83f76d4f8e0/lib/language_pack/helpers/bundler_wrapper.rb#L176

However bundler no longer errors on `bundle platform --ruby` depending on what version of bundler is running.

In the future we will move to get rid of the concept of "blessed" bundler versions and to BYOBv (bring your own bundler version) which means all that code will eventually go away. I don't want to remove the check and re-raise logic right now though. If we wanted to test this functionality we would need to move away from these local "unit-ish" tests and to a full blown hatchet test that tries to deploy an app.

joshwlewis (approved):
Nice. Thanks for extracting. This totally makes sense in isolation.

schneems:
It was merged in with https://github.com/heroku/heroku-buildpack-ruby/pull/1440 which was also approved. I clicked the "enable auto merge" but didn't see that my own comments were considered blocking so it wasn't merged in. That PR had the exact same commits anyway, so it's the exact same code. Leaving a note as to why I'm closing this after approval.

PR #1445, opened 2024-04-22
Deprecate beta CNB support

schneems:
CNB support was introduced into this buildpack as an experiment. The experiment is over, anyone wanting to use the official CNB (still in "preview" support) should use https://github.com/heroku/buildpacks-ruby

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Deprecate CNB support, CNB users should use [heroku/buildpacks-ruby](https://github.com/heroku/buildpacks-ruby) instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1445)
+- Deprecate CNB support in this buildpack; CNB support for Ruby is provided by [heroku/buildpacks-ruby](https://github.com/heroku/buildpacks-ruby) instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1445)
 
 ## [v268] - 2024-04-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Deprecate CNB support, CNB users should use [heroku/buildpacks-ruby](https://github.com/heroku/buildpacks-ruby) instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1445)
 
 ## [v268] - 2024-04-17
 
--- a/bin/build
+++ b/bin/build
@@ -15,7 +15,7 @@ source "$BIN_DIR/support/bash_functions.sh"
 heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$BUILDPACK_DIR"
 
 cat<<EOF 1>&2
-    CNB no longer maintained
+    The CNB implementation in this buildpack is no longer maintained
 
     The heroku/heroku-ruby-buildpack previously implemented the Cloud Native Buildpack \(CNB\) spec beta support, however this functionality is no longer supported or maintained.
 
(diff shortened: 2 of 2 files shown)


edmorley (commented):
I think you could remove this file (and the CNB subsets of `buildpack.toml`) now, given that the CNB in this repo implements Buildpack API 0.2, which has not been supported by our builders since last year:
https://github.com/heroku/heroku-buildpack-ruby/blob/5ce2c2eb026d4c72d0df82cb1bfc028bc2e2bb7b/buildpack.toml#L1
https://github.com/heroku/cnb-builder-images#available-images
https://github.com/buildpacks/lifecycle#supported-apis

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Deprecate CNB support, CNB users should use [heroku/buildpacks-ruby](https://github.com/heroku/buildpacks-ruby) instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1445)
```
joshwlewis, line 5: Optional: this sounds a little like we're 👎🏻 about CNB support. 

```suggestion
- Deprecate CNB support in this buildpack; CNB support for Ruby is provided by [heroku/buildpacks-ruby](https://github.com/heroku/buildpacks-ruby) instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1445)
```

bin/build
```diff
@@ -8,10 +8,18 @@ APP_DIR=$(pwd)
...
 
-# legacy buildpack uses $STACK 
+# legacy buildpack uses $STACK
 export STACK=$CNB_STACK_ID
 
 source "$BIN_DIR/support/bash_functions.sh"
 heroku_buildpack_ruby_install_ruby "$BIN_DIR" "$BUILDPACK_DIR"
 
+cat<<EOF 1>&2
+    CNB no longer maintained
```
joshwlewis, line 18:
```suggestion
    The CNB implementation in this buildpack is no longer maintained
```

joshwlewis (approved):
Looks good to me. Left some optional suggestions about making this sound less like we're abandoning CNB altogether and more like the functionality has moved.

schneems:
@edmorley I'm up for removing all this code, but the Ruby integrations run deep and I'm trying to get codon unblocked on heroku-24, however I've got no idea how people might be using this TBH, just that they might be. I know waypoint use(d?) heroku CNBs before we implemented ours in rust, but I'm not sure what mechanism they were using.

I saw this code while working on adding shellcheck so I can fix ruby bootstrapping on heroku-24 (the vendored version logic is incorrect on that stack, but all the logic lives in bash). 

In Ruby-land I like to deprecate, release, and then remove where possible. I was thinking of doing that here.

edmorley:
> @edmorley I'm up for removing all this code, but the Ruby integrations run deep and I'm trying to get codon unblocked on heroku-24, however I've got no idea how people might be using this TBH, just that they might be. I know waypoint use(d?) heroku CNBs before we implemented ours in rust, but I'm not sure what mechanism they were using.

Every single third party I've found (over several hours of searching, as part of https://github.com/heroku/cnb-builder-images/pull/429#issuecomment-1785095460) uses our buildpacks from the registry or via cnb-shim (which uses the classic buildpack from the classic buildpack registry's S3 bucket).

There is also no CNB publishing pipeline in this repo any more, so no one will execute the deprecation warning.

However, there is no harm in doing it this way if you prefer - I just wanted to make you aware that in my opinion it was already safe to delete, should you wish to do so :-)

edmorley:
> I know waypoint use(d?) heroku CNBs before we implemented ours in rust, but I'm not sure what mechanism they were using.

They use the builder image:
https://github.com/search?q=repo%3Ahashicorp%2Fwaypoint+%22heroku%2Fbuildpacks%3A20%22&type=code

xref:
https://github.com/hashicorp/waypoint/issues/4911

PR #1446, opened 2024-04-22
Fix heroku-24 bootstrap ruby support

schneems:
The Ruby buildpack needs Ruby to run. For the heroku-24 stack the location of ruby binaries changed to include an architecture specifier (amd64/arm64) but the logic for bootstrapping a Ruby version was not updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fix Heroku-24 Ruby buildpack bootstrapping logic (https://github.com/heroku/heroku-buildpack-ruby/pull/1446)
 - Deprecate CNB support in this buildpack; CNB support for Ruby is provided by [heroku/buildpacks-ruby](https://github.com/heroku/buildpacks-ruby) instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1445)
 
 ## [v268] - 2024-04-17
--- a/bin/support/bash_functions.sh
+++ b/bin/support/bash_functions.sh
@@ -30,7 +30,14 @@ install_bootstrap_ruby()
 {
   local bin_dir=$1
   local buildpack_dir=$2
-  local heroku_buildpack_ruby_dir="$buildpack_dir/vendor/ruby/$STACK"
+
+  # Multi-arch aware stack support
+  if [ "$STACK" == "heroku-24" ]; then
+    local arch=dpkg --print-architecture
+    local heroku_buildpack_ruby_dir="$buildpack_dir/vendor/ruby/$STACK/$arch"
+  else
+    local heroku_buildpack_ruby_dir="$buildpack_dir/vendor/ruby/$STACK"
+  fi
 
   # The -d flag checks to see if a file exists and is a directory.
   # This directory may be non-empty if a previous compile has
--- a/bin/support/bash_functions.sh
+++ b/bin/support/bash_functions.sh
@@ -33,7 +33,7 @@ install_bootstrap_ruby()
 
   # Multi-arch aware stack support
   if [ "$STACK" == "heroku-24" ]; then
-    local arch=dpkg --print-architecture
+    local arch=$(dpkg --print-architecture)
     local heroku_buildpack_ruby_dir="$buildpack_dir/vendor/ruby/$STACK/$arch"
   else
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fix Heroku-24 Ruby buildpack bootstrapping logic (https://github.com/heroku/heroku-buildpack-ruby/pull/1446)
 - Deprecate CNB support in this buildpack; CNB support for Ruby is provided by [heroku/buildpacks-ruby](https://github.com/heroku/buildpacks-ruby) instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1445)
```
schneems, line 6: Stacked on #1444, change the branch target to main before merging (before this comment is resolved).

buildpack.toml
```diff
@@ -30,6 +30,14 @@ dir = "vendor/ruby/heroku-20"
 url = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-22/ruby-3.1.4.tgz"
 dir = "vendor/ruby/heroku-22"
 
+[[publish.Vendor]]
+url = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-22/amd64/ruby-3.1.4.tgz"
+dir = "vendor/ruby/amd64/heroku-24"
+
+[[publish.Vendor]]
+url = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-22/arm64/ruby-3.1.4.tgz"
+dir = "vendor/ruby/arm64/heroku-24"
```
edmorley, line 39: I think we should only vendor the AMD64 Ruby binary in the classic buildpack registry release, since otherwise Codon is paying the boot time delay price for every build (Ruby or not) for an architecture that will never get used on Heroku.

Even if the ARM64 binary isn't included, that arch will still work on eg Dokku, since it will be dynamically downloaded by the bootstrap script.

bin/support/bash_functions.sh
```diff
@@ -30,7 +30,14 @@ install_bootstrap_ruby()
 {
   local bin_dir=$1
   local buildpack_dir=$2
-  local heroku_buildpack_ruby_dir="$buildpack_dir/vendor/ruby/$STACK"
+
+  # Multi-arch aware stack support
+  if [ "$STACK" == "heroku-24" ]; then
+    local arch=dpkg --print-architecture
```
edmorley, line 36: This needs wrapping in `$()` otherwise `arch` will be a string containing the command name/args, rather than the output of executing the command.

```suggestion
    local arch=$(dpkg --print-architecture)
```

bin/support/download_ruby
```diff
@@ -25,11 +25,23 @@ curl_retry_on_18() {
   return $ec
 }
 
+ruby_url() {
+  local stack=$1
+  local version=$2
+
+  if [ "$stack" == "heroku-24" ]; then
+    local arch=dpkg --print-architecture
```
edmorley, line 33:
```suggestion
    local arch=$(dpkg --print-architecture)
```

edmorley (changes requested):

buildpack.toml
schneems: Thanks for the review, I removed it. Can you take another look?

buildpack.toml
```diff
@@ -30,6 +30,10 @@ dir = "vendor/ruby/heroku-20"
 url = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-22/ruby-3.1.4.tgz"
 dir = "vendor/ruby/heroku-22"
 
+[[publish.Vendor]]
+url = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-22/amd64/ruby-3.1.4.tgz"
```
edmorley, line 34:
```suggestion
url = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-24/amd64/ruby-3.1.4.tgz"
```

edmorley (approved):

buildpack.toml
schneems: Good catch. I updated it locally and added a test that should catch this type of mistake going forward.

PR #1453, opened 2024-05-17
Testing for 1452 - Fix propshaft compatibility 

schneems:
Testing for https://github.com/heroku/heroku-buildpack-ruby/pull/1452

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fix issue with sprockets reference being nil when
+  propshaft gem is used instead (https://github.com/heroku/heroku-buildpack-ruby/pull/1452)
 
 ## [v270] - 2024-04-23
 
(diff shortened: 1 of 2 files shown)


edmorley:
@schneems You don't need to open a new PR for these - instead clone the commits and push to a branch on the repo as-is - the CI result will be posted to the original PR too (since GitHub's status checks API uses the SHA256 as the identifier for the CI result - not the PR/branch reference, so the most recent CI result for that SHA256 will show up on the original PR). Doing it this way means (a) no need to ask someone else for review (you can review the original PR yourself), (b) the original PR author gets full attribution for the merged PR.

In the event that you need to make additional changes to the contributed PR, you can still use the above workflow, but instead also need to push those commits to their PR branch (GitHub defaults to allowing you to push to the branches of a fork, when a PR is opened from that fork.)

schneems:
Thanks. In the past the compliance check never finishes. It didn't here either. But I was able to "re-sync" with github and then it worked.

Issue #1306, opened 2022-05-25
Amazon S3 is extremely slow


itsNikolay:
Download speed from Amazon S3 is 8KB/s:
(I checked it on 2 different internet providers and both have the same 8Kb/s)
```ruby
$ wget https://s3.amazonaws.com/heroku-nodebin/node/release/linux-x64/node-v16.13.1-linux-x64.tar.gz
--2022-05-25 23:13:04--  https://s3.amazonaws.com/heroku-nodebin/node/release/linux-x64/node-v16.13.1-linux-x64.tar.gz
Resolving s3.amazonaws.com (s3.amazonaws.com)... 54.231.201.200
Connecting to s3.amazonaws.com (s3.amazonaws.com)|54.231.201.200|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 32395116 (31M) [application/x-tar]
Saving to: ‘node-v16.13.1-linux-x64.tar.gz’

node-v16.13.1-linux-x64.tar.gz               1%[=>    ] 355.64K  7.39KB/s    eta 91m 36s
```
Should we add way to change Node URL  to bypass such case? 
Probably from ENV var? 
It's hardcoded now :(  https://github.com/heroku/heroku-buildpack-ruby/blob/main/lib/language_pack/helpers/nodebin.rb#L8
Or probably you can suggest another way ?

schneems:
It's possible to specify a different bucket for binaries via an env var (when running the buildpack, not at the app level). But it's not really a common use case. Heroku runs on AWS and uses an S3 bucket in the same data center as our builds. We don't really have much support for off-heroku installations right now. 91m is quite a long time though. That seems unusually high.

PR #1456, opened 2024-05-29
Update nodebin.rb to support CUSTOM_HEROKU_NODEBIN_HOST env

agungyuliaji:
ref: https://github.com/heroku/heroku-buildpack-ruby/issues/1306


agungyuliaji:
Apparently there is a multi buildpack step https://devcenter.heroku.com/articles/using-multiple-buildpacks-for-an-app and https://github.com/heroku/heroku-buildpack-nodejs support for custom binary urls.

My rails application deploys well (via dokku), with a custom nodejs endpoint according to the configuration

PR #1464, opened 2024-06-10
Remove experimental CNB support

edmorley:
The buildpack in this repo is primarily a classic Heroku buildpack, however, as part of the initial exploration into CNBs had experimental CNB support added some time ago.

However, the maintained Ruby CNB now exists in a separate repo: https://github.com/heroku/buildpacks-ruby

The experimental CNB support in this repo doesn't actually work any more since the buildpack API version it implements (v0.2) isn't supported by modern `lifecycle` versions - and attempts at building encounter this error:

```
ERROR: failed to set API for Buildpack 'heroku/ruby@0.1.4': buildpack API version '0.2' is incompatible with the lifecycle
```

In addition, it's not even possible to use this repo with a CNB build without having cloned it locally, since:
1. The CNB parts are no longer published
2. The old CNB release assets were deleted a year or so ago: https://github.com/heroku/heroku-buildpack-ruby/releases
3. Pack/lifecycle doesn't support cloning Git URLs.
4. Attempting to use a buildpack URL pointing at the GitHub gzip archive fails due to GitHub's nesting of the repo inside a subdirectory:

```
$ pack build --builder heroku/builder:22 --buildpack https://github.com/heroku/heroku-buildpack-ruby/archive/refs/heads/main.tar.gz ruby-test
...
Downloading from https://github.com/heroku/heroku-buildpack-ruby/archive/refs/heads/main.tar.gz
93.2 KB/-1 B
ERROR: failed to build: downloading buildpack: extracting from https://github.com/heroku/heroku-buildpack-ruby/archive/refs/heads/main.tar.gz: reading buildpack: reading buildpack.toml: could not find entry path 'buildpack.toml': not exist
```

Given both the Buildpack API error, and the GitHub URL issues no one is using the CNB implementation here, and so its dead code that should be removed to prevent confusion (eg over where the CNB lives).

A deprecation warning was previously added in #1445 (though as above it's unlikely anyone even saw that message, since the build was already erroring before it gets that far).


schneems (approved):
Thanks Ed.

PR #1469, opened 2024-06-19
Let bundler >= 2.2 handle windows platform

schneems:
Building on https://github.com/heroku/heroku-buildpack-ruby/pull/1458

Tests and documents Windows bundler 2.2+ behavior.

### DIFF
--- a/changelogs/unreleased/windows_gemfile.md
+++ b/changelogs/unreleased/windows_gemfile.md
@@ -12,7 +12,7 @@ Ruby applications that use Windows and Bundler 2.2+ will no longer have their `G
 > git commit -m "Upgrade bundler"
 ```
 
-This change is reflected in the [Deploying a Ruby Project Generated on Windows](https://devcenter.heroku.com/articles/bundler-windows-gemfile) Devcenter article. More information on the history of the behavior can be found [on this GitHub issue](https://github.com/heroku/heroku-buildpack-ruby/issues/1157).
+This change is reflected in the [Deploying a Ruby Project Generated on Windows](https://devcenter.heroku.com/articles/bundler-windows-gemfile) Dev Center article. More information on the history of the behavior can be found [on this GitHub issue](https://github.com/heroku/heroku-buildpack-ruby/issues/1157).
 
 <!--
 
--- a/changelogs/unreleased/windows_gemfile.md
+++ b/changelogs/unreleased/windows_gemfile.md
@@ -1,0 +1,67 @@
+## Native Gemfile.lock support for Windows Ruby users with Bundler 2.2+
+
+Ruby applications that use Windows and Bundler 2.2+ will no longer have their `Gemfile.lock` deleted and re-generated on deployment. Instead, they will rely on Bundler 2.2+'s support for multiple platforms to correctly resolve dependencies. We recommend any Windows Ruby user to upgrade their Bundler version in the `Gemfile.lock` to Bundler 2.2+ to take advantage of this behavior:
+
+> note
+> The leading `>` indicates a command prompt and should not be copied.
+
+```
+> gem install bundler
+> bundle update --bundler
+> git add Gemfile.lock
+> git commit -m "Upgrade bundler"
+```
+
+This change is reflected in the [Deploying a Ruby Project Generated on Windows](https://devcenter.heroku.com/articles/bundler-windows-gemfile) Devcenter article. More information on the history of the behavior can be found [on this GitHub issue](https://github.com/heroku/heroku-buildpack-ruby/issues/1157).
+
+<!--
+
+https://devcenter.heroku.com/admin/articles/315/edit
+
+# Using Bundler
+
+To use, install bundler run:
+
+> note
+> Commands are prefixed with `>` to indicate they should be run in a command prompt, do not copy the `>` character.
+
+```term
+> gem install bundler
+```
+
+Create a file named `Gemfile` in the root of your app specifying what gems are required to run it:
+
+```ruby
+source "https://rubygems.org"
+
+gem 'sinatra', '4.0'
+```
+
+This file should be added to the git repository since it is part of the app. You should also add the `.bundle` directory to your `.gitignore` file. Once you have added the `Gemfile`, it makes it easy for other developers to get their environment ready to run the app:
+
+```term
+> bundle install
+```
+
+This command ensures that all gems specified in the `Gemfile` are available for your application. Running `bundle install` also generates a `Gemfile.lock`, which should be added to your git repository. The `Gemfile.lock` ensures that deployed versions of gems on Heroku match the version installed locally on your development machine.
+
+>warning
+>If your `Gemfile.lock` specifies a bundler version prior to 2.2 and the `PLATFORMS` section of your `Gemfile.lock` contains Windows entries, such as `mswin` or `mingw`, then the `Gemfile.lock` file will be ignored on Heroku. We recommend upgrading to Bundler 2.2 or later.
+
+## Windows support with Bundler 2.2 and later
+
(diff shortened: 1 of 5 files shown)


changelogs/unreleased/windows_gemfile.md
```diff
@@ -0,0 +1,67 @@
...
+> The leading `>` indicates a command prompt and should not be copied.
+
+```
+> gem install bundler
+> bundle update --bundler
+> git add Gemfile.lock
+> git commit -m "Upgrade bundler"
+```
+
+This change is reflected in the [Deploying a Ruby Project Generated on Windows](https://devcenter.heroku.com/articles/bundler-windows-gemfile) Devcenter article. More information on the history of the behavior can be found [on this GitHub issue](https://github.com/heroku/heroku-buildpack-ruby/issues/1157).
```
edmorley, line 15:
```suggestion
This change is reflected in the [Deploying a Ruby Project Generated on Windows](https://devcenter.heroku.com/articles/bundler-windows-gemfile) Dev Center article. More information on the history of the behavior can be found [on this GitHub issue](https://github.com/heroku/heroku-buildpack-ruby/issues/1157).
```

edmorley (approved):

changelogs/unreleased/windows_gemfile.md
```diff
@@ -0,0 +1,67 @@
...
+<!--
+
+https://devcenter.heroku.com/admin/articles/315/edit
+
+# Using Bundler
+
+To use, install bundler run:
+
+> note
+> Commands are prefixed with `>` to indicate they should be run in a command prompt, do not copy the `>` character.
```
lilacstella, line 26: How is this md rendered? rn this looks like:
> note 
> Commands are prefixed... 

I think we can add a line in between here so it looks like

> note

> Commands are prefixed...

or using a colon
> Note: Commands are prefixed with...
```suggestion
> note

> Commands are prefixed with `>` to indicate they should be run in a command prompt, do not copy the `>` character.
```

lib/language_pack/ruby.rb
```diff
@@ -694,19 +694,32 @@ def build_bundler
...
-          Bundler will do a full resolve so native gems are handled properly.
-          This may result in unexpected gem versions being used in your app.
-          In rare occasions Bundler may not be able to resolve your dependencies at all.
-
-          https://devcenter.heroku.com/articles/bundler-windows-gemfile
-        WARNING
+        if bundler.supports_multiple_platforms?
+          puts "Windows `Gemfile.lock` detected, bundler #{@bundler.version} supports multiple platforms, no action taken."
+        else
+            File.unlink("Gemfile.lock")
```
lilacstella, line 700: does there still need to be a log bundle here?

lilacstella (changes requested):

changelogs/unreleased/windows_gemfile.md
schneems: > How is this md rendered? rn the github md render this looks like:

I'll try to find a stylesheet link, but the DevCenter supports some custom markdown blockquote identifiers. When it sees those, instead of producing a blockquote it styles them according to the first line. For example, in this article https://devcenter.heroku.com/articles/getting-started-with-rails6, the markdown for that callout is: 

```
>note
>Using dynos and databases to complete this tutorial counts towards your usage. We recommend using our [low-cost plans](https://blog.heroku.com/new-low-cost-plans) to complete this tutorial. Eligible students can apply for platform credits through our new [Heroku for GitHub Students program](https://blog.heroku.com/github-student-developer-program).
```

lib/language_pack/ruby.rb
schneems: That's a good question.

The mechanics behind this logging have been removed/disabled. We didn't clean them up because...it wasn't a priority and frankly I was worried about breaking something. Last time it was touched was https://github.com/heroku/heroku-buildpack-ruby/pull/1229 (found via git blame on GitHub's UI).

That `log` method seems to be defined here https://github.com/heroku/heroku-buildpack-ruby/blob/cf25ace0e09675e744603962e12451943ce979be/lib/language_pack/base.rb#L135 which calls https://github.com/heroku/heroku-buildpack-ruby/blob/cf25ace0e09675e744603962e12451943ce979be/lib/language_pack/base.rb#L217-L220 which shells out to `logger`. And that code was added by David Dollar in 2011, https://github.com/heroku/heroku-buildpack-ruby/commit/42a604366e609658971bd6bc56b32b06a964e057.

We're not currently using this functionality, so I think we should get rid of it. I will open an issue.

Our internal logging capabilities are in flux (constantly). I believe we're stabilizing on otel? soon when that happens, I think we should add logging to answer focused questions rather than expecting logging strings from 2011 to mean anything. We'll likely want to log windows builds, but we can add that functionality in when it comes time.

And FWIW we have other ways to get metrics for builds outside of these capabilities (so it's not as though we're operating with zero information). Buildpack logging helps to make getting those metrics easier and can allow us to collect targeted information as needed.

schneems: https://github.com/heroku/heroku-buildpack-ruby/issues/1470

lilacstella (approved):

schneems:
Merged now. I'm going to aim to deploy on monday.

lib/language_pack/ruby.rb
```diff
@@ -694,19 +694,32 @@ def build_bundler
...
-        warn(<<~WARNING, inline: true)
-          Removing `Gemfile.lock` because it was generated on Windows.
-          Bundler will do a full resolve so native gems are handled properly.
-          This may result in unexpected gem versions being used in your app.
-          In rare occasions Bundler may not be able to resolve your dependencies at all.
-
-          https://devcenter.heroku.com/articles/bundler-windows-gemfile
-        WARNING
+        if bundler.supports_multiple_platforms?
+          puts "Windows `Gemfile.lock` detected, bundler #{@bundler.version} supports multiple platforms, no action taken."
```
runephilosof-abtion, line 698: "No action taken": This text can be confusing, to people who don't know the history.
I think it is better to simply not print anything.

schneems: I have an output goal that is "anytime there's a change in behavior based on buildpack logic, developers should be able to tell by comparing the outputs." 

Humans are also bad at seeing "what's not there." I.e. If one build shows "Installing mail" but it's not in the other build, most people usually have to use diff tools to determine notice unless they're explicitly looking for that line.

Another goal is that if they have a failure related to that different action they should be able to trace that action in the same build. If someone's app started to fail they can scan up and might see that line and be curious only to discover that behavior is the cause of their problems and there's a bug in bundler.

So it's a little aggressive to log in the empty if case but I feel like it's a possibly VERY large change to the windows users and they deserve to know about it.

Now about "no action taken." If that phrase is confusing then we can remove it later or ask users what might be better. What I'm trying to avoid with that line is when you show people something, they want to know what happens next because of that thing.

You can see this with a storytelling narrative principle called "checkov's gun" which loosely is "if a gun is introduced in the first act it should go off in the third." Here's a funny video that is satire but demonstrates the principle https://www.youtube.com/watch?v=PQOvv_b9TpY. The viewer keeps waiting for it to "go off."

The problem with saying simply:

```
"Windows `Gemfile.lock` detected, bundler #{@bundler.version} supports multiple platforms"
```

Is that the reader will implicitly be waiting for the seemingly missing information. They will be thinking "why did you tell me that" as well as "well, what will you do with that information?"

The current output might be a little curious or weird. Maybe even eyebrow-raising. But it won't be anxiety or stress-inducing.

So I agree it's off, but it's also purposeful in it's off-ness.

runephilosof-abtion: It is a good point. I think something like this will ensure that people who was not aware of the previous behavior and notice the message are not confused and use time on investigating.
```suggestion
          puts "Windows `Gemfile.lock` detected, bundler #{@bundler.version} supports multiple platforms, no action taken (before June 25th, Heroku would have removed the lock file)."
```

PR #1490, opened 2024-08-12
Add Rails 8 support

chaadow:
with the development of rails 8 that has already started in [May 2024](https://rubyonrails.org/2024/5/17/this-week-in-rails)

I think we can add support for rails 8. ( I do personally maintain a production app that's on rails `8.0.0.alpha` aka master branch of rails repo)


schneems:
Merged in #1498

PR #1534, opened 2025-01-06
Update default Ruby version to 3.3.7

schneems:
The latest Ruby is now 3.4.x. Our ideal default policy is to use the latest release of the prior year's version. This PR updates to the latest Ruby 3.3.x version which is 3.3.7.


andreas-venturini:
@schneems could now default to 3.3.7 https://www.ruby-lang.org/en/news/2025/01/15/ruby-3-3-7-released/

edmorley (approved):

PR #1551, opened 2025-03-03
Add a mechanism for documenting Ruby versions via automation

schneems:
Our current deploy process looks like this for releasing a Ruby version:

- Manually make a PR to the Ruby buildpack with the version in the changelog
- Get it approved by another engineer
- Pre-Release the buildpack via automation
- Merge the pre-release PR
- Pull changes locally and run deploy command `bundle exec rake buildpack:release`
- Then trigger S3 changes via running automations on https://github.com/heroku/docker-heroku-ruby-builder

In the future there's a desire to have the https://github.com/heroku/docker-heroku-ruby-builder generate an "inventory" file that tracks binary checksums and then the buildpack will require a PR to utilize new versions. However, that's a non-trivial amount of work and Ruby versions keep getting released on Fridays when I'm on vacation and no one else is around.

With this change the process can look like this:

- Trigger this new workflow which updates the CHANGELOG.md and runs the pre-release automation in one
- Merge that PR
- Pull changes locally and run deploy command `bundle exec rake buildpack:release`
- Then trigger S3 changes via running automations on https://github.com/heroku/docker-heroku-ruby-builder

Which is a substantial workflow savings. This task can be removed once we implement the full inventory workflow (sometime in the vague future), but this is a quality of life upgrade for the short term.

### DIFF
--- a/.github/workflows/document_ruby_version.yml
+++ b/.github/workflows/document_ruby_version.yml
@@ -1,0 +1,43 @@
+name: Add Ruby version to changelog && prepare release
+run-name: "Add ${{ inputs.is_jruby && 'J' || ''}}Ruby ${{ inputs.ruby_version }} to the CHANGELOG.md and prepare a release"
+
+on:
+  workflow_dispatch:
+    inputs:
+      ruby_version:
+          description: "The Ruby version to announce"
+          type: string
+          required: true
+      is_jruby:
+          description: "JRuby version (as opposed to MRI)"
+          type: boolean
+          default: false
+          required: false
+
+# Disable all GITHUB_TOKEN permissions, since the GitHub App token is used instead.
+permissions: {}
+
+jobs:
+  prepare-release:
+    uses: heroku/languages-github-actions/.github/workflows/_classic-buildpack-prepare-release.yml@latest
+    secrets: inherit
+    with:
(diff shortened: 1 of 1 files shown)


.github/workflows/document_ruby_version.yml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      ruby_version:
+          description: "The Ruby version to announce"
+          type: string
+          required: true
+      is_jruby:
+          description: "JRuby version (as opposed to MRI)"
```
edmorley, line 12: (The existing description here sounded like the field was for a version string rather than a bool)

```suggestion
          description: "JRuby release? (as opposed to MRI)"
```

edmorley (approved):
I haven't tested the new `sed` command (that's new here compared to the `custom_update_command` used in the existing workflow), but this is easily iterated on once merged by triggering test PRs - so deferring to you for testing :-)

.github/workflows/document_ruby_version.yml
schneems: Thanks! Reminded me of this scene https://www.youtube.com/watch?v=X3zfP14pLxc

PR #1555, opened 2025-03-05
Add the ruby version to the current release

schneems:
This task runs after the new version is injected into the CHANGELOG. That means that if we add something under "unreleased" at this point, then it shows up in the wrong section. This commit fixes it.

### DIFF
--- a/.github/workflows/document_ruby_version.yml
+++ b/.github/workflows/document_ruby_version.yml
@@ -28,8 +28,7 @@ jobs:
 
         sed --in-place "/## \[v${NEW_VERSION}\] - ${DATE_TODAY}/a\\
         \\
-        - ${{ inputs.is_jruby && 'J' || ''}}Ruby ${{inputs.ruby_version}} is now available
-        " CHANGELOG.md
+        - ${{ inputs.is_jruby && 'J' || ''}}Ruby ${{inputs.ruby_version}} is now available" CHANGELOG.md
 
         sed --in-place --regexp-extended \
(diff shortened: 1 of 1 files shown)


schneems:
Generated PR https://github.com/heroku/heroku-buildpack-ruby/pull/1557/files

image

.github/workflows/document_ruby_version.yml
```diff
@@ -24,11 +24,12 @@ jobs:
...
         set -euo pipefail
+        DATE_TODAY="$(date --utc --iso-8601)"
 
-        sed --in-place '/## \[Unreleased\]/a\
-        \
+        sed --in-place "/## \[v${NEW_VERSION}\] - ${DATE_TODAY}/a\\
+        \\
         - ${{ inputs.is_jruby && 'J' || ''}}Ruby ${{inputs.ruby_version}} is now available
-        ' CHANGELOG.md
+        " CHANGELOG.md
```
edmorley, line 32: In the generated PR I see an extra trailing newline is added. I think this could be resolved by either moving this closing quote to the line before, or else adjusting the matching term to include a trailing `\n` (perhaps like `${DATE_TODAY}\n` ? though I haven't tested)

edmorley (approved):

.github/workflows/document_ruby_version.yml
schneems: I added a commit that moves the filename to the same line that we're adding and it has the same outcome:

image

I think that these lines might come from the prepare release task and it's not considering the use case of an empty prepare. This is what I get using the regular prepare task without adding a version

image

PR #1568, opened 2025-03-27
Ensure workflow uses same bundler version

schneems:


edmorley (approved):
CI seems to be failing from an unrelated test?
https://dashboard.heroku.com/pipelines/ac057663-170b-4bdd-99d0-87560eb3a570/tests/2297
https://github.com/heroku/heroku-buildpack-ruby/blob/c9d002af6add6b3be4d04518f41f5ce761fa2fcb/spec/helpers/outdated_ruby_version_spec.rb#L22

schneems:
> CI seems to be failing from an unrelated test?

That code is testing the version suggestion logic. Ruby 3.1.7 was released recently so it's now the latest. For the other cases I'm picking an already EOL version of Ruby, but heroku-24 and multi-stack suggestions have a minimum of 3.1. Thankfully, 3.1.7 just went EOL, so this test shouldn't fail like this in the future (unless there's some emergency 3.1.8 release for some reason, but that's not likely).

PR #1569, opened 2025-04-01
Add observability

schneems:
Commits are structured in order

### DIFF
--- a/lib/heroku_build_report.rb
+++ b/lib/heroku_build_report.rb
@@ -1,0 +1,53 @@
+require 'yaml'
+require 'pathname'
+
+# Observability reporting for builds
+#
+# Example usage:
+#
+#   HerokuBuildReport::GLOBAL.capture(
+#     "ruby_version" => "3.4.2"
+#   )
+module HerokuBuildReport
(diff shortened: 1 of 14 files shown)


lib/heroku_build_report.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# Observability reporting for builds
+#
+# Example usage:
+#
+#   HerokuBuildReport::GLOBAL.capture(
+#     "ruby_version" => "3.4.2"
+#   )
+module HerokuBuildReport
+  # Accumulates data in memory and writes it to the specified path in YAML format
```
edmorley, line 12: When I first read this comment I wondered what happened if the buildpack exited partway through (eg was the report lost since it's "in memory") - but I see that a write occurs as part of every `.capture()` call, so there will be no data loss (and no need for a final writer in a `rescue` somewhere). Is that something we can make clearer in this comment? :-)

edmorley (approved):
Thank you for adding this - exciting to see more languages start implementing `bin/report`! :-)

lib/heroku_build_report.rb
schneems: I added some extra docs

PR #1576, opened 2025-04-08
Add `.ruby-version` formats to observability metrics

schneems:
Rails 8 stopped generating `RUBY VERSION` in the Gemfile.lock and started generating a `.ruby-version` file. We will support this file in the future, but there is no spec for it.

This commit will observe the formats to help guide implementation decisions. It can also tell us how many people's `.ruby-version` files already align with their `Gemfile.lock` versions.

The logic is inside of the bundler wrapper class as this is the first class to read the ruby version from the Gemfile.lock via `$ bundle platform --ruby`.

### DIFF
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -112,8 +112,14 @@ class LanguagePack::Helpers::BundlerWrapper
 
     contents = @gemfile_lock_path.read(mode: "rt")
     bundled_with = contents.match(BUNDLED_WITH_REGEX)
+    dot_ruby_version_file = @gemfile_lock_path.join("..").join(".ruby-version")
     @report.capture(
-      "bundler.bundled_with" => bundled_with&.[]("version") || "empty"
+      "bundler.bundled_with" => bundled_with&.[]("version") || "empty",
+      # We use this bundler class to detect the Requested ruby version from the Gemfile.lock
+      # Rails 8 stopped generating `RUBY VERSION` in the Gemfile.lock and started generating
+      # a `.ruby-version` file. This will observe the formats to help guide implementation
+      # decisions
+      "ruby.dot_ruby_version" => dot_ruby_version_file.exist? ? dot_ruby_version_file.read&.strip : nil
     )
     @version = self.class.detect_bundler_version(
       contents: contents,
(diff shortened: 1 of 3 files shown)


lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -112,19 +112,25 @@ def initialize(options = {})
...
     bundled_with = contents.match(BUNDLED_WITH_REGEX)
+    dot_ruby_version_file = @gemfile_lock_path.join("..").join(".ruby-version")
     @report.capture(
-      "bundled_with" => bundled_with&.[]("version") || "empty"
+      "bundler.bundled_with" => bundled_with&.[]("version") || "empty",
+      # We use this bundler class to detect the Requested ruby version from the Gemfile.lock
+      # Rails 8 stopped generating `RUBY VERSION` in the Gemfile.lock and started generating
+      # a `.ruby-version` file. This will observe the formats to help guide implementation
+      # decisions
+      "ruby.dot_ruby_version" => dot_ruby_version_file.exist? ? dot_ruby_version_file.read&.strip : nil
```
edmorley, line 122: Will the current implementation correctly escape newlines and other special characters within the file when outputting the YAML document?

edmorley (approved):

lib/language_pack/helpers/bundler_wrapper.rb
schneems:
```
irb(main):007> yaml = {"key" => "value", "empty" => nil, "another key" => 2 }.to_yaml and puts yaml
---
key: value
empty:
another key: 2
```

edmorley: I mean if `.ruby-version` is multi-line and/or contains a mixture of say single and double quotes.

schneems: We are stripping the value so there won't be trailing whitespace.

```
irb(main):016> " yolo  ".strip
=> "yolo"
```

Here's how it handles trailing whitespace (as an FYI) as well as escaping single and double quotes:

```
irb(main):014> yaml = {"key" => "value\n\n\n\n", "single" => "'single", "double" => "\"double", "mixed" => "'mixed\"", "other" => 'item'}.to_yaml and puts yaml
---
key: |+
  value



single: "'single"
double: "\"double"
mixed: '''mixed"'
other: item
```

edmorley: > We are stripping the value so there won't be trailing whitespace.

Yeah, I was meaning more similar to users that do eg this for their `.python-version` file (the below is valid syntax supported by pyenv etc):

```
# some comment about why they chose this version

3.13
```

Thank you for checking the quotes etc  :-)

PR #1486, opened 2024-07-20
Use `--update=none` instead of `-n` for `copy` in `cache.rb`

JanNash:
Fixes #1485.

As per the man page (https://man7.org/linux/man-pages/man1/cp.1.html):

image

image

image

PR #1583, opened 2025-04-24
Use `--update=none` instead of `-n` for `copy` in `cache.rb`

schneems:
Run tests for https://github.com/heroku/heroku-buildpack-ruby/pull/1486

### DIFF
--- a/lib/language_pack/cache.rb
+++ b/lib/language_pack/cache.rb
@@ -58,7 +58,7 @@ class LanguagePack::Cache
     return unless @cache_base
 
     dest ||= path
-    copy (@cache_base + path), dest, '-a -n'
+    copy (@cache_base + path), dest, '-a --update=none'
   end
 
   # copy cache contents
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fix warning message about `cp -n` (https://github.com/heroku/heroku-buildpack-ruby/pull/1583)
 
 ## [v302] - 2025-04-16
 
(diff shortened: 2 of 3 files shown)


runesoerensen (approved):

lib/language_pack/cache.rb
```diff
@@ -58,7 +58,7 @@ def load_without_overwrite(path, dest=nil)
     return unless @cache_base
 
     dest ||= path
-    copy (@cache_base + path), dest, '-a -n'
+    copy (@cache_base + path), dest, '-a --update=none'
```
edmorley, line 61: @schneems Seen in a [support ticket](https://heroku.support/search?utf8=%E2%9C%93&q=%22cp%3A+option+%27--update%27%22), the Ruby build is failing for a Heroku-20 app, with:

```
remote: cp: option '--update' doesn't allow an argument
remote: Try 'cp --help' for more information.
```

It looks like the version of `cp` that's included with Ubuntu 20.04 and 22.04 only supports `--update` not `--update=...`:
https://manpages.ubuntu.com/manpages/focal/en/man1/cp.1.html
https://manpages.ubuntu.com/manpages/jammy/en/man1/cp.1.html

>        -u, --update
>               copy  only when the SOURCE file is newer than the destination file or when the destination file is
>               missing

Compare to the Ubuntu 24.04 man-page:
https://manpages.ubuntu.com/manpages/noble/en/man1/cp.1.html

>        --update[=UPDATE]
>               control which existing files are updated; UPDATE={all,none,older(default)}.  See below

It seems CI coverage is missing for at least this feature on older stacks?

PR #1585, opened 2025-04-28
Fix cache code on herouku-22

schneems:
Reference:

https://github.com/heroku/heroku-buildpack-ruby/pull/1583#pullrequestreview-2799517426

Internal ticket 1570694.

Ubuntu 22.04 does not support `--update=<value>` they only support `--update <value>`. Which would cause an error:

```
cp: option '--update' doesn't allow an argument
Try 'cp --help' for more information.
```


schneems:
This won't fix it, and clearly doesn't exercise the code

schneems:
The issue is with heroku-20 not heroku-22 which is why the tests weren't failing. I'm jumping ship to a new PR and I'll have to manually validate it

PR #1586, opened 2025-04-28
Fix `cp` update none on Heroku 20

schneems:
In https://github.com/heroku/heroku-buildpack-ruby/pull/1583 a warning was bypassed by switching from `-n` to `--update none`. Unfortunately this triggers a problem with the version of `cp` on `heroku-20` (which is EOL in 2 days) due to not having that flag:

```
remote: cp: option '--update' doesn't allow an argument
remote: Try 'cp --help' for more information.
```

Internal ticket link https://heroku.support/1570694.

GUS-W-18382830

### DIFF
--- a/lib/language_pack/cache.rb
+++ b/lib/language_pack/cache.rb
@@ -58,7 +58,12 @@ class LanguagePack::Cache
     return unless @cache_base
 
     dest ||= path
-    copy (@cache_base + path), dest, '-a --update=none'
+
+    if ENV["STACK"] == "heroku-20"
+      copy (@cache_base + path), dest, "-a -n"
+    else
+      copy (@cache_base + path), dest, "-a --update=none"
+    end
   end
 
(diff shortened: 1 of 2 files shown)


joshwlewis (approved):

lib/language_pack/cache.rb
```diff
@@ -58,7 +58,12 @@ def load_without_overwrite(path, dest=nil)
     return unless @cache_base
 
     dest ||= path
-    copy (@cache_base + path), dest, '-a --update=none'
+
+    if ENV["STACK"] == "heroku-20"
```
edmorley, line 62: @schneems This needs to be both Heroku-20 and Heroku-22, since only Heroku-24 has the new flag. See the man page links I included in:
https://github.com/heroku/heroku-buildpack-ruby/pull/1583#discussion_r2063787708

PR #1591, opened 2025-05-04
Remove Heroku-20 support

schneems:
Since Heroku-20 is has reached end-of-life:
https://devcenter.heroku.com/changelog-items/3230

- Removed:

  - [x] "20.04"
  - [x] "heroku-20"
  - [x] "builder:20"
  - [x] Removes Ruby < 3.1 support
  - [x] Removes bundler < 2.3 support
  
  
GUS-W-14707069


lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -36,8 +36,7 @@ class LanguagePack::Helpers::BundlerWrapper
   include LanguagePack::ShellHelpers
 
   BLESSED_BUNDLER_VERSIONS = {}
-  # Heroku-20's oldest Ruby verison is 2.5.x which doesn't work with bundler 2.4
-  BLESSED_BUNDLER_VERSIONS["1"] = "1.17.3"
+  # Heroku-22's oldest Ruby verison is 3.1
```
colincasey, line 39:
```suggestion
  # Heroku-22's oldest Ruby version is 3.1
```

spec/helpers/download_presence_spec.rb
```diff
@@ -21,6 +21,9 @@
     download = LanguagePack::Helpers::DownloadPresence.new(
       multi_arch_stacks: ["heroku-24"],
       file_name: 'ruby-3.0.5.tgz',
+      # Heroku 20 is not longer supported however heroku-22 and heroku-24
```
colincasey, line 24: Does this test make sense to keep if you have to use the `heroku-20` stack for it to be functional?

lib/language_pack/ruby.rb
```diff
@@ -924,13 +776,12 @@ def attribute(name, value, force_string = false)
...
 <% end %>
         DATABASE_YML
-      end
     end
   end
 
   def rake
     @rake ||= begin
-      rake_gem_available = bundler.has_gem?("rake") || ruby_version.rake_is_vendored?
+      rake_gem_available = true
```
colincasey, line 784: Since this is now always `true`, is this variable even necessary? Is there code not already accounted for this PR that depends on this boolean for conditional logic?

lib/language_pack/ruby.rb
```diff
@@ -1236,7 +1009,6 @@ def load_bundler_cache
 
     old_rubygems_version = @metadata.read(ruby_version_cache).strip if @metadata.exists?(ruby_version_cache)
     old_stack = @metadata.read(stack_cache).strip if @metadata.exists?(stack_cache)
-    old_stack ||= DEFAULT_LEGACY_STACK
```
colincasey, line 1239: Now that this default is gone, is it possible for this value to be `nil`? E.g.; could this line of code be reached and, if so, what would happen? 
```rb
puts "Purging Cache. Changing stack from #{old_stack} to #{@stack}"
```

colincasey (approved):
@schneems I've got a couple questions for you but they aren't blockers.

lib/language_pack/ruby.rb
schneems: Good point.

The ruby buildpack has the concept of "metadata" that it stores between builds. This is durable information that is unfortunately stored in the cache. When this stack based cache clearing code was introduced, we were not persisting the state of the old stack in metadata for 100% of deployed apps, so if one was missing, we defaulted back to this value. 

Tthat is fixed (100% of deploys that have a prior deploy, have this metadata value). However it's worth checking that the first deploy doesn't mistakenly say that it's clearing the cache. I'll need to validate that and get back to you.

spec/helpers/download_presence_spec.rb
schneems: I need at least 2 stacks with versions where the binaries have different versions. That won't be true again until heroku-26 is released. Then we can update the test:

```ruby
  it "handles multi-arch transitions for files that do not exist" do
    download = LanguagePack::Helpers::DownloadPresence.new(
      multi_arch_stacks: ["heroku-24"],
      file_name: 'ruby-3.0.5.tgz',
      # Heroku 20 is not longer supported however heroku-22 and heroku-24
      # have identical Ruby versions supported, this test can be updated to
      # use heroku-24 and heroku-26 (when that stack is released)
```

PR #1595, opened 2025-05-16
Update default Ruby version to 3.3.8

schneems:
Also updates the "bootstrap" version of Ruby that the buildpack uses to run itself to match the default version. Previously, it was pinned to 3.1.x due to needing to support Bundler 1.x for the heroku-20.

GUS-W-18542568

### DIFF
--- a/changelogs/unreleased/default_ruby.md
+++ b/changelogs/unreleased/default_ruby.md
@@ -1,0 +1,22 @@
+## Default Ruby version for new apps is now 3.3.8
+
+The [default Ruby version for new Ruby applications is 3.3.8](https://devcenter.heroku.com/articles/ruby-support#default-ruby-version-for-new-apps). You’ll only get the default if the application does not specify a ruby version.
+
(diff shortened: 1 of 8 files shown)


dzuelke (approved):

changelogs/unreleased/default_ruby.md
```diff
@@ -0,0 +1,22 @@
+## Default Ruby version for new apps is now 3.3.8
+
+The [default Ruby version for new Ruby applications is 3.3.8](https://devcenter.heroku.com/articles/ruby-support#default-ruby-version-for-new-apps). You’ll only get the default if the application does not specify a ruby version.
+
+Heroku highly recommends specifying your desired Ruby version. You can specify a Ruby version in your `Gemfile:
```
schneems, line 5: Blerg, missing backtick to close out `Gemfile` here. Caught before publishing the changelog, though.

changelogs/unreleased/default_ruby.md
```diff
@@ -0,0 +1,22 @@
...
+
+```term
```
schneems, line 7: And this should be `ruby` and not `term`

PR #1601, opened 2025-05-21
Refactor RubyVersion class

schneems:
The RubyVersion class is used to hold information about the application's Ruby version. It was previously tightly coupled to the parsing `bundle platform-- ruby` output. Now, its interface is standardized, but makes it into a simple store that can be populated by different information sources. This can be used to compare two different data sources for differences in the future.

Commits are small and ordered. There should be no net change to the end user behavior.

## Context of this work

Today, the order of operations for running the buildpack looks like this:

- Install bootstrap (buildpack) ruby version
- Detect and install a limited number of "blessed" bundler versions (from S3) 
- Use that version with the bootstrap Ruby to run `bundle platform --ruby` to detect the application Ruby version.

Moving into the future, an eventual goal is to be able to avoid running `bundle platform --ruby` with the bootstrap version of ruby (used by the buildpack). This execution exposes incompatibilities between the bootstrap ruby version and versions of bundler. Instead, the order will eventually be:

- Install bootstrap ruby version
- Read and parse the Gemfile.lock to determine Ruby version and install it
- Install the arbitrary bundler version found in Gemfile.lock (from `gem install bundler -v <version>`)

To support this, I want to populate RubyVersion from different sources at the same time so I can compare them, instrument them, and ensure consistent platform behavior. The work here is a stepping stone to this eventual goal.

Worth noting that the proposed future goal is the same order of execution as the CNB (without the need to install a bootstrap ruby as the CNB is written in Rust).

### DIFF
--- a/spec/helpers/ruby_version_spec.rb
+++ b/spec/helpers/ruby_version_spec.rb
@@ -21,7 +21,7 @@ describe "RubyVersion" do
     ruby_version   = LanguagePack::RubyVersion.new(bundler_output: "ruby-#{version_number}-p0")
     version        = "ruby-#{version_number}"
 
-    expect(ruby_version.version_without_patchlevel).to eq(version)
+    expect(ruby_version.version_for_download).to eq(version)
     expect(ruby_version.next_logical_version).to eq("ruby-2.5.1")
     expect(ruby_version.next_logical_version).to eq("ruby-2.5.1")
     expect(ruby_version.next_logical_version(2)).to eq("ruby-2.5.2")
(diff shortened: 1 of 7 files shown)


spec/helpers/ruby_version_spec.rb
```diff
@@ -84,10 +84,89 @@
...
-        expect(ruby_version.version_without_patchlevel).to eq(version)
+        expect(ruby_version.version_for_download).to eq(version)
+        expect(ruby_version.engine_version).to eq(version_number)
+        expect(ruby_version.engine).to eq(:ruby)
+      end
+    end
+  end
+
+  it "detects RC Ruby from Gemfile.lock" do
+    Hatchet::App.new("default_ruby").in_directory_fork do |_|
```
colincasey, line 98: Is the app name `default_ruby` accurate for this test and will it conflict with the test that follows which also uses `default_ruby` as the test name?

schneems: > will it conflict with the test that follows which also uses default_ruby as the test name?

This `in_directory_fork` moves the app to a tempdir and then runs the code in the block in a forked Ruby process, this gives us process isolation for things like env var modifications, and the tempdir allows for competing disk modifications (without conflict). So this should be perfectly safe.

> Is the app name default_ruby accurate for this test 

In this case the base `default_ruby` app isn't really being used. The body of the test replaces both the Gemfile and Gemfile.lock. I use this app by default (heh) as it's the smallest fixture I've got. I could build up an app from scratch instead of using a fixture in the future.

colincasey (approved):

PR #1610, opened 2025-06-03
Introduce `bundle list` based parser

schneems:
This allows us to determine what versions of gems an application is using after `bundle install` has run. Currently, the buildpack works by loading bundler internals into memory and using those to parse the `Gemfile.lock`. That approach introduces strong compatibility requirements for coupling between Ruby versions and bundler versions.

Additionally, recent versions of `bundle install` no longer output the list of gems installed:

```
$ bundle _2.3.27_ install
Using rake 13.2.1
Using base64 0.2.0
...
Using toml-rb 4.0.0
Bundle complete! 11 Gemfile dependencies, 31 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
```

```
$ bundle _2.4.22_ install
Bundle complete! 11 Gemfile dependencies, 31 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
```

It's helpful to have a list of installed gems in the build output for debugging purposes. We can use this `bundle list` output as a proxy for the old `bundle install` output.


edmorley (approved):

schneems:
As a public followup: The metrics show a large number of differences. On analysis, it comes from `bundle list` respecting the installed groups (test/development/production) while the current (legacy) method of parsing the Gemfile.lock does not. That means that gems such as "capybara" are found via the current method, but not by the updated `bundle list` method.

PR #1641, opened 2025-09-08
Default PUMA_PERSISTENT_TIMEOUT to match router 2.0

schneems:
Router 2.0 introduced keepalive support. It will close connections idle after 90 seconds https://devcenter.heroku.com/articles/http-routing#keepalives. We want to avoid a situation where they send a request right before puma closes the connection. We can do that by setting it to the same timeout the router uses + some value (such as 5 seconds). 

Puma 7.0 was just released, and I helped make that value configurable via env var https://github.com/puma/puma/pull/3378. This sets the value for any puma user to do this manually, in their `config/puma.rb`, and for Puma 7+ users they will get that value by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


schneems:
That's been merged and deployed https://devcenter.heroku.com/changelog-items/3412

PR #1651, opened 2025-10-03
Bundler 2.6.9 and 2.7.2

schneems:
These downloads are tested via:

```
rspec ./spec/helpers/fetcher_spec.rb 
```

GUS-W-19797826

### DIFF
--- a/lib/language_pack/helpers/bundler_wrapper.rb
+++ b/lib/language_pack/helpers/bundler_wrapper.rb
@@ -43,8 +43,9 @@ class LanguagePack::Helpers::BundlerWrapper
   BLESSED_BUNDLER_VERSIONS["2.6"] = "2.6.9"
   BLESSED_BUNDLER_VERSIONS["2.7"] = "2.7.2"
 
-  SMALLEST = BLESSED_BUNDLER_VERSIONS.sort.first.first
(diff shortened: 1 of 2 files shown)


lib/language_pack/helpers/bundler_wrapper.rb
```diff
@@ -40,18 +40,22 @@ class LanguagePack::Helpers::BundlerWrapper
   BLESSED_BUNDLER_VERSIONS["2.3"] = "2.3.25"
   BLESSED_BUNDLER_VERSIONS["2.4"] = "2.4.22"
   BLESSED_BUNDLER_VERSIONS["2.5"] = "2.5.23"
-  BLESSED_BUNDLER_VERSIONS["2.6"] = "2.6.2"
+  BLESSED_BUNDLER_VERSIONS["2.6"] = "2.6.9"
+  BLESSED_BUNDLER_VERSIONS["2.7"] = "2.7.2"
+
+  SMALLEST = BLESSED_BUNDLER_VERSIONS.sort.first.first
+  LARGEST = BLESSED_BUNDLER_VERSIONS.sort.last.first
```
edmorley, line 47: (for the future, fine for now) When the bundler version reaches 2.10 in the distant future, this sorting of strings will break since 2.10 will be sorted before 2.3 I presume?

edmorley (approved):

PR #1733, opened 2026-04-02
Update anchor link to nodejs section in Ruby devcenter article

mathias:
Fixes the anchor link in build output so that it points to the right place in the doc.


schneems:
Thanks!

schneems (approved):

schneems (approved):

PR #1739, opened 2026-05-13
Support apps using the trilogy adapter

byroot:
It's the new recommended adapter for MySQL.


dannyfallon:
I tried doing this in #1714 but #1715 was the fix implemented

byroot:
Ah my bad, I should have asked before.

PR #1740, opened 2026-05-13
Use more portable detection code

byroot:
Use `uname` for architecture detection.
`dpkg` isn't available on Amazon Linux 2023 and that's what we are building under.

Also use `command -v` over `which`, as it's supported by more shells.

### DIFF
--- a/lib/language_pack/base.rb
+++ b/lib/language_pack/base.rb
@@ -18,7 +18,7 @@ class LanguagePack::Base
   VENDOR_URL = ENV["BUILDPACK_VENDOR_URL"] || "https://heroku-buildpack-ruby.s3.dualstack.us-east-1.amazonaws.com"
   ROOT_DIR = File.expand_path("../../..", __FILE__)
   AMD_ONLY_STACKS = ["heroku-22"]
-  KNOWN_ARCHITECTURES = ["amd64", "arm64"]
+  KNOWN_ARCHITECTURES = ["amd64", "arm64", "x86_64", "aarch64"]
 
   attr_reader :app_path, :bundler, :cache, :environment_name, :stack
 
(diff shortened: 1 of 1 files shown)


edmorley:
>  dpkg isn't available on Amazon Linux 2023 and that's what we are building under.

Just be aware that the Ruby binaries used by this buildpack are compiled for Ubuntu (and specifically our Ubuntu base images that have a certain set of system libraries installed), so trying to run them on alternative distros is likely to lead to pain and issues (and definitely isn't officially supported by this buildpack).

> Also use `command -v` over `which`, as it's supported by more shells.

This change was already made in #1716.

lib/language_pack/base.rb
```diff
@@ -40,7 +40,7 @@ def initialize(app_path:, arch:, bundler:, cache_path:, environment_name:, gemfi
   end
 
   def self.get_arch
-    command = "dpkg --print-architecture"
+    command = "uname -m"
```
edmorley, line 43: I haven't looked closely, but will this even work as is?

I believe the arch value here is what gets directly fed into the S3 URL, so I'm pretty sure this will 404 as is. (Given the current URLs have values like `amd64` in them and not `x86_64`)

byroot:
> This change was already made in #1716.

Woops, sorry I'm trying to contribute our private patches upstream, I didn't notice this one :/

byroot:
> Just be aware that the Ruby binaries used by this buildpack are compiled for Ubuntu

Yes, we're using our own binaries. So that's not a problem for us.

lib/language_pack/base.rb
byroot: Ah good catch 🤦 . I'm afraid this patch we'll have to keep private. Sorry for the noise.

edmorley: No worries! :-)

PR #1738, opened 2026-05-13
Fix BUNDLE_CONFIG path

byroot:
`Pathname#join` can have somewhat surpising behavior when passed an absolute path:

```ruby
>> Pathname.new("/home").join("/.bundle/config").to_s
=> "/.bundle/config"
>> File.join("/home", "/.bundle/config")
=> "/home/.bundle/config"
```

PR #1742, opened 2026-05-15
Test: Fix BUNDLE_CONFIG path (from #1738)

schneems:
Testing branch for https://github.com/heroku/heroku-buildpack-ruby/pull/1738 to run CI with secrets.


colincasey (approved):
Does this need a test added?

PR #1744, opened 2026-05-20
Add DotRubyVersionFile parser for .ruby-version files

schneems:
Parses .ruby-version file contents and produces a LanguagePack::RubyVersion instance. Uses service object pattern: initialize stores contents, call returns a Result struct with ruby_version and a warnings array.

Supported formats:
- Plain version: `3.4.8`
- Ruby prefix: `ruby-3.4.8`
- With @org suffix: `ruby-3.4.8@company-name`
- Pre-release (dot or dash): `3.4.0.rc1`, `3.4.0-preview2`
- Comments (`#`) and blank lines are ignored

Invalid inputs produce warnings:
- Version specifiers (>=, ~>, etc.)
- Multiple version lines
- Unparseable content
- JRuby prefix (current implementation needs engine version + target ruby version, can't find an existing syntax in the wild that supports jruby)

GUS-W-22244611

### DIFF
--- a/lib/language_pack/helpers/dot_ruby_version_file.rb
+++ b/lib/language_pack/helpers/dot_ruby_version_file.rb
@@ -1,0 +1,118 @@
+require "language_pack/shell_helpers"
+require "language_pack/ruby_version"
+
+module LanguagePack
+  module Helpers
+    # Parses contents of `.ruby-version` file to transform it into a RubyVersion
+    #
+    # Example:
+    #
+    #   version = DotRubyVersionFile.new(
+    #     contents: File.read(".ruby-version")
+    #   ).call
+    #   assert_eq [], version.warnings
+    #   assert_eq :ruby, version.ruby_version.engine
+    class DotRubyVersionFile
+      VERSION_PATTERN = /\A(?<version>\d+\.\d+\.\d+)(?:[.-](?<pre>\S+))?\z/
+      JRUBY_PATTERN = /\Ajruby-/i
+      SPECIFIER_PATTERN = />=|<=|~>|>|</
+
+      Result = Data.define(:ruby_version, :warnings)
+
+      def initialize(contents:)
+        @contents = contents
+      end
+
+      def call
+        warnings = []
+        lines = meaningful_lines
+
+        if lines.empty?
+          return Result.new(ruby_version: nil, warnings: warnings)
+        end
+
+        line = lines.first
+        version_string = strip_ruby_prefix_and_at_suffix(line)
+
+        if lines.length > 1
+          warnings << <<~EOF
+            The `.ruby-version` file contains multiple version lines.
+            Only a single version is supported. Remove additional lines.
+
+            Contents:
+
+            ```
+            #{@contents}
+            ```
+          EOF
+        end
+
+        if version_string.match?(JRUBY_PATTERN)
+          warnings << <<~EOF
+            JRuby not supported in `.ruby-version` file.
+
+            The `.ruby-version` file contains a JRuby version however the
+            JRuby engine is not supported by `.ruby-version` on Heroku at this time.
+
+            Contents:
+
+            ```
+            #{@contents}
+            ```
+          EOF
+        end
+
+        if version_string.match?(SPECIFIER_PATTERN)
+          warnings << <<~EOF
+            Cannot parse version specifiers in `.ruby-version` file.
+
+            Only exact versions are supported such as `3.4.8` or `ruby-3.4.8`.
+            Got:
+
+            ```
+            #{@contents}
+            ```
+          EOF
+        end
+
+        if warnings.any?
+          Result.new(ruby_version: nil, warnings: warnings)
+        elsif (match = version_string.match(VERSION_PATTERN))
+          Result.new(
+            ruby_version: LanguagePack::RubyVersion.new(
+              pre: match[:pre],
+              engine: :ruby,
+              default: false,
+              ruby_version: match[:version],
+              engine_version: match[:version]
+            ),
+            warnings: warnings
+          )
+        else
+          warnings << <<~EOF
+            Cannot parse Ruby version from `.ruby-version` file.
+
+            Expected format: `3.4.8` or `ruby-3.4.8`. Got:
+
+            ```
+            #{@contents}
+            ```
+          EOF
+          Result.new(ruby_version: nil, warnings: warnings)
+        end
+      end
+
+      private def meaningful_lines
+        @contents.gsub(/\s*#.*/, "").each_line.filter_map { |line|
+          stripped = line.strip
+          stripped unless stripped.empty?
+        }
+      end
+
+      private def strip_ruby_prefix_and_at_suffix(line)
+        version = line.delete_prefix("ruby-")
+        version.split("@").first
+      end
+    end
+  end
+end
--- a/lib/language_pack/helpers/dot_ruby_version_file.rb
+++ b/lib/language_pack/helpers/dot_ruby_version_file.rb
@@ -92,7 +92,8 @@ module LanguagePack
           warnings << <<~EOF
             Cannot parse Ruby version from `.ruby-version` file.
 
-            Expected format: `3.4.8` or `ruby-3.4.8`. Got:
+            Only full version specifiers with major, minor, and patch are supported
+            such as `3.4.8` or `ruby-3.4.8`. Got:
 
             ```
             #{@contents}
--- a/lib/language_pack/helpers/dot_ruby_version_file.rb
+++ b/lib/language_pack/helpers/dot_ruby_version_file.rb
@@ -112,7 +112,7 @@ module LanguagePack
 
       private def strip_ruby_prefix_and_at_suffix(line)
         version = line.delete_prefix("ruby-")
-        version.split("@").first
+        version.split("@").first || ""
       end
     end
   end
--- a/spec/helpers/dot_ruby_version_file_spec.rb
+++ b/spec/helpers/dot_ruby_version_file_spec.rb
@@ -138,41 +138,41 @@ describe "DotRubyVersionFile" do
     result = LanguagePack::Helpers::DotRubyVersionFile.new(contents: "ruby >= 3.1.6, < 3.3").call
     expect(result.ruby_version).to be_nil
     expect(result.warnings.length).to eq(1)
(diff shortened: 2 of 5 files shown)


lib/language_pack/helpers/dot_ruby_version_file.rb
```diff
@@ -0,0 +1,119 @@
...
+    #
+    # Example:
+    #
+    #   version = DotRubyVersionFile.new(
+    #     contents: File.read(".ruby-version")
+    #   ).call
+    #   assert_eq [], version.warnings
+    #   assert_eq :ruby, version.ruby_version.engine
+    class DotRubyVersionFile
+      VERSION_PATTERN = /\A(?<version>\d+\.\d+\.\d+)(?:[.-](?<pre>\S+))?\z/
```
colincasey, line 16: Do we do anything with the `pre` value? Could the regex for that be constrained to something less permissive than `\S+` (e.g.; only alphanumeric)?

spec/helpers/dot_ruby_version_file_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    expect(result.ruby_version).to be_nil
+    expect(result.warnings.length).to eq(1)
+    expect(result.warnings.first).to include("JRuby")
+  end
+
+  it "warns for version specifiers with >=" do
+    result = LanguagePack::Helpers::DotRubyVersionFile.new(contents: "ruby >= 3.1.6, < 3.3").call
+    expect(result.ruby_version).to be_nil
+    expect(result.warnings.length).to eq(1)
+    expect(result.warnings.first).to include("version specifiers")
```
colincasey: This assertion (also repeated elsewhere) can match multiple warning messages which is probably not what you intended.

lib/language_pack/helpers/dot_ruby_version_file.rb
```diff
@@ -0,0 +1,119 @@
...
+            #{@contents}
+            ```
+          EOF
+        end
+
+        if version_string.match?(JRUBY_PATTERN)
+          warnings << <<~EOF
+            JRuby not supported in `.ruby-version` file.
+
+            The `.ruby-version` file contains a JRuby version however the
```
colincasey, line 54: What's the recommended action for this warning? Specify JRuby in the Gemfile?

colincasey (approved):
@schneems since this is just the parser and it's not wired up to anything, it might be worth mentioning in the PR body that decisions like version resolution precedence, surfacing warnings, telemetry, etc. will come later.

lib/language_pack/helpers/dot_ruby_version_file.rb
schneems: It is technically any value before the space. It could be an emoji, and that would be valid. In practice it's alphanumeric and really restricted to only "pre" and "rc" with number suffixes. The only suffix that would break our logic was if it was `@` since we're splitting on it. I think this is good.

lib/language_pack/helpers/dot_ruby_version_file.rb
schneems: The warnings will be non-blocking. To remove the warning, someone could delete their `.ruby-version` file. But I want it to keep showing so someone doesn't say "dang it, I updated the `.ruby-version` why isn't it being used!"

I've captured N days of `.ruby-version` files, and so far no one has tried to use `jruby-` in one. Therefore this warning is mostly pre-emptive, very unlikely to hit IRL (at least for now).

Also, while I'm here: The reason we can't support jruby (as is) is that we require two values, the engine version i.e. JRuby version, and the target compatibility version i.e. `4.0.3` (like the ruby/ruby` version). Given there's no/limited prior art here, I don't want to force people into that naming. We could support JRuby in the future, but I would want either a manifest file or to save jruby binaries with their raw jruby version (instead of engine version + target compat version).

spec/helpers/dot_ruby_version_file_spec.rb
schneems: Good catch. Thanks. Updating.

lib/language_pack/helpers/dot_ruby_version_file.rb
```diff
@@ -0,0 +1,119 @@
+require "language_pack/shell_helpers"
+require "language_pack/ruby_version"
+
+module LanguagePack
+  module Helpers
+    # Parses contents of `.ruby-version` file to transform it into a RubyVersion
```
edmorley, line 6: > Supported formats:
> - Plain version: `3.4.8`
> - Ruby prefix: `ruby-3.4.8`
> - With @org suffix: `ruby-3.4.8@company-name`
> - Pre-release (dot or dash): `3.4.0.rc1`, `3.4.0-preview2`
> - Comments (`#`) and blank lines are ignored

A few of these seem surprising to me (eg the `@org suffix` one) given they aren't supported by `.python-version` and that was mostly created by pyenv that was inspired by rbenv (from which I imagine `.ruby-version` support similarly originated). 

What tools support the extended syntax like that?

Is there an official spec for `.ruby-version` somewhere, or is it like Python in that `.python-version` isn't an officially specified file format? (Context: For Python I chose to only support the simpler syntax formats, and also intentionally reject files containing rarely used features we don't support, like multiple versions etc, rather than trying to handle everything)

lib/language_pack/helpers/dot_ruby_version_file.rb
edmorley, line 16: Claude says the following:

- **4-part versions silently mis-parse** (`lib/language_pack/helpers/dot_ruby_version_file.rb:16`). The regex `\A(?<version>\d+\.\d+\.\d+)(?:[.-](?<pre>\S+))?\z` accepts `3.4.8.0` as `version=3.4.8, pre=0` and `9.4.9.0` (a JRuby-shaped version without `jruby-` prefix) the same way. The buildpack will then try to download `ruby-3.4.8.0.tgz` and fail downstream with a confusing message. Either reject 4+ segments or restrict `pre` to non-numeric-leading tokens. Add a regression test pinning intended behavior.

lib/language_pack/helpers/dot_ruby_version_file.rb
```diff
@@ -0,0 +1,119 @@
...
+              ruby_version: match[:version],
+              engine_version: match[:version]
+            ),
+            warnings: warnings
+          )
+        else
+          warnings << <<~EOF
+            Cannot parse Ruby version from `.ruby-version` file.
+
+            Only full version specifiers with major, minor, and patch are supported
```
edmorley, line 95: Claude says the following:

- **Conflicting "version specifier" terminology in user-facing warnings** (`dot_ruby_version_file.rb:67` vs `:95`). Line 67 rejects "version specifiers" (`>=`, `~>`); line 95 says "Only full version specifiers with major, minor, and patch are supported" — same phrase, opposite meaning, in adjacent warnings. Rename line 95 to "Only full version numbers" / "complete `MAJOR.MINOR.PATCH` versions".

lib/language_pack/helpers/dot_ruby_version_file.rb
```diff
@@ -0,0 +1,119 @@
...
+  module Helpers
+    # Parses contents of `.ruby-version` file to transform it into a RubyVersion
+    #
+    # Example:
+    #
+    #   version = DotRubyVersionFile.new(
+    #     contents: File.read(".ruby-version")
+    #   ).call
+    #   assert_eq [], version.warnings
+    #   assert_eq :ruby, version.ruby_version.engine
```
edmorley, line 14: Claude says the following:

- **Class-level docstring uses invented assertion syntax** (`dot_ruby_version_file.rb:13-14`). `assert_eq` is not RSpec, Minitest, or anything in this repo. Sibling `gemfile_lock.rb:17-18` uses `expect(...).to eq(...)`. Fix to match.

lib/language_pack/helpers/dot_ruby_version_file.rb
```diff
@@ -0,0 +1,119 @@
...
+    class DotRubyVersionFile
+      VERSION_PATTERN = /\A(?<version>\d+\.\d+\.\d+)(?:[.-](?<pre>\S+))?\z/
+      JRUBY_PATTERN = /\Ajruby-/i
+      SPECIFIER_PATTERN = />=|<=|~>|>|</
```
edmorley, line 18: Claude says the following:

- **Bare `<`, `<=`, `>` specifiers untested.** `SPECIFIER_PATTERN = />=|<=|~>|>|</` covers them; tests only exercise `>=`, `~>`, and combined `>= ... <`.

lib/language_pack/helpers/dot_ruby_version_file.rb
```diff
@@ -0,0 +1,119 @@
...
+        else
+          warnings << <<~EOF
+            Cannot parse Ruby version from `.ruby-version` file.
+
+            Only full version specifiers with major, minor, and patch are supported
+            such as `3.4.8` or `ruby-3.4.8`. Got:
+
+            ```
+            #{@contents}
+            ```
```
edmorley, line 100: Claude says the following:

- **Triple-backtick fences embed `@contents` raw** (lines 44, 59, 72). A `.ruby-version` containing a backtick line breaks rendering wherever output is markdown-rendered. Use `.inspect` or 4+ fences.

lib/language_pack/helpers/dot_ruby_version_file.rb
schneems: > What tools support the extended syntax like that?

Instead of looking at existing tools, I looked at what people were using while deploying on Heroku. Several had that pattern. 

Looking into it, it looks like it comes from RVM and the gemset concept (a way to isolate/limit installed libraries to a namespace) `rvm use 2.2.2@tptapp --create`. https://github.com/rvm/rvm/blob/e5795eddb2413d3c4de5bf85f5314ecd5d2fc118/scripts/functions/selector#L318

There's no official `.ruby-version` spec.

lib/language_pack/helpers/dot_ruby_version_file.rb
schneems: I originally thumbsed this down, but realized someone could use a bad "version specifier" even if it's not technically a valid bundler one like `=>4.0.5` or `=4.0.5`. If we see any of these symbols `/((>|<|~|=)+)/` I think we can warn about them. Better yet, extract the string we found that makes us think it's a version specifier .

I'll add that to my upcoming PR.

PR #1745, opened 2026-05-28
Capture `.ruby-version` differences with `Gemfile.lock` 

schneems:
To roll out support for `.ruby-version` files, I need to determine what to do when there's a conflict (i.e. both `Gemfile.lock` and `.ruby-version` have valid contents, but they disagree). To inform that decision, we can parse the and compare the two files and send the values to honeycomb to query.

The primary questions (and corresponding queries) look like this:

- **How many apps have .ruby-version but no version in Gemfile.lock?** `WHERE dot_ruby_version.version EXISTS AND gemfile_lock.ruby_version.default = true`
- **How many apps have .ruby-version that matches Gemfile.lock?** `WHERE dot_ruby_version.vs_gemfile_lock = "match"`
- **How many differ?** `WHERE dot_ruby_version.vs_gemfile_lock != "match" AND dot_ruby_version.vs_gemfile_lock EXISTS`
- **When they disagree, which file has the higher version?** `GROUP BY dot_ruby_version.vs_gemfile_lock, COUNT` filtered to `WHERE dot_ruby_version.vs_gemfile_lock != "match"`

GUS-W-22244654

### DIFF
--- a/lib/language_pack.rb
+++ b/lib/language_pack.rb
@@ -35,10 +35,18 @@ module LanguagePack
     metadata = LanguagePack::Metadata.new(cache_path: cache_path)
     new_app = metadata.empty?
 
+    dot_ruby_version_path = app_path.join(".ruby-version")
+    dot_ruby_version_result = if dot_ruby_version_path.exist?
+      LanguagePack::Helpers::DotRubyVersionFile.new(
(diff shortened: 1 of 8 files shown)


Malax (approved):
Looks good to me for what it intents to solve. 👍🏻

PR #1761, opened 2026-08-25
Download Yarn from npm registry instead of heroku-nodebin S3

schneems:
The `heroku-nodebin` S3 bucket is being sunset, so point the Yarn download at the public npm registry tarball instead (`https://registry.npmjs.org/yarn/-/yarn-<version>.tgz`).

The npm tarball is structurally identical to the S3 one (both extract to `package/` with the same `bin/yarn` and `bin/yarnpkg`), so the existing `Fetcher#fetch_untar` + `strip_components: 1` install path and the resulting on-disk layout are unchanged. Only the URL string changes; note npm's filename convention has no `v` prefix and uses `.tgz`.

Add a unit spec guarding that `Nodebin.yarn` points at `registry.npmjs.org` so it can't silently regress back to the S3 bucket.

W-21368290

### DIFF
--- a/spec/helpers/nodebin_spec.rb
+++ b/spec/helpers/nodebin_spec.rb
@@ -1,0 +1,13 @@
+require "spec_helper"
+
+describe LanguagePack::Helpers::Nodebin do
+  describe ".yarn" do
+    it "downloads Yarn from the npm registry (not the heroku-nodebin S3 bucket)" do
+      url = LanguagePack::Helpers::Nodebin.yarn["url"]
+
+      expect(url).to include("registry.npmjs.org")
+      expect(url).to_not include("heroku-nodebin")
+      expect(url).to eq("https://registry.npmjs.org/yarn/-/yarn-#{LanguagePack::Helpers::Nodebin::YARN_VERSION}.tgz")
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/helpers/nodebin_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "spec_helper"
+
+describe LanguagePack::Helpers::Nodebin do
+  describe ".yarn" do
```
colincasey, line 4: Does this spec invoke the Yarn binary at all? If not, could it? 

I'm sure this PR should be fine as the S3 tarball and the package tarball should be equivalent in terms of content so if there were no problems extracting and running the S3 tarball, there should be none here. It would be nice to verify Yarn binary execution was successful regardless.

colincasey (approved):

spec/helpers/nodebin_spec.rb
schneems: It is invoked by proxy via hatchet tests. This one asserts we're running `yarn install`

```
  it "works with jsbundling" do
    Hatchet::Runner.new("rails-jsbundling").tap do |app|
      app.deploy do
        expect(app.output).to include("yarn install")
        expect(app.output).to include("Asset precompilation completed")
      end
    end
  end
  ```

PR #1765, opened 2026-09-06
Fix typos in comments, docs, and user-facing messages

Avicennasis:
## Summary

Fix 10 typos across 5 files. Mostly comments and YARD docs; four are in strings that reach a user.

### Comments and YARD docs (no runtime effect)

- **lib/language_pack/helpers/outdated_ruby_version.rb**: `existance of` -> `existence of` (3 occurrences, lines 106, 136, 163)
- **lib/language_pack/helpers/outdated_ruby_version.rb**: `the last elment in the series` -> `element`
- **lib/rake/deploy_check.rb**: `# Returns tuthy value` -> `truthy` (the sibling method's comment four lines below already reads "Returns a truthy value")
- **lib/language_pack/shell_helpers.rb**: `@option options [Error] :error_class ... defaults to Standard Error` -> `StandardError` (the code on the next line literally does `options[:error_class] || StandardError`)

### Message strings (text-only; please sanity-check these four)

- **lib/language_pack/rails3.rb**: build warning read `For more information can be found in this article:`, which conflates "For more information, see" with "More information can be found". Now `More information can be found in this article:`.
- **lib/language_pack/ruby_version.rb**: `BadVersionError` message `Can not parse Ruby Version:` -> `Cannot parse Ruby Version:`.
- **lib/language_pack/helpers/outdated_ruby_version.rb**: internal `raise` message `Must initalize threads with \`call()\` before joining` -> `initialize`.
- **lib/rake/deploy_check.rb**: `check_sync!` raises a `<<~EOM` heredoc whose second line begins with a stray, unmatched `"`, left over from when this was a double-quoted string. Inside a heredoc it is just a literal character, so the released error message renders as `"Make sure that you've pulled: ...`. Removed the stray quote.

I grepped `lib/`, `bin/` and `spec/` for every string touched here: none of them appear in any test assertion, so no spec changes were needed. `ruby -c` passes on all five files.

A CHANGELOG.md entry is included under `## [Unreleased]` so the `check-changelog` job passes; happy to drop it and take the `skip changelog` label instead if you would rather not spend a changelog line on this.

No identifiers renamed, no behavior changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed typos in code comments, YARD docs, and user-facing warning and error messages (no behavior change)
 
 ## [v362] - 2026-08-25
 
--- a/lib/language_pack/helpers/outdated_ruby_version.rb
+++ b/lib/language_pack/helpers/outdated_ruby_version.rb
@@ -53,7 +53,7 @@ class LanguagePack::Helpers::OutdatedRubyVersion
   def join
     return unless can_check?
     return true if @already_joined
-    raise "Must initalize threads with `call()` before joining" if @minor_version_threads.empty?
+    raise "Must initialize threads with `call()` before joining" if @minor_version_threads.empty?
 
     @eol_versions = @eol_version_threads.map(&:value).compact
     @minor_versions = @minor_version_threads.map(&:value).compact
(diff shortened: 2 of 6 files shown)


schneems:
I pulled your code into https://github.com/heroku/heroku-buildpack-ruby/pull/1767 so integration tests (which deploy to live heroku and need valid creds) can run. Overall looks good. The `Can not` -> `Cannot` is not strictly wrong, but I like the change. Thanks. I'll need approval from another engineer before that PR can be merged.

